PR #16, opened 2024-07-31
Create Build Evidence.

osaidwtd:


evidence/create_build.go
```diff
@@ -0,0 +1,128 @@
...
+	"strings"
+)
+
+type createEvidenceBuild struct {
+	createEvidenceBase
+	project string
+	build   string
+}
+
+func NewCreateEvidenceBuild(serverDetails *coreConfig.ServerDetails, predicateFilePath string, predicateType string, key string, keyId string,
```
denys-jfrog, line 21: Optional: as the types are the same, the `string` may be put just once:
```go
func NewCreateEvidenceBuild(serverDetails *coreConfig.ServerDetails, 
	predicateFilePath, predicateType, key, keyId, project, build string) Command {
```

evidence/create_build.go
```diff
@@ -0,0 +1,128 @@
...
+	err = c.uploadEvidence(envelope, subject)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *createEvidenceBuild) buildBuildInfoSubjectPath(artifactoryClient artifactory.ArtifactoryServicesManager) (string, error) {
+	build := strings.Split(c.build, ":")
```
denys-jfrog, line 67: Please, raise the question about the input parameters one more time.
This split means that neither Build Name nor Build Number may contain `:` (this is not the case).
Hence, it means that we do not support Builds with name/number having `:`.
Needs to be confirmed explicitly.

evidence/create_build.go
```diff
@@ -0,0 +1,128 @@
...
+	return buildInfoPath + "@" + buildInfoChecksum, nil
+}
+
+func getBuildLatestTimestamp(name string, number string, project string, artifactoryClient artifactory.ArtifactoryServicesManager) (string, error) {
+	buildInfo := services.BuildInfoParams{
+		BuildName:   name,
+		BuildNumber: number,
+		ProjectKey:  project,
+	}
+	res, b, err := artifactoryClient.GetBuildInfo(buildInfo)
```
denys-jfrog, line 94: Please, rename `b` to something like `ok`.

evidence/create_build.go
```diff
@@ -0,0 +1,128 @@
...
+	if !b {
+		errorMessage := fmt.Sprintf("failed to find build, name:%s, number:%s, project: %s", name, number, project)
+		if err != nil {
```
denys-jfrog, line 97: The order of conditions looks wrong, the `err` should be checked first:
```go
if err != nil {
    return "", err
}
if !ok {
    errorMessage := fmt.Sprintf("failed to find build, name:%s, number:%s, project: %s", name, number, project)
    return "", errorutils.CheckErrorf(errorMessage)
}
```

evidence/create_build.go
osaidwtd: done

evidence/create_build.go
osaidwtd: done

evidence/create_build.go
osaidwtd: done

evidence/create_build.go
osaidwtd: I talked with Noam, there is a chance to change all other operations to be splitter to --name --version.
will take it in a separate Jira ticket.

osaidwtd:
Create Evidence Build-Info Using CLI

PR #21, opened 2024-08-11
Create package evidence using cli

osaidwtd:


evidence/cli/flags.go
```diff
@@ -24,14 +24,17 @@ const (
...
+	packageRepoName      = "package-repo-name"
 
 	// Unique evidence flags
 	predicate       = "predicate"
 	predicateType   = "predicate-type"
 	subjectRepoPath = "subject-repo-path"
 	subjectSha256   = "subject-sha256"
 	key             = "key"
-	keyId           = "key-name"
+	KeyAlias        = "key-alias"
```
denys-jfrog, line 37: No need to export this constant, can be `keyAlias`.

evidence/create_package.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
+	coreConfig "github.com/jfrog/jfrog-cli-core/v2/utils/config"
+	"github.com/jfrog/jfrog-client-go/artifactory"
+	"github.com/jfrog/jfrog-client-go/artifactory/services"
+	"github.com/jfrog/jfrog-client-go/metadata"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+)
+
+const leadArtifactQueryTemplate = `{
```
lesnerd, line 17: Why putting a const if you dont use it as const?
Also, it is used only in one place just copy it to line 124

evidence/cli/command_package.go
```diff
@@ -0,0 +1,48 @@
...
+		epc.ctx.GetStringFlagValue(KeyAlias),
+		epc.ctx.GetStringFlagValue(packageName),
+		epc.ctx.GetStringFlagValue(packageVersion),
+		epc.ctx.GetStringFlagValue(packageRepoName))
+	return epc.execute(createCmd)
+}
+
+func (epc *evidencePackageCommand) validateEvidencePackageContext(ctx *components.Context) error {
+	if !ctx.IsFlagSet(packageVersion) || assertValueProvided(ctx, packageVersion) != nil {
+		return errorutils.CheckErrorf("'packageVersion' is a mandatory field for creating a Package evidence: --%s", packageVersion)
```
denys-jfrog, line 42: In this command, and also in the existing build/release-bundle commands: we should not log the internal names of the variables - having `packageVersion` or `packageRepoName` in the logs is confusing.

Instead, we should mention only those names, which a client is supposed to use, for example:
`--%s is a mandatory parameter for creating a Package evidence`

evidence/model/graphql.go
```diff
@@ -0,0 +1,26 @@
+package model
+
```
denys-jfrog, line 2: Sounds like `metadata.go` and `MetadataResponse` / `PackageVersionsResponse`.

denys-jfrog (approved):

evidence/model/graphql.go
osaidwtd: done

evidence/create_package.go
osaidwtd: i tried to change it, but i think also is not bad to have it private const here since its too long and a bit confusing to have it directly inside the small method.

PR #26, opened 2024-11-05
Evidence CLI - evidence does not support basic auth

osaidwtd:


evidence/cli/command_cli.go
```diff
@@ -105,6 +105,11 @@ func evidenceDetailsByFlags(ctx *components.Context) (*coreConfig.ServerDetails,
 		return nil, errors.New("platform URL is mandatory for evidence commands")
 	}
 	platformToEvidenceUrls(serverDetails)
+
+	if serverDetails.GetUser() != "" && serverDetails.GetPassword() != "" {
```
denys-jfrog, line 109: Could be even just `serverDetails.GetUser() != ""`, without a check for a password?

osaidwtd: @denys-jfrog if you try with web-login you get user with token not password.
we still need both validation.

PR #29, opened 2024-11-27
Evd creation

oshratZairi:
1. Add evd_signing_key to the environment variables and use it for evidence creation if the key flag is not provided.
2. Add evd_key_alias to the environment variables and use it, if available, instead of requiring the alias_key flag.

### DIFF
--- a/evidence/cli/command_cli.go
+++ b/evidence/cli/command_cli.go
@@ -70,13 +70,40 @@ func validateCreateEvidenceCommonContext(ctx *components.Context) error {
 	if !ctx.IsFlagSet(predicate) || assertValueProvided(ctx, predicate) != nil {
 		return errorutils.CheckErrorf("'predicate' is a mandatory field for creating evidence: --%s", predicate)
 	}
+
 	if !ctx.IsFlagSet(predicateType) || assertValueProvided(ctx, predicateType) != nil {
 		return errorutils.CheckErrorf("'predicate-type' is a mandatory field for creating evidence: --%s", predicateType)
 	}
-	if !ctx.IsFlagSet(key) || assertValueProvided(ctx, key) != nil {
-		return errorutils.CheckErrorf("'key' is a mandatory field for creating evidence: --%s", key)
+
+	if err := ensureKeyExists(ctx, key); err != nil {
+		return err
 	}
+
+	if !ctx.IsFlagSet(keyAlias) {
+		setKeyAliasIfProvided(ctx, keyAlias)
+	}
+
 	return nil
+}
+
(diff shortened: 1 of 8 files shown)


evidence/cli/command_cli.go
```diff
@@ -70,15 +70,42 @@ func validateCreateEvidenceCommonContext(ctx *components.Context) error {
...
 }
 
+func ensureKeyExists(ctx *components.Context, key string) error {
+	if ctx.IsFlagSet(key) && assertValueProvided(ctx, key) == nil {
+		return nil
+	}
+
+	signingKeyValue, _ := getEnvVariable(evdSigningKey)
+	if signingKeyValue == "" {
+		return errorutils.CheckErrorf("'key' or EVD_KEY must be provided when creating evidence: --%s", key)
```
lesnerd, line 96: I dont understand what should I do as a user.
1. Provide a param: `key`
2. Provide a environment variable under the name `EVD_KEY`

evidence/cli/command_cli.go
```diff
@@ -70,15 +70,42 @@ func validateCreateEvidenceCommonContext(ctx *components.Context) error {
...
+	}
+	ctx.AddStringFlag(key, signingKeyValue)
+	return nil
+}
+
+func setKeyAliasIfNeeded(ctx *components.Context, keyAlias string) {
+	evdKeyAliasValue, _ := getEnvVariable(evdKeyAlias)
```
lesnerd, line 103: Maybe the keyalias was provided as a parameter and the key was read from the environment variable

oshratZairi: if the key was provided as a param - we won't get here; the code in line 82 checks
if !ctx.IsFlagSet(keyAlias) {
       setKeyAliasIfNeeded(ctx, keyAlias)
}
if the key was provided as param - we do nothing and the param value will be taken. if not - we will check if we have env var - if so - we will use it.

evidence/cli/command_cli.go
oshratZairi: I agree. need a better msg for the user. The user must provide: --key OR define env var EVD_SIGNING_KEY. 
do u have a better suggestion for msg?

evidence/cli/command_cli.go
lesnerd: Then change the function name, if you call `setKeyAliasIfNeeded` you indeed need alias.
By the way it still not guaranteed to have such thing.

lesnerd:
.

lesnerd (approved):

PR #47, opened 2025-02-18
Migrate life cycle commands

bhanurp:
- [ ] This pull request is on the dev branch.
-----
- Migrated all lifecycle commands from `jfrog-cli` to `jfrog-cli-artifactory`.

## Depends On
https://github.com/jfrog/jfrog-cli-core/pull/1351


cliutils/distribution/distribute.go
```diff
@@ -0,0 +1,52 @@
+package distribution
+
+import (
+	"github.com/jfrog/jfrog-cli-core/v2/common/spec"
+	pluginsCommon "github.com/jfrog/jfrog-cli-core/v2/plugins/common"
+	"github.com/jfrog/jfrog-cli-core/v2/plugins/components"
+	distributionUtils "github.com/jfrog/jfrog-client-go/utils/distribution"
+)
+
+func CreateDefaultDistributionRules(c *components.Context) *spec.DistributionRules {
```
RobiNino, line 10: This file already exists in distribution/cli/distribute.go

cliutils/flagkit/flags.go
```diff
@@ -0,0 +1,201 @@
+package flagkit
```
RobiNino, line 1: Since this is only used by distribution / lifecycle commands, I suggest placing it under a non common package.

RobiNino (changes requested):

cliutils/flagkit/flags.go
bhanurp: it is only used distribution, lifecycle since we have moved only those 2 categories for now, all command flags including artifactory or any other types if moved in future can be placed here hence the common package.

Removed unused flags from `jfrg-cli`

cliutils/distribution/distribute.go
bhanurp: Moved it from cli-core since it is related to lifecycle and distribution commands. I couldnt find it in `cli-core/common/cliutils` is it part of any other package?

RobiNino: It exists in the [CLI](https://github.com/jfrog/jfrog-cli/blob/dev/utils/distribution/distribute.go). Please delete the duplication and use the one in this PR.

RobiNino (changes requested):

cliutils/distribution/distribute.go
bhanurp: yeah but the link above points to a different branch, the [PR](https://github.com/jfrog/jfrog-cli/pull/2878/files#diff-5c1add75e996fb79bdaadcf72c7942f2e811d4a634b2956be913bba3583d6076) I have raised from `migrate-life-cycle-commands` on jfrog-cli already takes care of it. Are you referring to any other file?

PR #55, opened 2025-03-17
Allow Build Publish with `overwrite` Flag When Build Number is New

bhanurp:
- [x] This pull request is on the dev branch.
-----
### **Issue**
When attempting to publish a build with the `overwrite` flag, the process fails if the `buildName` already exists in Artifactory but the `buildNumber` is new. The failure occurs due to a missing build number in Artifactory.

### **Current Behavior**
- Publishing fails with an error stating that the build number is not available.

### **Solution**
- Ensure that the build info is pushed even if the specified `buildNumber` is not yet available in Artifactory.


itsmeleela (approved):
Looks good to me

PR #91, opened 2025-06-24
Add properties for python artifacts

reshmifrog:
- [x] This pull request is on the main branch.
-----
Issue: When uploading Python artifacts using jf twine and publishing build info, the artifact path in the build info page shows ”No path found (externally resolved or deleted/overwritten)”. Even though the path of artifact is visible under the build.json. This was due to build-related properties like build.name and build.number not being set on the artifacts during the upload.

fix: Updated the upload process logic to ensure that the necessary properties — build.name and build.number are properly set on the artifacts when using jf twine upload.

### DIFF
--- a/artifactory/commands/python/twine.go
+++ b/artifactory/commands/python/twine.go
@@ -191,8 +191,8 @@ func (tc *TwineCommand) uploadAndCollectBuildInfo() error {
 
 	var fileInitials string
 	for _, arg := range artifacts {
-		if strings.HasSuffix(arg.Path, ".tar.gz") {
-			fileInitials = strings.TrimSuffix(arg.Path, ".tar.gz")
+		if strings.HasSuffix(arg.Name, ".tar.gz") {
+			fileInitials = strings.TrimSuffix(arg.Name, ".tar.gz")
 		}
 	}
 
--- a/artifactory/commands/python/twine.go
+++ b/artifactory/commands/python/twine.go
@@ -7,9 +7,6 @@ import (
 	"os/exec"
 	"strings"
 
-	ioutils "github.com/jfrog/gofrog/io"
-	"github.com/jfrog/jfrog-client-go/utils/io/content"
-
 	"github.com/jfrog/build-info-go/build"
 	"github.com/jfrog/build-info-go/utils/pythonutils"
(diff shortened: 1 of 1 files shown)


artifactory/commands/python/twine.go
```diff
@@ -176,6 +179,53 @@ func (tc *TwineCommand) uploadAndCollectBuildInfo() error {
...
+	servicesManager, err := rtUtils.CreateServiceManager(tc.serverDetails, -1, 0, false)
+	if err != nil {
+		return err
+	}
+
+	searchReader, err := servicesManager.SearchFiles(searchParams)
+	if err != nil {
+		log.Error("Failed to get uploaded twine package: ", err.Error())
+		return err
+	}
```
bhanurp, line 216: An extra API call of search can be avoided by reusing details of artifacts object, you try something like below
```
var artifactReaders []services.ResultItem
for _, artifact := range artifacts {
	artifactReaders = append(artifactReaders, services.ResultItem{
		Repo: tc.targetRepo,
		Path: artifact.Path,
		Name: artifact.Name,
	})
}

propsParams := services.PropsParams{
	Reader: services.NewResultItemsReader(artifactReaders),
	Props:  fmt.Sprintf("build.name=%s;build.number=%s,build.timestamp=%s", buildName, buildNumber, buildGenericDetails.Timestamp),
}

_, err = servicesManager.SetProps(propsParams)
if err != nil {
	log.Warn("Unable to set build properties: ", err, "\nThis may cause build to not properly link with artifact, please add build name and build number properties on the artifacts manually")
	return err
}
```

artifactory/commands/python/twine.go
```diff
@@ -194,3 +244,17 @@ func (tc *TwineCommand) isRepoConfigFlagProvided() bool {
...
+			`"repo": "%s",` +
+			`"$or": [{` +
+			`"$and":[{` +
+			`"path": {"$match": "*"},` +
+			`"name": {"$match": "%s*"}` +
+			`}]` +
+			`}]` +
+			`}`
+	return fmt.Sprintf(itemsPart, repo, fileInitial)
+}
```
bhanurp, line 260: This is not required if the same functionality can be achieved without search operation like suggested above.

bhanurp (changes requested):

bhanurp (approved):

PR #101, opened 2025-07-10
RTECO-255 - Handled aggregate condition correctly

naveenku-jfrog:
What: Added the condition to not include the previous build affected issue when aggregate option is false. This was a bug due to which previous build version affected issues were added to the current build affected issue irrespective of the aggregate field value.

- [ ] All [tests](https://github.com/jfrog/jfrog-cli-artifactory#tests) passed. If this feature is not already covered by the tests, I added new tests.
- [ ] All [static analysis checks](https://github.com/jfrog/jfrog-cli-artifactory/actions/workflows/analysis.yml) passed.
-----

### DIFF
--- a/artifactory/commands/buildinfo/addgit.go
+++ b/artifactory/commands/buildinfo/addgit.go
@@ -163,6 +163,10 @@ func (config *BuildAddGitCommand) collectBuildIssues() ([]buildinfo.AffectedIssu
 	}
 
 	var foundIssues []buildinfo.AffectedIssue
+	if !config.issuesConfig.Aggregate {
+		return foundIssues, nil
+	}
+
 	logRegExp, err := createLogRegExpHandler(config.issuesConfig, &foundIssues)
 	if err != nil {
 		return nil, err


artifactory/commands/buildinfo/addgit.go
```diff
@@ -163,6 +163,10 @@ func (config *BuildAddGitCommand) collectBuildIssues() ([]buildinfo.AffectedIssu
 	}
 
 	var foundIssues []buildinfo.AffectedIssue
+	if !config.issuesConfig.Aggregate {
```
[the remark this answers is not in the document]
naveenku-jfrog: Addressed it.

artifactory/commands/buildinfo/addgit.go
```diff
@@ -174,6 +175,11 @@ func (config *BuildAddGitCommand) collectBuildIssues() ([]buildinfo.AffectedIssu
 	if err != nil {
 		return nil, err
 	}
+
+	if !config.issuesConfig.Aggregate && len(foundIssues) > 1 {
+		return foundIssues[:1], nil
```
bhanurp, line 180: is the order preserved? means can we assume always that issue at index 0 is the issue number for current commit?

naveenku-jfrog: Changed the logic.

bhanurp (approved):

PR #129, opened 2025-08-18
Add support of docker/oci subject resolution by registry, image name and sha256

mnsboev:
-----
This pull request introduces support for Docker and OCI subject auto-resolution from Sigstore bundles using the registry, image name, and SHA-256, which should be provided in the Sigstore bundle provenance. All these changes are made for the evidence create command (jf evd create) with the --sigstore-bundle flag.

### DIFF
--- a/evidence/create/create_custom.go
+++ b/evidence/create/create_custom.go
@@ -20,6 +20,8 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 )
 
+const subjectsLimit = 10
+
 type createEvidenceCustom struct {
 	createEvidenceBase
 	subjectRepoPaths      []string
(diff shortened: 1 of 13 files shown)


selalererjfrog (approved):

evidence/sigstore/subject_extractor.go
```diff
@@ -2,43 +2,79 @@ package sigstore
...
-		if subject, ok := subjects[0].(map[string]any); ok {
-			if name, ok := subject["name"].(string); ok && name != "" {
-				return name
-			}
+// extractRepoPathFromStatement extracts the repository path and SHA256 checksum from a statement
+// The statement should contain a "subject" array with at least one subject object
+// Each subject should have a "name" field and optionally a "digest" field with SHA256
+func extractRepoPathFromStatement(statement map[string]any) (string, string) {
+	if statement == nil {
+		return "", ""
```
alenon, line 51: I’d strongly recommend returning errors instead of just empty strings for name or SHA256. 
That way the issue is clear. 
If you don’t want to do that, at least add some logs—otherwise it’s a total black box.

evidence/create/create_custom.go
```diff
@@ -60,26 +68,51 @@ func (c *createEvidenceCustom) Run() error {
...
 
-	err = c.validateSubject()
-	if err != nil {
-		return err
+	var errors []error
+	var successfulSubjects []string
+	var failedSubjects []string
+
+	for _, subjectRepoPath := range c.subjectRepoPaths {
+		if err := c.validateSubject(subjectRepoPath); err != nil {
```
alenon, line 84: If we’re expecting dozens of subjects, we should probably use goroutines to run things in parallel — otherwise it could take too long.
@avishayh - what do you think?

alenon (changes requested):

PR #164, opened 2025-09-17
Added jf stats rt function

naveenku-jfrog:
-----
Description:
Added APIs for handling statistics related.

Depend on other PRs: Yes,

https://github.com/jfrog/jfrog-cli/pull/3090
https://github.com/jfrog/jfrog-client-go/pull/1189


stats/stats.go
```diff
@@ -0,0 +1,688 @@
...
+		return err
+	}
+	sa.AccessManager = *accessManager
+
+	lifecycleServicesManager, err := utils.CreateLifecycleServiceManager(serverDetails, false)
+	if err != nil {
+		return err
+	}
+	sa.LifecycleServiceManager = *lifecycleServicesManager
+
```
bhanurp, line 91: nitpick: lot of unnecessary new lines before every line of code it decreases readability can you please remove them?

stats/stats.go
```diff
@@ -0,0 +1,688 @@
...
+}
+
+func (rb ReleaseBundleInfo) String() string {
+	var builder strings.Builder
+
+	builder.WriteString(fmt.Sprintf("ReleaseBundleName: %s\n", rb.ReleaseBundleName))
+	builder.WriteString(fmt.Sprintf("RepositoryKey: %s\n", rb.RepositoryKey))
+	builder.WriteString(fmt.Sprintf("ProjectKey: %s\n", rb.ProjectKey))
+
+	return builder.String()
```
bhanurp, line 190: can we solve it using fields tags instead of a new custom string function for every struct define fields so that which has to be written can also be easily controlled.

stats/stats.go
```diff
@@ -0,0 +1,688 @@
...
+	clientUtils "github.com/jfrog/jfrog-client-go/artifactory/services/utils"
+	"github.com/jfrog/jfrog-client-go/jpd"
+	"github.com/jfrog/jfrog-client-go/lifecycle"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+	"strings"
+)
+
+var displayLimit int
+
+type StatsArtifactory struct {
```
bhanurp, line 23: nitpick: ArtifactoryStats is more readable

stats/stats.go
```diff
@@ -0,0 +1,688 @@
...
+		for _, filter = range processingOrders {
+			allResultsMap[filter] = GetStatsUsingFilter(commandMap[filter])
+		}
+		updateProjectInArtifactoryInfo(&allResultsMap)
+	}
+	return sa.PrintAllResults(allResultsMap)
+}
+
+func (sa *StatsArtifactory) PrintAllResults(results map[string]interface{}) error {
+	for _, product := range printingOrders {
```
bhanurp, line 252: you don't need to print based on product only if stats subcommand rt it should come to jfrog-cli-artifactory.

stats/stats.go
```diff
@@ -0,0 +1,688 @@
+package stats
```
bhanurp, line 1: - File is very huge lets cut short move all utility functions into separate file in same package something like stats_utils.go.

- ArtifactoryStats struct{} should have methods like
    - WriteTable
    - WriteText
    - WriteJson
probably think about creating an interface StatsWriter{} with above methods.

- All formatting based on flag should go to above methods.

- Create a method called filter and apply filters here basically your Writers should call filter. and filter should call the right API's 

- If the filter has only releaseBundle then there is no need to make other API calls.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/distribution/reference v0.6.0
 	github.com/forPelevin/gomoji v1.3.1
 	github.com/gookit/color v1.6.0
+	github.com/jedib0t/go-pretty/v6 v6.6.8
```
bhanurp, line 10: we cant add this unless there are more details on license, maintenance of the package and what is memory footprint it is bringing to jfrog-cli binary

bhanurp (changes requested):

go.mod
naveenku-jfrog: Its already used here: https://github.com/jfrog/jfrog-cli-core/blob/master/go.mod#L15

stats/stats.go
naveenku-jfrog: Addressed it.

stats/stats.go
```diff
@@ -0,0 +1,688 @@
...
+
+func PrintArtifactoryStats(stats *ArtifactoryInfo) {
+	log.Output("--- Artifactory Statistics ---")
+	log.Output("Total Projects: ", stats.ProjectsCount)
+	log.Output("Total No of Binaries: ", stats.StorageInfo.BinariesCount)
+	log.Output("Total Binaries Size: ", stats.StorageInfo.BinariesSize)
+	log.Output("Total No of Artifacts: ", stats.StorageInfo.ArtifactsCount)
+	log.Output("Total Artifacts Size: ", stats.StorageInfo.ArtifactsSize)
+	log.Output("Storage Type: ", stats.StorageInfo.StorageType)
+	log.Output()
```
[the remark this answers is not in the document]
naveenku-jfrog: Addressed it.

go.mod
naveenku-jfrog: I have used "github.com/jedib0t/go-pretty/v6/text" just to change text color, otherwise whole logic is inside cli-core only.

bhanurp (approved):

PR #306, opened 2025-11-12
Fix build-docker-create with multiple tags in docker buildx metadata file

galusben:
…file with multiple tags (197)

- Added tests for splitting multi-tag Docker image strings in various formats.
- Improved error handling for cases with no valid images found.
- Implemented logic to derive repository from image when not explicitly set.

-----
This PR fix the issue described here: https://github.com/jfrog/jfrog-cli-artifactory/issues/197
In a nutshell, with current JFROG cli - buildx implementation, there is no support in JFrog CLI to add 2 tags for the same buildx command.

### DIFF
--- a/artifactory/commands/container/buildcreate.go
+++ b/artifactory/commands/container/buildcreate.go
@@ -9,6 +9,7 @@ import (
 	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
 	"github.com/jfrog/jfrog-client-go/artifactory"
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
 type BuildDockerCreateCommand struct {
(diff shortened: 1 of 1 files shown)


artifactory/commands/container/buildcreate.go
```diff
@@ -49,17 +50,31 @@ func (bdc *BuildDockerCreateCommand) Run() error {
...
+		return errorutils.CheckErrorf("no valid images found in image file")
+	}
+
+	// Get the repo argument (if provided) to use as fallback
+	// The repo from each image takes precedence to handle cases where tags might be in different repositories
+	fallbackRepo, _ := bdc.GetRepo()
+
 	for _, image := range images {
+		// Always try to get repo from the image first (takes precedence)
+		repo, err := bdc.getRepoFromImage(image, serviceManager)
```
fluxxBot: should we add debug log pointing out that the mandatory repository field part of the command args is not used but the is extracted from image name?

fluxxBot (approved):

artifactory/commands/container/buildcreate.go
fluxxBot: https://github.com/galusben/jfrog-cli-artifactory/blob/5b53aa6dd6d2c0406e45cab4bca4befa7077dfa2/artifactory/commands/container/buildcreate.go#L79, this one should do, log method is available in client-go -> "github.com/jfrog/jfrog-client-go/utils/log"

PR #343, opened 2026-01-14
Enhance Gradle init script handling by querying Java's user home

igalsi-t:
Fix Gradle init script written to wrong path in containers by using Java's user.home instead of $HOME


igalsi-t:

artifactory/commands/gradle/gradle.go
```diff
@@ -383,7 +384,14 @@ func GenerateInitScript(config InitScriptAuthConfig) (string, error) {
 func WriteInitScript(initScript string) error {
```
sverdlov93: unit tests for the new case

PR #346, opened 2026-01-15
Add curation CVS 404 handling for npm install

basel1322:
- Add handle404Errors to detect blocked packages during npm install/ci
- Check if 404/ETARGET errors are actually 403 (blocked by curation)

image

### DIFF
--- a/artifactory/commands/npm/npmcommand.go
+++ b/artifactory/commands/npm/npmcommand.go
@@ -41,7 +41,16 @@ const (
 )
 
 var (
-	npm404ErrorPattern      = regexp.MustCompile(`(?i)(404|ETARGET|notarget|No matching version found)`)
+	// npm404ErrorPattern matches common npm error indicators for missing packages.
+	// Example input: "notarget No matching version found for lodash@99.0.0"
+	// Matches: "ETARGET", "notarget", "No matching version found"
+	npm404ErrorPattern = regexp.MustCompile(`(?i)(404|ETARGET|notarget|No matching version found)`)
+
+	// packageWithVersionRegex extracts "pkg@version" from npm's "No matching version found"
+	// Example input: "No matching version found for @angular/core@15.0.0."
+	// Output: "@angular/core@15.0.0."
+	// Example input: "No matching version found for lodash@4.17.21"
+	// Output: "lodash@4.17.21"
 	packageWithVersionRegex = regexp.MustCompile(`No matching version found for\s+([@a-zA-Z0-9][a-zA-Z0-9._/-]*(?:/[a-zA-Z0-9._-]+)?@[0-9][0-9a-zA-Z._-]*)`)
 )
 


agrasth:
Can you also pls add appropriate label on the PR for github action `PR Label Check` failure.

agrasth (approved):

artifactory/commands/npm/npmcommand.go
```diff
@@ -314,9 +330,123 @@ func (nc *NpmCommand) Run() (err error) {
...
+	if strings.HasPrefix(fullName, "@") {
+		scopeAndName := fullName[1:]
+		scope, name, found := strings.Cut(scopeAndName, "/")
+		if !found {
+			return "", "", "", false
+		}
+		return scope, name, version, true
+	}
+
+	return "", fullName, version, true
```
bhanurp: making sure version is not empty and removed redundant new lines across funtion
```suggestion
func parsePackageSpec(packageSpec string) (scope, name, version string, ok bool) {
	versionSeparator := strings.LastIndex(packageSpec, "@")
	if versionSeparator == -1 || versionSeparator == 0 {
		return "", "", "", false
	}
	fullName := packageSpec[:versionSeparator]
	
	// npm error line is "No matching version found for pkg@version.", the trailing period is sentence punctuation and gets captured by the regex.
	version = strings.TrimSuffix(packageSpec[versionSeparator+1:], ".")
	if version == "" {
  	    return "", "", "", false
    }
	if strings.HasPrefix(fullName, "@") {
		scopeAndName := fullName[1:]
		scope, name, found := strings.Cut(scopeAndName, "/")
		if !found {
			return "", "", "", false
		}
		return scope, name, version, true
	}

	return "", fullName, version, true
```

cliutils/flagkit/flags.go
```diff
@@ -364,6 +364,7 @@ const (
 	npmDetailedSummary = npmPrefix + detailedSummary
 	runNative          = "run-native"
 	npmWorkspaces      = "workspaces"
+	DisableCVSCheck    = "disable-cvs-check"
```
bhanurp: why is it exposed to other package?
```suggestion
	disableCVSCheck    = "disable-cvs-check"
```

artifactory/commands/npm/npmcommand.go
```diff
@@ -314,9 +330,123 @@ func (nc *NpmCommand) Run() (err error) {
...
+			if blockedErr := nc.handle404Errors(err); blockedErr != nil {
+				err = blockedErr
+			}
+		}
+	}
 	return
 }
 
+// handle404Errors checks if a 404/ETARGET error is actually a blocked package (403)
+func (nc *NpmCommand) handle404Errors(installErr error) error {
```
bhanurp: handle only when curation is available on artifactory server. Is there a way to check it before even proceeding with handling 404 errors?

bhanurp (changes requested):

fluxxBot (changes requested):

artifactory/commands/npm/npmcommand.go
basel1322: I wish it was possible. for that case we added a flag that we can disable it when needed

bhanurp (approved):

PR #354, opened 2026-01-29
Support nested path for docker build

fluxxBot:
-----

Description:
It was found that uploading docker images inside a nested path caused failure in attaching build properties in the artifacts due to invalid path construction:
eg: docker-repo/myorg/image:latest when uploaded, cli tried to attach properties on docker-repo/image:latest, as a result attaching properties caused 404 not found failure.

Solution:
instead of using image full name which returned image:latest, we are using image full name without repo which returns myorg/image:latest

depends on:
1. https://github.com/jfrog/jfrog-cli/pull/3330


artifactory/commands/ocicontainer/docker_manifest.go
```diff
@@ -204,8 +207,28 @@ func (h *FatManifestHandler) getManifestShaListForImage(imageReference name.Refe
 // getLayersForManifestSha searches for layers across all manifest SHAs
 func (h *FatManifestHandler) getLayersForManifestSha(imageRef string, manifestShas []string, repository string) ([]utils.ResultItem, []utils.ResultItem, error) {
 	var foldersToApplyProps []utils.ResultItem
-	searchablePathForManifest := h.createSearchablePathForDockerManifestContents(imageRef, manifestShas)
 
+	// Get the image tag folder path (contains list.manifest.json)
+	image := NewImage(imageRef)
```
naveenku-jfrog: NewImage(imageRef)
Above instance is created again in func createSearchablePathForDockerManifestContents(). Can we use same image instance there?

PR #379, opened 2026-03-03
conan-recipe-not-in-root-dir

reshmifrog:
-----
Problem:
When running conan install --requires zlib/1.2.11 (without a conanfile.py or conanfile.txt), the JFrog CLI fails with the error "no conanfile.py or conanfile.txt found" because the build-info collection logic always expects a recipe file to exist, even though Conan 2.x officially supports installing dependencies directly via --requires and --tool-requires without any conanfile.
Solution:
Modified loadConanfile() in build-info-go to gracefully set default project metadata (name from directory, empty version, default user/channel) instead of failing when no conanfile is found. Updated buildGraphInfoArgs() to pass --requires and --tool-requires flags directly to conan graph info when no conanfile exists, and wired CLI overrides (--name, --version, --user, --channel) through jfrog-cli-artifactory so the build-info module ID can still be customized.

### DIFF
--- a/.github/workflows/cli-integration-tests.yml
+++ b/.github/workflows/cli-integration-tests.yml
@@ -523,24 +523,42 @@ jobs:
     name: Docker ubuntu
     needs: Detect-Deps
     if: github.event_name == 'workflow_dispatch' || github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'safe to test')
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - name: ubuntu
+            version: 24.04
+        disable-containerd-snapshotter: [ true, false ]
     runs-on: ubuntu-24.04
     env:
       JFROG_CLI_LOG_LEVEL: DEBUG
     steps:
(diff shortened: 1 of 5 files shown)


artifactory/commands/conan/command.go
```diff
@@ -303,6 +301,86 @@ func (c *ConanCommand) ServerDetails() (*config.ServerDetails, error) {
...
+// extractFlagValue checks if arg matches the given flag (either --flag=value or --flag value form)
+// and returns the extracted value, or the current value if no match.
+func extractFlagValue(arg string, idx int, args []string, flag, current string) string {
+	prefix := flag + "="
+	if strings.HasPrefix(arg, prefix) {
+		return strings.TrimPrefix(arg, prefix)
+	}
+	if arg == flag && idx+1 < len(args) {
+		return args[idx+1]
+	}
```
naveenku-jfrog: jf conan install . --name --version
For above arguments: code will set the name to --version because it simply takes idx+1.
Fix: Check if args[idx+1] starts with - before assigning it as a value.

PR #414, opened 2026-04-01
Improved error message when user tries to install SKILL blocked by xray

bhanurp:
-----
## Description
Improved message when install fails to download skills which is blocked by xray

### DIFF
--- a/skills/commands/install/install.go
+++ b/skills/commands/install/install.go
@@ -89,6 +89,9 @@ func (ic *InstallCommand) Run() error {
 
 	zipPath, err := ic.downloadZip(tmpDir)
 	if err != nil {
+		if strings.Contains(err.Error(), "403") {
+			return ic.diagnoseDownloadForbidden(err)
+		}
 		return fmt.Errorf("download failed: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


itsmeleela (approved):

skills/commands/install/install.go
```diff
@@ -187,6 +190,25 @@ func (ic *InstallCommand) downloadZip(tmpDir string) (string, error) {
...
 }
 
+// diagnoseDownloadForbidden checks the Xray status API when a download returns 403.
+// If the artifact is blocked by Xray, it returns a specific error message.
+// Otherwise, it returns the original error.
+func (ic *InstallCommand) diagnoseDownloadForbidden(originalErr error) error {
+	artifactPath := fmt.Sprintf("%s/%s/%s-%s.zip", ic.slug, ic.version, ic.slug, ic.version)
+	sm, err := utils.CreateServiceManager(ic.serverDetails, 3, 0, false)
+	if err != nil {
+		return fmt.Errorf("download blocked (403): %w", originalErr)
```
agrasth, line 200: Should we check for the actual HTTP status code from the error. 
something like `if errors.As(err, &httpErr) && httpErr.StatusCode == 403 {`

fluxxBot (approved):

skills/commands/install/install.go
```diff
@@ -187,6 +190,25 @@ func (ic *InstallCommand) downloadZip(tmpDir string) (string, error) {
...
+	}
+	resp, err := sm.GetSkillXrayStatus(ic.repoKey, artifactPath)
+	if err != nil {
+		return fmt.Errorf("download blocked (403): %w", originalErr)
```
agrasth, line 204: We can consider logging the diagnostic failure for debugging: `log.Debug("Failed to create service manager for Xray status check: %v", err)`

skills/commands/install/install.go
bhanurp: this can be an improvement as of now we dont have access to http status code, we can improve jfrog-client-go to return status code

skills/commands/install/install.go
bhanurp: no need of debug instead of originalErr we can add err so it is clear download blocked with 403

PR #415, opened 2026-04-07
fix/build-filter-in-pattern-aql

reshmifrog:
-----
**Problem:**  
 When a spec has both build AND pattern, the old code didn't recognize it as a build search. It treated it as a pattern-only search (WILDCARD), which meant:

  1. It generated an AQL query from the pattern — completely ignoring the build name
  2. This AQL scanned the entire repository matching the pattern (could be millions of files)
  3. Only after downloading all those results, it filtered them by the build's checksums in memory

 For large repositories like the customer's docker-local-ash, this made the command run for an extremely long time or timeout.

 Why Not Use a REST API Instead of AQL?

 We explored this. Artifactory has several REST APIs:

  - Build Info API (GET /api/build/{name}/{number}) — returns build metadata, but artifact paths are relative to the module, not the full Artifactory path. You'd need extra lookups to find where each file actually lives.
  - Build Artifacts API (GET /api/builds/buildArtifacts/...) — returns file locations, but has indexing delays for recently-published builds (our CI tests caught this — it returned 0 results for a build that was just published).
  - Checksum Search (GET /api/search/checksum) — requires one HTTP call per artifact. For a build with 500 artifacts, that's 500 HTTP round-trips.

 None of these REST APIs support combining a build filter with a pattern filter in a single call. We'd always need multiple API calls + client-side merging, which adds complexity and is slower than a single well-scoped AQL query.

 **The Fix:**

  We flipped the search strategy from "scan by pattern first, filter by build second" to "search by build first, filter by pattern second".

 A build typically has 50–500 artifacts. A broad pattern can match millions of files. So searching the build first and then filtering by pattern is dramatically faster.

**What we changed:**

  1. Routing — When the spec has both build and pattern, we now route it to the BUILD strategy instead of WILDCARD
  2. Pattern filter — After the BUILD strategy fetches the build's artifacts (small set), we added a step that filters those results to only include files matching the pattern
  3. AQL over dedicated API — When a pattern is present, we use build-scoped AQL (artifact.module.build.name = "X") instead of the dedicated build-artifacts REST API, because the REST API has indexing delays for recently-published builds

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -198,7 +198,7 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
-replace github.com/jfrog/jfrog-client-go => github.com/reshmifrog/jfrog-client-go v1.52.1-0.20260412212001-56962407f470
+replace github.com/jfrog/jfrog-client-go => github.com/reshmifrog/jfrog-client-go v1.52.1-0.20260412222345-fd75f1bb2fc4
 
 // replace github.com/jfrog/build-info-go => github.com/reshmifrog/build-info-go v1.10.11-0.20260303032831-71878c7210bf
 
--- a/go.mod
+++ b/go.mod
@@ -198,7 +198,7 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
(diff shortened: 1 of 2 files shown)


naveenku-jfrog (changes requested):
Resolve conflicts

naveenku-jfrog (approved):

PR #416, opened 2026-04-10
fix/reference-token-privilege-escalation

reshmifrog:
-----
Problem: When a user configures JFrog CLI with a restricted reference token (e.g Viewer scope) as the password, the CLI automatically creates a new token using the deprecated Artifactory API (POST /api/security/token) with scope=member-of-groups:*. This deprecated API does not enforce scope ceiling, so it issues a fully privileged token escalating from read-only to full delete permissions.
Solution: Migrated the CLI's token creation from the deprecated Artifactory API to the current Access API (POST /access/api/v1/tokens) with scope=applied-permissions/user. The Access API enforces server-side scope ceiling if the authenticating credential has restricted scope, the server rejects the request with 403 Forbidden, preventing privilege escalation at its root.

### DIFF
--- a/.github/tools/detect-deps/main.go
+++ b/.github/tools/detect-deps/main.go
@@ -121,9 +121,9 @@ func detectDependency(name, modulePath string, replaces map[string]Replace, curr
 		// Parse the replace target
 		newPath := replace.New.Path
 		if strings.HasPrefix(newPath, "github.com/") {
-			// Extract repo and version/ref
+			// Extract repo and version/ref, stripping Go module major version suffixes (e.g. /v2)
 			parts := strings.TrimPrefix(newPath, "github.com/")
-			repo := parts
+			repo := stripMajorVersionSuffix(parts)
 			ref := replace.New.Version
 
 			if ref != "" {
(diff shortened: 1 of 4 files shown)


.github/tools/detect-deps/main.go
fluxxBot, line 1: why do we need this change? can you add a example here?

.github/workflows/cli-integration-tests.yml
```diff
@@ -6,8 +6,7 @@ on:
     branches:
       - main
 
-  pull_request_target:
-    types: [labeled]
```
fluxxBot, line 10: why?

fluxxBot (approved):

.github/workflows/cli-integration-tests.yml
reshmifrog: To trigger the docker tests , after adding the jfrog platform in the docker test suite , reverting it.

.github/tools/detect-deps/main.go
reshmifrog: Go modules with major version v2+ use a /v2 suffix in their module path (e.g., github.com/jfrog/jfrog-cli-core/v2), but the actual GitHub repository name is just jfrog/jfrog-cli-core there's no /v2 in the URL. Without stripping this suffix, the detect-deps tool constructs invalid GitHub API URLs like github.com/reshmifrog/jfrog-cli-core/v2, which fail when trying to resolve branches or commit SHAs. This fix strips the /vN suffix so the tool builds correct GitHub URLs (e.g., reshmifrog/jfrog-cli-core), matching how GitHub repos are actually named.

itsmeleela (approved):

.github/tools/detect-deps/main.go
itsmeleela: Why this was never an issue so far? Because these URLs are like this from long time right?

itsmeleela (approved):

PR #421, opened 2026-04-15
RTECO-992 - Added support for uv package manager

agrasth:
-----
Adds `NativeUVCommand` — a transparent `uv` proxy with Artifactory credential injection and build-info collection, following the same pattern as `ConanCommand`.

### Credential injection (auth priority chain)
Mirrors UV's own priority order exactly — jf injects credentials only as a last resort when no native UV auth mechanism is present:
1. `UV_INDEX_*_USERNAME` env var (native, step 2)
2. Credentials embedded in index URL (native, step 3)
3. `~/.netrc` entry for the index host (native, step 4)
4. Same-host index credentials reused for publish (step 4b — jf-specific)
5. jf server config credentials — injected only when index/publish URL host matches the configured server (fallback)

`UV_DEFAULT_INDEX` and `UV_INDEX_URL` are both recognised as global index env vars. When `--server-id` is explicitly passed, the host-mismatch check is bypassed so credentials are injected regardless of URL hostname.

### Build-info collection
- **Dependencies:** collected from `uv.lock` via `UVFlexPack`; `uv pip list --format=json` used as ground truth after `sync`/`install`/`add`/`remove` — correctly handles all dev/group flag combinations
- **Artifacts:** populated only after `jf uv publish` confirms upload to Artifactory (AQL path lookup); `jf uv build` alone does not add artifacts to build-info
- **AQL enrichment:** single batched `$or` query using `name-version*` + `name_version*` wildcards — one HTTP call for all deps; direct URL and git deps skip AQL with a clear log
- **Direct URL deps:** `sha1`/`md5` computed by streaming the HTTPS URL; git deps logged as not enrichable
- **Help flags:** `jf uv publish -h` and similar bypass all auth and Artifactory calls to prevent credential exposure in uv's help output
- **`--script` support:** `jf uv run --script <path>` collects build-info from the script's adjacent `<script>.py.lock` (created via `uv lock --script` if absent); inline `[[tool.uv.index]]` in PEP 723 metadata is parsed for credential injection

### DIFF
--- a/artifactory/commands/python/native_uv.go
+++ b/artifactory/commands/python/native_uv.go
@@ -19,6 +19,7 @@ import (
 	coreConfig "github.com/jfrog/jfrog-cli-core/v2/utils/config"
 	"github.com/jfrog/jfrog-client-go/artifactory/services"
 	specutils "github.com/jfrog/jfrog-client-go/artifactory/services/utils"
+	"github.com/jfrog/jfrog-client-go/auth"
 	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
@@ -76,12 +77,17 @@ func (c *NativeUvCommand) Run() error {
 		return fmt.Errorf("failed to get working directory: %w", err)
 	}
 
+	// Parse pyproject.toml once and reuse across the entire Run invocation.
+	pyproject := parseUvPyproject(workingDir)
+
 	// Resolve publish URL: explicit arg > pyproject.toml
 	deployerRepo := c.deployerRepo
+	// use a local args copy so we don't mutate the receiver
+	args := append([]string(nil), c.args...)
 	if c.commandName == "publish" && deployerRepo == "" {
-		if tomlURL := uvPublishURLFromToml(workingDir); tomlURL != "" {
+		if tomlURL := pyproject.Tool.Uv.PublishURL; tomlURL != "" {
 			deployerRepo = tomlURL
-			c.args = append(c.args, "--publish-url", tomlURL)
+			args = append(args, "--publish-url", tomlURL)
 			log.Info("Using publish URL from pyproject.toml [tool.uv]: " + tomlURL)
 		}
 	}
@@ -94,7 +100,7 @@ func (c *NativeUvCommand) Run() error {
 	}
 
 	log.Info(fmt.Sprintf("Running UV %s.", c.commandName))
-	if err := runUvBinary(append([]string{c.commandName}, c.args...)); err != nil {
+	if err := runUvBinary(append([]string{c.commandName}, args...)); err != nil {
 		return fmt.Errorf("uv %s failed: %w", c.commandName, err)
 	}
 
@@ -114,7 +120,10 @@ func (c *NativeUvCommand) Run() error {
 func (c *NativeUvCommand) injectCredentials(workingDir, deployerRepo string, serverDetails *coreConfig.ServerDetails) {
 	user := serverDetails.User
 	pass := serverDetails.Password
-	if pass == "" {
+	if serverDetails.AccessToken != "" {
+		if user == "" {
+			user = auth.ExtractUsernameFromAccessToken(serverDetails.AccessToken)
+		}
 		pass = serverDetails.AccessToken
 	}
 	if user == "" || pass == "" {
@@ -123,20 +132,19 @@ func (c *NativeUvCommand) injectCredentials(workingDir, deployerRepo string, ser
 
 	injectedAny := false
 
+	// UV_INDEX_URL is UV's legacy global index env var. We log it for visibility but
+	// still proceed with per-index injection below, because UV_INDEX_URL only affects
+	// the default index and does not supply credentials for named [[tool.uv.index]] entries.
 	if os.Getenv("UV_INDEX_URL") != "" {
-		log.Info("UV auth: UV_INDEX_URL is set globally (native)")
+		log.Info("UV auth: UV_INDEX_URL is set globally (native); per-index credential injection still proceeds below")
 	}
 
 	for _, idx := range uvReadIndexesFromToml(workingDir) {
 		envName := uvIndexEnvName(idx.Name)
 		userKey := "UV_INDEX_" + envName + "_USERNAME"
 		switch {
-		case os.Getenv(userKey) != "":
-			log.Info(fmt.Sprintf("UV auth [index %q]: using env var %s (native, step 2)", idx.Name, userKey))
-		case uvURLHasEmbeddedCredentials(idx.URL):
-			log.Info(fmt.Sprintf("UV auth [index %q]: using credentials embedded in URL (native, step 3)", idx.Name))
-		case uvNetrcHasCredentials(idx.URL):
-			log.Info(fmt.Sprintf("UV auth [index %q]: using %s (native, step 4)", idx.Name, uvNetrcPath()))
+		case uvIndexHasNativeCredentials(idx.URL, userKey):
+			log.Info(fmt.Sprintf("UV auth [index %q]: native credentials already present (env var, embedded URL, or netrc)", idx.Name))
 		default:
 			if uvHostMatchesServer(idx.URL, serverDetails.ArtifactoryUrl) {
 				os.Setenv(userKey, user)
@@ -171,24 +179,11 @@ func runUvBinary(args []string) error {
 	return cmd.Run()
 }
 
-// uvGetCommandName strips the first non-flag argument from a slice and returns
-// (commandName, remainingArgs).
-func uvGetCommandName(orgArgs []string) (string, []string) {
-	cmdArgs := make([]string, len(orgArgs))
-	copy(cmdArgs, orgArgs)
-	for i, arg := range cmdArgs {
-		if !strings.HasPrefix(arg, "-") {
-			return arg, append(cmdArgs[:i], cmdArgs[i+1:]...)
-		}
-	}
-	return "", cmdArgs
-}
-
 // ── TOML types ───────────────────────────────────────────────────────────────
 
 type uvIndexEntry struct {
-	Name string
-	URL  string
+	Name string `toml:"name"`
+	URL  string `toml:"url"`
 }
 
 type uvToolUv struct {
@@ -438,7 +433,7 @@ func uvGetBuildInfo(workingDir string, buildConfiguration *buildUtils.BuildConfi
 					sd, _ = coreConfig.GetDefaultServerConf()
 				}
 				if sd != nil {
-					if indexURL := uvIndexURLFromToml(workingDir); indexURL != "" && !uvServerHostMatches(indexURL, sd.ArtifactoryUrl) {
+					if indexURL := uvIndexURLFromToml(workingDir); indexURL != "" && !uvHostMatchesServer(indexURL, sd.ArtifactoryUrl) {
 						log.Warn(fmt.Sprintf(
 							"UV build-info: jf server config host (%s) differs from index URL host (%s) — "+
 								"dependency checksum enrichment (sha1/md5) will be skipped. "+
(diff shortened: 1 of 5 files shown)


artifactory/commands/python/uv.go
```diff
@@ -0,0 +1,74 @@
...
+
+type UvCommand struct {
+	PythonCommand
+}
+
+func NewUvCommand() *UvCommand {
+	return &UvCommand{PythonCommand: *NewPythonCommand(pythonutils.UV)}
+}
+
+func (pc *UvCommand) UpdateDepsChecksumInfoFunc(dependenciesMap map[string]entities.Dependency, srcPath string) error {
```
bhanurp, line 25: this repo should have most of the business logic, but jfrog cli looks to be having majority can you check other packages and make required changes?

agrasth: updated it.

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+	return c
+}
+
+func (c *NativeUVCommand) SetBuildConfiguration(bc *buildUtils.BuildConfiguration) *NativeUVCommand {
+	c.buildConfiguration = bc
+	return c
+}
+
+func (c *NativeUVCommand) CommandName() string {
+	return "rt_uv_native"
```
naveenku-jfrog, line 69: Why you are not returning c.commandName ?

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+	deployerRepo       string
+	buildConfiguration *buildUtils.BuildConfiguration
+}
+
+// NewNativeUVCommand creates a new NativeUVCommand instance.
+func NewNativeUVCommand() *NativeUVCommand {
+	return &NativeUVCommand{}
+}
+
+func (c *NativeUVCommand) SetCommandName(name string) *NativeUVCommand {
```
naveenku-jfrog, line 43: Can we use something like nuvc or uvc inplace of c?

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+
+func (c *NativeUVCommand) ServerDetails() (*coreConfig.ServerDetails, error) {
+	return uvResolveServerDetails(c.serverID)
+}
+
+// Run executes the UV command with auth injection and optional build info collection.
+func (c *NativeUVCommand) Run() error {
+	// Help flags (-h / --help) and the bare "help" sub-command must bypass all auth
+	// injection and Artifactory calls. Injecting credential env vars before calling uv
+	// causes uv to print them in its help output, exposing secrets.
```
naveenku-jfrog, line 80: Remove these unncessary comments.

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+
+	// Parse pyproject.toml once and reuse across the entire Run invocation.
+	pyproject := parseUvPyproject(workingDir)
+
+	// Resolve publish URL: explicit arg > pyproject.toml
+	deployerRepo := c.deployerRepo
+	// use a local args copy so we don't mutate the receiver
+	args := append([]string(nil), c.args...)
+	if c.commandName == "publish" && deployerRepo == "" {
+		if tomlURL := pyproject.Tool.Uv.PublishURL; tomlURL != "" {
```
naveenku-jfrog, line 98: What is pyproject is nil or pyproject.Tool is nil?

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+	}
+
+	// Use script inline indexes when --script is active; fall back to pyproject.toml entries.
+	indexes := scriptIndexes
+	if len(indexes) == 0 {
+		indexes = uvReadIndexesFromToml(workingDir)
+	}
+	for _, idx := range indexes {
+		envName := uvIndexEnvName(idx.Name)
+		userKey := "UV_INDEX_" + envName + "_USERNAME"
```
naveenku-jfrog: Can we store these hardcoded string as constants?

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+		switch {
```
naveenku-jfrog: There are only 2 conditions, can we remove the switch statement?

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+	return coreConfig.GetSpecificConfig(serverID, true, true)
+}
+
+// uvApplyPublishAuth selects and applies publish credentials following UV's priority chain.
+// explicitServerID is true when the user passed --server-id explicitly; in that case the
+// host-mismatch check is bypassed so credentials are always injected for the publish URL.
+func uvApplyPublishAuth(publishURL, workingDir string, serverDetails *coreConfig.ServerDetails, user, pass string, explicitServerID bool) {
+	switch {
+	case os.Getenv("UV_PUBLISH_TOKEN") != "":
+		log.Info("UV auth [publish]: using UV_PUBLISH_TOKEN (native, step 2a)")
```
naveenku-jfrog: Why these 2a, 2b lines are getting printed in logs.

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+
+// uvEnrichDepsFromArtifactory fetches sha1/md5 for all dependencies in a single batched AQL call.
+//
+// Dep IDs are in "name:version" format; actual filenames use "name-version[-platform].whl" or
+// "name-version.tar.gz" with either hyphens or underscores in the name. We search both variants
+// so platform-specific wheels (e.g. macos arm64) and universal wheels both resolve correctly.
+// uvEnrichDirectURLChecksums computes sha1/md5 for direct URL deps (source = { url = "..." })
+// by streaming the file from its URL. Git deps (source = { git = "..." }) are skipped because
+// rebuilding from source would be required to produce a deterministic archive.
+// If a URL is inaccessible (network error, auth required) the dep retains sha256-only — no error.
```
naveenku-jfrog: Simplify these comments.

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+				Checksum: localArtifact.Checksum,
+			})
+			break
+		}
+		if err := searchReader.Close(); err != nil {
+			log.Warn("Failed to close search reader:", err)
+		}
+	}
+
+	bi.Modules[0].Artifacts = artifacts
```
naveenku-jfrog, line 1085: What is modules is nil? in this case accessing modules[0] will fail the program.

artifactory/commands/python/native_uv.go
```diff
@@ -0,0 +1,1129 @@
...
+
+func runUvBinary(args []string) error {
+	cmd := exec.Command("uv", args...)
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+// ── TOML types ───────────────────────────────────────────────────────────────
```
naveenku-jfrog, line 414: What is this comment saying here?

naveenku-jfrog (changes requested):

artifactory/commands/python/native_uv.go
agrasth: `CommandName()` returns the internal build-system identifier (`"rt_uv_native"`) that jfrog-cli-core uses to route and identify commands — it is distinct from the uv subcommand name (sync, lock, etc.) stored in `c.commandName`. Returning `c.commandName` here would expose the uv subcommand to jfrog-cli-core routing logic, which is incorrect.

artifactory/commands/python/native_uv.go
agrasth: `c` is the idiomatic Go convention for command struct receivers — it is short, unambiguous in context, and consistent with how jfrog-cli and jfrog-cli-core define all other commands. Introducing `uvc` or `nuvc` would deviate from the project-wide pattern without adding clarity.

artifactory/commands/python/native_uv.go
agrasth: The comment explains a non-obvious security constraint: injecting credential env vars before calling `uv -h` causes uv to print them in its help output. Without this comment, a future reader removing the guard would inadvertently re-introduce credential exposure. Keeping it.

artifactory/commands/python/native_uv.go
agrasth: `parseUvPyproject` returns a zero-value `uvPyprojectToml` struct on any error (missing file, parse error). Since all fields are embedded structs (not pointers), accessing `pyproject.Tool.Uv.PublishURL` on a zero-value struct returns an empty string safely — no nil dereference is possible.

artifactory/commands/python/native_uv.go
agrasth: Extracted as helper functions `uvIndexUsernameKey(envName)` and `uvIndexPasswordKey(envName)` — avoids string concatenation repetition and makes the env var naming scheme explicit in one place.

artifactory/commands/python/native_uv.go
agrasth: Replaced the switch with `if/else`. The switch had only 2 branches (case + default) so if/else is cleaner.

artifactory/commands/python/native_uv.go
agrasth: Lowered to `log.Debug` — these are internal auth priority chain details (UV's own native mechanisms, steps 2a/2b). Only the fallback cases where jf actually injects credentials remain at Info level, which is the actionable signal for users.

artifactory/commands/python/native_uv.go
agrasth: Simplified to a single-line doc comment: `uvEnrichDirectURLChecksums fetches sha1/md5 for direct-URL deps by streaming each file. Git deps are skipped (no downloadable archive). Unreachable URLs retain sha256-only.`

artifactory/commands/python/native_uv.go
agrasth: Added a defensive `len(bi.Modules) == 0` check before the `Modules[0]` access, with a comment pointing to the existing guard at function entry. The function-entry guard already returns an error for empty modules, but the defensive check makes it explicit at the point of access.

artifactory/commands/python/native_uv.go
agrasth: That's the `// ── TOML types ──` section-header divider that groups the TOML struct definitions together. It's a visual separator common in larger Go files to make the code navigable. Happy to remove it if it causes confusion.

artifactory/commands/python/native_uv.go
naveenku-jfrog: But then what's the use of commandName() then?

naveenku-jfrog (dismissed):

artifactory/commands/python/native_uv.go
agrasth: `CommandName()` is required by the `commands.Command` interface in jfrog-cli-core and is used for **usage telemetry** — it is passed to `CollectMetrics()` and `reportUsage()` (CallHome) so JFrog Analytics can track which command types users are running. We return the stable string `"rt_uv_native"` rather than the dynamic uv subcommand (sync/lock/etc.) because the interface expects a fixed command-type identifier, not a per-invocation value. Every other command follows the same pattern: `MvnCommand` returns `"rt_mvn_command"`, `ConanCommand` returns `"rt_conan"`, etc.

bhanurp (dismissed):

naveenku-jfrog (dismissed):

bhanurp (approved):

naveenku-jfrog (approved):

PR #427, opened 2026-04-20
fix(helm): separate build-info modules by type

### LINKED ISSUE #426: Helm OCI build-info modules can merge into Docker modules when ids match
## Describe the bug

When collecting build-info for Helm OCI operations, Helm modules can be incorrectly merged into Docker modules if both modules share the same `name:version`.

The issue is caused by Helm build-info merge helpers using only `module.Id` to identify an existing module. In mixed-package builds, that assumption is too weak because modules from different ecosystems can legitimately have the same ID while still representing different module types.

## Current behavior

If a Docker module already exists in the build-info with the same `name:version` as a Helm OCI module, Helm artifacts and/or module data can be merged into the Docker module instead of remaining in a separate Helm module.

This leads to incorrect build-info output and order-dependent behavior.

Example of the problematic flow:

- Docker produces a module with ID `my-app:1.0.0`
- Helm OCI later produces a module with the same ID `my-app:1.0.0`
- Helm build-info helpers match by `Id` only
- Helm artifacts are appended to the Docker module or the module merge result depends on insertion order

Observed impact:

- build-info can contain merged module content from different package types
- the final output is inconsistent depending on whether Docker or Helm is processed first
- the generated build-info does not accurately represent the published artifacts

## Reproduction steps

One reproducible scenario is:

1. Start with build-info that already contains a Docker module with:
   - `Type = docker`
   - `Id = my-app:1.0.0`
2. Run a Helm OCI flow that produces a Helm module with:
   - `Type = helm`
   - `Id = my-app:1.0.0`
3. Let the Helm helpers append artifacts or merge modules into the existing build-info
4. Inspect the resulting modules in the saved build-info

The issue can be reproduced directly in helper-level tests around:

- `addArtifactsInBuildInfo`
- `appendModuleInExistingBuildInfo`

## Expected behavior

Modules should be merged only when both the module type and the module ID match.

That means:

- same `Type` + same `Id` -> merge
- different `Type` + same `Id` -> stay separate

The resulting build-info should be stable and independent of processing order.

## JFrog CLI-Core version

v2.60.1-0.20260414083544-243b4d55328b

## JFrog CLI-Artifactory version

Current checkout from this repository based on:

github.com/jfrog/jfrog-cli-artifactory

## JFrog CLI version (if applicable)

jf version 2.100.0

## Operating system type and version

macOS darwin arm64

## JFrog Artifactory version

N.A

## Additional context

- This report is intentionally focused on the bug itself.
- The bug is in build-info module merge semantics, not in Artifactory storage-path resolution.
- A correct fix should merge modules only when both `Type` and `Id` match.

vjda:
-----

Fixes #426

## Summary

This PR fixes the Helm OCI build-info bug reported in #426, where
Helm and Docker modules could be merged together when they shared the same
`name:version`.

The linked issue contains the full bug report and reproduction details. In
short, Helm build-info helpers were matching modules by `Id` only, which made
cross-type collisions possible in mixed Helm/Docker build-info flows.

## Proposed Fix

This change makes module matching use `(Type, Id)` instead of `Id` alone.

### Implementation details

- `addArtifactsInBuildInfo` now receives the expected module type and only
  appends artifacts when both `module.Type` and `module.Id` match
- `appendModuleInExistingBuildInfo` now merges modules only when both fields
  match
- the Helm push flow now passes `entities.ModuleType("helm")` explicitly when
  adding artifacts into build-info

This preserves the external module ID format while correcting the internal
merge semantics.

## Why this approach

Using `(Type, Id)` as the merge identity is the smaller and safer fix because:

- it addresses the real collision source directly
- it preserves existing output format expectations
- it keeps the change localized and avoids broader refactoring

## Test Coverage

Added table-driven tests covering:

- same ID + same type -> modules still merge
- same ID + different type -> modules stay separate
- Docker -> Helm order independence
- Helm -> Docker order independence
- empty type matching semantics

Validation run locally:

- `go test -v ./artifactory/commands/helm/...`
- `go vet ./artifactory/commands/helm/...`

## Files changed

| File | Change |
|------|--------|
| `artifactory/commands/helm/utils.go` | Match build-info modules by `(Type, Id)` in both merge helpers |
| `artifactory/commands/helm/push.go` | Pass Helm module type explicitly when adding pushed artifacts |
| `artifactory/commands/helm/utils_test.go` | Add regression coverage for cross-type separation and order independence |

## Notes for Reviewers

- This PR intentionally does not change the external module ID format
- The fix is scoped to Helm build-info merge behavior only
- The goal is to preserve existing same-type merge behavior while preventing
  Helm/Docker cross-type collisions

### DIFF
--- a/artifactory/commands/helm/push.go
+++ b/artifactory/commands/helm/push.go
@@ -64,7 +64,7 @@ func handlePushCommand(buildInfo *entities.BuildInfo, helmArgs []string, service
 	for _, artLayer := range artifactsLayers {
 		artifacts = append(artifacts, artLayer.ToArtifact())
 	}
-	addArtifactsInBuildInfo(buildInfo, artifacts, chartName, chartVersion)
+	addArtifactsInBuildInfo(buildInfo, artifacts, chartName, chartVersion, entities.ModuleType("helm"))
 	removeDuplicateArtifacts(buildInfo)
 	return saveBuildInfo(buildInfo, buildName, buildNumber, project)
 }
--- a/artifactory/commands/helm/utils.go
+++ b/artifactory/commands/helm/utils.go
@@ -151,13 +151,13 @@ func removeDuplicateDependencies(buildInfo *entities.BuildInfo) {
 	}
 }
 
-func addArtifactsInBuildInfo(buildInfo *entities.BuildInfo, artifacts []entities.Artifact, chartName, chartVersion string) {
+func addArtifactsInBuildInfo(buildInfo *entities.BuildInfo, artifacts []entities.Artifact, chartName, chartVersion string, moduleType entities.ModuleType) {
 	if buildInfo == nil {
 		return
 	}
 	moduleId := fmt.Sprintf("%s:%s", chartName, chartVersion)
 	for moduleIdx, module := range buildInfo.Modules {
-		if module.Id == moduleId {
+		if module.Type == moduleType && module.Id == moduleId {
 			module.Artifacts = append(module.Artifacts, artifacts...)
 			buildInfo.Modules[moduleIdx] = module
 		}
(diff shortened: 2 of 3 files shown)


vjda:
Addressed both copilot's comments that point to the same real gap.

The new `(Type, Id)` matching is still the correct fix for the original bug,
but after tightening the merge identity, the flow also needs to guarantee that
the target Helm module exists before appending artifacts.

The follow-up change fixes that in `appendModuleAndBuildAgentIfAbsent(...)`:
instead of creating a Helm module only when `buildInfo.Modules` is empty, it now
ensures that the expected Helm module `(Type=helm, Id=chartName:chartVersion)`
exists.

I also added focused regression coverage for the mixed-module scenarios.

PR #428, opened 2026-04-20
RTECO-1052 - Docker verification customisable through flags

fluxxBot:
-----

Docker image id verification: fix OOM, make slow path opt-out
PR #410 introduced daemon.WithUnbufferedOpener() in containerManager.Id() to stop the Docker push flow from OOMing on large images — daemon.Image(ref) buffers the entire docker save tarball in memory. That change fixed the OOM but caused a second regression: on Docker 29 (which most customers now run) and on resource-constrained agents, the unbuffered opener stalls the push for 5+ minutes on multi-GB images. During that stall the jf docker push appears hung, sitting between the Pushed line from the Docker daemon and the subsequent Setting properties... build-info step.

The root cause is daemon.Image(...).Manifest() invoking a full docker save over the whole image just to read the config digest. On Docker 29 the docker images --format {{.ID}} fallback cannot be used because it now reports the manifest digest, not the config digest, so we can't cheaply substitute it.

What this PR does?
Two changes, both scoped to the Docker-client path inside containerManager.Id():

Replace WithUnbufferedOpener with WithFileBufferedOpener — spools the docker save stream to a temp file instead of RAM, permanently removing the OOM risk without needing any flag. Memory usage stays bounded regardless of image size.
Add a Push-only opt-in to skip the local id lookup entirely via a new env var JFROG_CLI_SKIP_DOCKER_IMAGE_ID_VERIFICATION. When set to a truthy value on a docker push, Id() short-circuits and returns an empty string; the build-info builder then adopts the Artifactory-side manifest's Config.Digest for layer lookups, bypassing docker save completely. Pull and all other flows ignore the flag and always verify.
Default behavior is unchanged: we still verify the local id, just with file-backed buffering instead of in-memory buffering.

Depends on:
1. https://github.com/jfrog/jfrog-cli/pull/3447

### DIFF
--- a/artifactory/commands/ocicontainer/containermanager.go
+++ b/artifactory/commands/ocicontainer/containermanager.go
@@ -27,13 +27,13 @@ const MinSupportedApiVersion string = "1.31"
 // Docker login error message
 const LoginFailureMessage string = "%s login failed for: %s.\n%s image must be in the form: registry-domain/path-in-repository/image-name:version."
 
-// SkipDockerImageIdVerificationEnv, when set to "true", tells the CLI to skip
-// the local docker image id lookup (which internally triggers `docker save`
-// on the entire image tarball to read the config digest) and the subsequent
-// manifest verification. Intended for large images where the lookup becomes
-// prohibitively slow. The Artifactory-side manifest's Config.Digest is used
+// EnforceDockerImageIdVerificationEnv, when set to "true", re-enables the
+// local docker image id lookup (which internally triggers `docker save` on the
+// entire image tarball to read the config digest) and the subsequent manifest
+// verification. By default this step is skipped because it is prohibitively
+// slow on large images; the Artifactory-side manifest's Config.Digest is used
 // in place of the local value for build-info construction.
-const SkipDockerImageIdVerificationEnv = "JFROG_CLI_SKIP_DOCKER_IMAGE_ID_VERIFICATION"
+const EnforceDockerImageIdVerificationEnv = "JFROG_CLI_ENFORCE_DOCKER_IMAGE_ID_VERIFICATION"
 
 func NewManager(containerManagerType ContainerManagerType) ContainerManager {
 	return &containerManager{Type: containerManagerType}
@@ -76,12 +76,14 @@ func (containerManager *containerManager) RunNativeCmd(cmdParams []string) error
 // Get image ID
 func (containerManager *containerManager) Id(image *Image) (string, error) {
 	if containerManager.GetContainerManagerType() == DockerClient {
-		// Opt-out: when this env var is set, skip the expensive daemon.Image
-		// call (which triggers `docker save` over the whole image) and let the
-		// caller fall back to the config digest reported by Artifactory's manifest.
+		// By default, skip the expensive daemon.Image call (which triggers
+		// `docker save` over the whole image) and let the caller fall back to
+		// the config digest reported by Artifactory's manifest. Set
+		// JFROG_CLI_ENFORCE_DOCKER_IMAGE_ID_VERIFICATION=true to opt back in
+		// to the slow but strict local verification path.
 		// See isVerifiedManifest / buildInfoBuilder for the matching consumer logic.
-		if strings.EqualFold(os.Getenv(SkipDockerImageIdVerificationEnv), "true") {
-			log.Debug(SkipDockerImageIdVerificationEnv + "=true -> skipping local image id lookup and verification")
+		if !strings.EqualFold(os.Getenv(EnforceDockerImageIdVerificationEnv), "true") {
+			log.Debug("Skipping local image id lookup and verification (set " + EnforceDockerImageIdVerificationEnv + "=true to enforce)")
 			return "", nil
 		}
 		ref, err := name.ParseReference(image.Name())
--- a/artifactory/commands/ocicontainer/containermanager.go
+++ b/artifactory/commands/ocicontainer/containermanager.go
@@ -27,6 +27,14 @@ const MinSupportedApiVersion string = "1.31"
 // Docker login error message
 const LoginFailureMessage string = "%s login failed for: %s.\n%s image must be in the form: registry-domain/path-in-repository/image-name:version."
 
+// SkipDockerImageIdVerificationEnv, when set to "true", tells the CLI to skip
+// the local docker image id lookup (which internally triggers `docker save`
+// on the entire image tarball to read the config digest) and the subsequent
+// manifest verification. Intended for large images where the lookup becomes
+// prohibitively slow. The Artifactory-side manifest's Config.Digest is used
+// in place of the local value for build-info construction.
+const SkipDockerImageIdVerificationEnv = "JFROG_CLI_SKIP_DOCKER_IMAGE_ID_VERIFICATION"
+
 func NewManager(containerManagerType ContainerManagerType) ContainerManager {
 	return &containerManager{Type: containerManagerType}
 }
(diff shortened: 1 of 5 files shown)


artifactory/commands/ocicontainer/containermanager.go
```diff
@@ -68,6 +76,16 @@ func (containerManager *containerManager) RunNativeCmd(cmdParams []string) error
 // Get image ID
 func (containerManager *containerManager) Id(image *Image) (string, error) {
 	if containerManager.GetContainerManagerType() == DockerClient {
+		// By default, skip the expensive daemon.Image call (which triggers
+		// `docker save` over the whole image) and let the caller fall back to
+		// the config digest reported by Artifactory's manifest. Set
+		// JFROG_CLI_ENFORCE_DOCKER_IMAGE_ID_VERIFICATION=true to opt back in
+		// to the slow but strict local verification path.
+		// See isVerifiedManifest / buildInfoBuilder for the matching consumer logic.
+		if !strings.EqualFold(os.Getenv(EnforceDockerImageIdVerificationEnv), "true") {
```
bhanurp: parse the bool value

bhanurp: lets make this change is only applicable for commandType Push
  
    
      
        Suggested change
      
    
    
      
          
            
            		if !strings.EqualFold(os.Getenv(EnforceDockerImageIdVerificationEnv), "true") {
          
          
            
            		if strings.EqualFold(os.Getenv(SkipDockerImageIdVerificationEnv), "true") {

artifactory/commands/ocicontainer/containermanager.go
```diff
@@ -27,6 +27,14 @@ const MinSupportedApiVersion string = "1.31"
 // Docker login error message
 const LoginFailureMessage string = "%s login failed for: %s.\n%s image must be in the form: registry-domain/path-in-repository/image-name:version."
 
+// EnforceDockerImageIdVerificationEnv, when set to "true", re-enables the
+// local docker image id lookup (which internally triggers `docker save` on the
+// entire image tarball to read the config digest) and the subsequent manifest
+// verification. By default this step is skipped because it is prohibitively
+// slow on large images; the Artifactory-side manifest's Config.Digest is used
+// in place of the local value for build-info construction.
+const EnforceDockerImageIdVerificationEnv = "JFROG_CLI_ENFORCE_DOCKER_IMAGE_ID_VERIFICATION"
```
bhanurp: lets make it opt in instead of opt out. 

```suggestion
const SkipDockerImageIdVerificationEnv = "JFROG_CLI_SKIP_DOCKER_IMAGE_ID_VERIFICATION"
```

artifactory/commands/ocicontainer/containermanager.go
bhanurp: it should be `WithFileBufferedOpener`
```suggestion
		localImage, err := daemon.Image(ref, daemon.WithFileBufferedOpener())
```

artifactory/commands/ocicontainer/containermanager_test.go
```diff
@@ -1,86 +1,33 @@
...
+// By default (env var unset) Id() must short-circuit, returning "" without
+// invoking daemon.Image / docker save. This is the performance fix for #3382.
+// The test does NOT require a running Docker daemon — proof that no daemon
+// call is made.
+func TestDockerClientId_DefaultSkipsDaemon(t *testing.T) {
+	// Force the env var to "false" to protect against a caller having exported
+	// it from the shell.
+	t.Setenv(EnforceDockerImageIdVerificationEnv, "false")
 
-func getImageSizeMB(t *testing.T, imageName string) uint64 {
```
bhanurp, line 21: why are the tests removed?

artifactory/commands/ocicontainer/containermanager_test.go
bhanurp, line 1: Add tests for Id when env is set and unset.

bhanurp (changes requested):

artifactory/commands/ocicontainer/containermanager.go
itsmeleela: @bhanurp is right that this should be inverted to JFROG_CLI_SKIP_DOCKER_IMAGE_ID_VERIFICATION with if EqualFold(env, "true"). The default behavior stays the same (verification is skipped)

artifactory/commands/ocicontainer/containermanager.go
itsmeleela: Agree on this point too. The OOM issue was reported specifically in the push flow. Skipping the local ID lookup on pulls could silently affect behavior which we are not sure yet

artifactory/commands/ocicontainer/localagent.go
```diff
@@ -168,7 +168,16 @@ func (labib *localAgentBuildInfoBuilder) search(imagePathPattern string) (result
 }
 
 // Verify manifest by comparing sha256, which references to the image digest. If there is no match, return nil.
+// By default the local image id lookup is skipped (imageSha2 is empty) and we
+// accept the Artifactory manifest as-is, adopting its Config.Digest for
+// subsequent build-info lookups. Set
+// JFROG_CLI_ENFORCE_DOCKER_IMAGE_ID_VERIFICATION=true to re-enable strict local verification.
 func (labib *localAgentBuildInfoBuilder) isVerifiedManifest(imageManifest *manifest) bool {
+	if labib.buildInfoBuilder.imageSha2 == "" {
```
itsmeleela, line 180: This function is doing two things right - verifying the manifest AND updating builder state. A function called isVerified***** shouldn't be mutating imageSha2 as a side effect, it's easy to miss and hard to debug. Consider either renaming this to `resolveAndVerifyManifest` or extracting the assignment out to the caller.

artifactory/commands/ocicontainer/containermanager.go
```diff
@@ -68,6 +76,16 @@ func (containerManager *containerManager) RunNativeCmd(cmdParams []string) error
...
 func (containerManager *containerManager) Id(image *Image) (string, error) {
 	if containerManager.GetContainerManagerType() == DockerClient {
+		// By default, skip the expensive daemon.Image call (which triggers
+		// `docker save` over the whole image) and let the caller fall back to
+		// the config digest reported by Artifactory's manifest. Set
+		// JFROG_CLI_ENFORCE_DOCKER_IMAGE_ID_VERIFICATION=true to opt back in
+		// to the slow but strict local verification path.
+		// See isVerifiedManifest / buildInfoBuilder for the matching consumer logic.
+		if !strings.EqualFold(os.Getenv(EnforceDockerImageIdVerificationEnv), "true") {
+			log.Debug("Skipping local image id lookup and verification (set " + EnforceDockerImageIdVerificationEnv + "=true to enforce)")
```
itsmeleela: @bhanurp @fluxxBot - The log here says we're skipping the lookup but doesn't say what digest will be used instead. I feel since this is the path most customers will hit, add the Artifactory digest to the log message so it's useful when debugging. Something like: `"Skipping local image id lookup; will use Artifactory digest: " + <digest>`

WDYT?

artifactory/commands/ocicontainer/containermanager.go
fluxxBot: have added this but IMO FileBufferedOpener comes into picture when we're accesing saved tarball mulitple times, in our case we use it only once.
Also upon writing to a file from the buffer will add little overhead as well

artifactory/commands/ocicontainer/containermanager_test.go
fluxxBot: done

artifactory/commands/ocicontainer/localagent.go
fluxxBot: renamed and added debug log

artifactory/commands/ocicontainer/containermanager.go
fluxxBot: done

artifactory/commands/ocicontainer/containermanager.go
fluxxBot: done

artifactory/commands/ocicontainer/containermanager_test.go
fluxxBot: it was actually combined into one, but I have restored now

artifactory/commands/ocicontainer/containermanager.go
fluxxBot: done

artifactory/commands/ocicontainer/containermanager.go
fluxxBot: have added this, but pulls will still be slower

bhanurp (approved):

PR #431, opened 2026-04-23
RTECO-1055 - Treat pnpm workspace as standalone

fluxxBot:
-----
### Problem

`jf pnpm install` unconditionally ran `pnpm ls -r --depth Infinity --json` when collecting build-info. This ignored the user's workspace intent in two scenarios:

1. **Sub-package invocation** — `cd apps/web-app && jf pnpm install` produced a multi-module build-info containing every workspace package, not just `web-app`.
2. **`--ignore-workspace` flag** — `pnpm install --ignore-workspace` was honored by pnpm during install, but the subsequent `pnpm ls -r` re-enumerated the entire monorepo, leaking sibling/root deps into the build-info.

Downstream SBOM / Xray consumers therefore saw phantom dependencies that didn't actually belong to the built package.

### Fix

The build-info `pnpm ls` call now matches the user's intent:

- **`isPnpmWorkspaceSubPackage`** — uses `pnpm root -w` plus symlink-resolved path compare. When the working directory is a workspace sub-package, `-r` is dropped and only the current package's deps are collected.
- **`extractLsForwardFlags`** — forwards `--ignore-workspace` from the install args to `pnpm ls`, so build-info respects the flag.
- **`buildPnpmLsArgs`** — also drops `-r` when `--ignore-workspace` is forwarded (pnpm would otherwise emit concatenated JSON arrays that aren't parseable by `encoding/json`).

Workspace-root invocation without `--ignore-workspace` keeps the existing `-r` multi-module behavior — no regression for current CI flows.

### Tests

- Unit tests for `extractLsForwardFlags`, `buildPnpmLsArgs` (all branches), `samePath`, `isPnpmWorkspaceSubPackage`.
- Companion E2E tests in `jfrog-cli` (linked PR) — verified end-to-end against a real Artifactory instance.

depends on:
1. https://github.com/jfrog/jfrog-cli/pull/3451

### DIFF
--- a/artifactory/commands/pnpm/install.go
+++ b/artifactory/commands/pnpm/install.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/jfrog/build-info-go/build"
@@ -100,8 +102,13 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
 	}
 	log.Debug("Server details. Artifactory URL:", pic.serverDetails.ArtifactoryUrl)
 
+	scopeToCurrentPackage := isPnpmWorkspaceSubPackage(pic.workingDirectory)
+	if scopeToCurrentPackage && userRequestedWorkspaceRoot(pic.pnpmArgs) {
+		log.Debug("--workspace-root/-w specified; collecting build-info for the entire workspace despite sub-package cwd.")
+		scopeToCurrentPackage = false
+	}
 	log.Debug("Running pnpm ls to collect dependency tree...")
-	projects, err := runPnpmLs(pic.workingDirectory)
+	projects, err := runPnpmLs(pic.workingDirectory, extractLsForwardFlags(pic.pnpmArgs), scopeToCurrentPackage)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


artifactory/commands/pnpm/install.go
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+// the caller keeps the existing recursive behavior.
+func isPnpmWorkspaceSubPackage(workingDir string) bool {
+	cmd := exec.Command("pnpm", "root", "-w")
+	cmd.Dir = workingDir
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Debug(fmt.Sprintf("pnpm workspace detection skipped (%s; output: %q); falling back to recursive pnpm ls.", err.Error(), strings.TrimSpace(string(out))))
+		return false
+	}
+	workspaceRoot := filepath.Dir(strings.TrimSpace(string(out)))
```
bhanurp: https://pkg.go.dev/path/filepath#Dir
gives "." as returned string when the path is empty so the check `if workspaceRoot == "" {` becomes invalid

artifactory/commands/pnpm/install.go
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+	return resolvedA == resolvedB
+}
+
+// extractLsForwardFlags returns flags from the user's pnpm install args that must
+// be forwarded to `pnpm ls` so the dependency tree respects the same workspace
+// scope as the install command (e.g. --ignore-workspace).
+func extractLsForwardFlags(installArgs []string) []string {
+	var forwarded []string
+	for _, arg := range installArgs {
+		if arg == "--ignore-workspace" {
```
bhanurp: the scenario when user gives flag as `--ignore-workspace=true` is handled?

artifactory/commands/pnpm/install.go
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+	if workspaceRoot == "" {
+		return false
+	}
+	log.Debug(fmt.Sprintf("Invoked inside workspace sub-package (workspace root: %s); scoping build-info to current package.", workspaceRoot))
+	return true
+}
+
+// samePath reports whether a and b refer to the same directory after resolving
+// symlinks. Falls back to a plain string comparison if either path cannot be
+// resolved (e.g. one side no longer exists on disk).
```
bhanurp: nit```suggestion
// samePath reports whether a and b refer to the same directory after resolving
// symlinks. Falls back to a plain string comparison if either path cannot be
// resolved (e.g. one side no longer exists on disk). Returns false if symlink resolution fails on either side, only direct string equality is used as a fallback.
```

File: artifactory/commands/pnpm/install.go [hunk 4]
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+// but is NOT the workspace root (i.e. the user invoked the command from a single
+// workspace package). When true, build-info should be scoped to that package only.
+//
+// Detection is delegated to pnpm itself via `pnpm root -w`, which prints the
+// workspace root's node_modules path when inside a workspace and errors with a
+// non-zero exit code otherwise. Any error (not a workspace, pnpm missing, etc.)
+// is logged at debug level and treated as "not a sub-package" — in that case
+// the caller keeps the existing recursive behavior.
+func isPnpmWorkspaceSubPackage(workingDir string) bool {
+	cmd := exec.Command("pnpm", "root", "-w")
```
bhanurp, line 134: NIT:
instead of running pnpm root it is a good idea to check pnpm workspace root via go, to avoid extra process spawning

artifactory/commands/pnpm/install.go
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Debug(fmt.Sprintf("pnpm workspace detection skipped (%s; output: %q); falling back to recursive pnpm ls.", err.Error(), strings.TrimSpace(string(out))))
+		return false
+	}
+	workspaceRoot := filepath.Dir(strings.TrimSpace(string(out)))
+	// pnpm resolves symlinks in its output (e.g. on macOS /var → /private/var), so
+	// compare canonicalized paths to avoid false "sub-package" detection when the
+	// working directory happens to contain a symlink.
+	if samePath(workspaceRoot, workingDir) {
```
bhanurp, line 160: check if this scenario is valid, workspaceRoot returns "." and workingDir has a value since they wont match does it treats as sub package exists?

artifactory/commands/pnpm/install.go
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+	args := []string{"ls"}
+	if !scopeToCurrentPackage && !slices.Contains(extraArgs, "--ignore-workspace") {
+		args = append(args, "-r")
+	}
+	args = append(args, "--depth", "Infinity", "--json")
+	args = append(args, extraArgs...)
+	return args
+}
+
+func runPnpmLs(workingDir string, extraArgs []string, scopeToCurrentPackage bool) ([]pnpmLsProject, error) {
```
bhanurp, line 301: nit:
```suggestion
func runPnpmLs(workingDir string, extraArgs []string) ([]pnpmLsProject, error) {
scopeToCurrentPackage := isPnpmWorkspaceSubPackage(workingDir)
```

artifactory/commands/pnpm/install.go
bhanurp: worth checking other flags like --filter, --prod, --dev, --no-optional, --workspace-root etc which can alter the dependencies collected for build info.

artifactory/commands/pnpm/install.go
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+//
+// Detection is delegated to pnpm itself via `pnpm root -w`, which prints the
+// workspace root's node_modules path when inside a workspace and errors with a
+// non-zero exit code otherwise. Any error (not a workspace, pnpm missing, etc.)
+// is logged at debug level and treated as "not a sub-package" — in that case
+// the caller keeps the existing recursive behavior.
+func isPnpmWorkspaceSubPackage(workingDir string) bool {
+	cmd := exec.Command("pnpm", "root", "-w")
+	cmd.Dir = workingDir
+	out, err := cmd.CombinedOutput()
```
bhanurp: any specific reason for reading `combinedOuyput` ? shouldn't it be cmd.Output() ? where we need to read only stdout.

artifactory/commands/pnpm/install.go
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+// isPnpmWorkspaceSubPackage returns true when workingDir is inside a pnpm workspace
+// but is NOT the workspace root (i.e. the user invoked the command from a single
+// workspace package). When true, build-info should be scoped to that package only.
+//
+// Detection is delegated to pnpm itself via `pnpm root -w`, which prints the
+// workspace root's node_modules path when inside a workspace and errors with a
+// non-zero exit code otherwise. Any error (not a workspace, pnpm missing, etc.)
+// is logged at debug level and treated as "not a sub-package" — in that case
+// the caller keeps the existing recursive behavior.
+func isPnpmWorkspaceSubPackage(workingDir string) bool {
```
bhanurp, line 133: do we need this sub package detection?
- scenario A where user goes to sub package but still in a workspace and executes pnpm install it performs install for workspace which means build info should capture what has happened on the CI
- scenario B where user goes to sub package still in a workspace and executes pnpm install it performs install and jfrog cli collects only dependencies for the package in the build-info good for build scan but jfrog cli is not recording all the information happening on CI environment.
we need to make a trade off

bhanurp (changes requested):

artifactory/commands/pnpm/install.go
fluxxBot: yes

artifactory/commands/pnpm/install.go
fluxxBot: We deliberately delegate to pnpm's own logic so we stay compatible with pnpm's workspace resolution rules — which now include pnpm-workspace.yaml, but pnpm has historically extended this (e.g. --workspace-root-dir overrides, packages patterns with negation). Reimplementing it in Go means tracking pnpm releases. The one-time ~80ms subprocess spawn is negligible against the seconds-to-minutes of checksum/AQL fetches that follow. Happy to revisit if you think otherwise.

artifactory/commands/pnpm/install.go
fluxxBot: Kept scopeToCurrentPackage as an explicit parameter so runPnpmLs stays a pure command builder/runner that's unit-testable without invoking pnpm root -w (it has no other reason to spawn pnpm twice). The detection happens once in collectAndSaveBuildInfo, which is the right orchestration level. The unit tests for buildPnpmLsArgs exercise all four (scope × ignore-workspace) combinations cleanly because of this separation. If the goal is fewer params at the call site, I can wrap with a small adapter, but I'd prefer not to fold detection into runPnpmLs.

artifactory/commands/pnpm/install.go
fluxxBot: Customer ask in RTECO-1055 was explicitly scenario B — per-package SBOM from the natural cd apps/web-app && jf pnpm install flow. Their pain: scanning all monorepo deps when only one package was deployed produced noisy/wrong Xray findings. This design preserves both behaviors and lets cwd be the user's signal:

1. workspace root → multi-module (scenario A, current behavior, no regression)
2. sub-package → single-module (scenario B, customer ask)
3. --ignore-workspace → single-module (explicit override, works from anywhere)

artifactory/commands/pnpm/install.go
fluxxBot: I have came across certain places where in stderr only records `exit code 1`, but the actual failure reason lives in stdout, also sometimes entire failure is contained in stderr, so I preferred this path. Any suggestions?

artifactory/commands/pnpm/install.go
fluxxBot: added guard clause for this

artifactory/commands/pnpm/install.go
fluxxBot: done

artifactory/commands/pnpm/install.go
fluxxBot: handled

artifactory/commands/pnpm/install.go
fluxxBot: handled it now

artifactory/commands/pnpm/install.go
bhanurp: the only worry here if there is addition of warnings or deprecations it will be added to stderr and reading combined output will fail since parsing will fail.
Since we are only parsing the output from stdout below is a good approach

you can think of an approach like below

```
var stdout, stderr bytes.Buffer
      cmd.Stdout = &stdout
      cmd.Stderr = &stderr
          log.Debug(fmt.Sprintf(
              "pnpm workspace detection skipped (%s; stdout: %q; stderr: %q); falling back to recursive
  pnpm ls.",
              err.Error(),
              strings.TrimSpace(stdout.String()),
              strings.TrimSpace(stderr.String())))
          return false
      }
      workspaceRoot := filepath.Dir(strings.TrimSpace(stdout.String()))
```

artifactory/commands/pnpm/install.go
bhanurp: what is the claude added?

artifactory/commands/pnpm/install.go
```diff
@@ -114,8 +117,82 @@ func (pic *PnpmInstallCommand) collectAndSaveBuildInfo() error {
...
+	workspaceRoot := filepath.Dir(strings.TrimSpace(string(out)))
+	// pnpm resolves symlinks in its output (e.g. on macOS /var → /private/var), so
+	// compare canonicalized paths to avoid false "sub-package" detection when the
+	// working directory happens to contain a symlink.
+	if samePath(workspaceRoot, workingDir) {
+		log.Debug("Invoked at pnpm workspace root; collecting build-info for all workspace packages.")
+		return false
+	}
+	if workspaceRoot == "" {
+		return false
```
bhanurp: this still become unreachable code since workspaceRoot can never be ""

artifactory/commands/pnpm/install.go
bhanurp: couldn't find other flags is it in different file or PR?

artifactory/commands/pnpm/install.go
bhanurp: to summarize:
user should run `--ignore-workspace` to collect build info specific to sub package and remove `--ignore-workspace` to collect build info related to whole workspace?

fluxxBot: user should run pnpm install in the particular workspace or `--ignore-workspace` in that particular workspace to collect dependencies of the required one, the problem was that -r flag was being appended in the npm ls which collected dependencies of entire repo

artifactory/commands/pnpm/install.go
fluxxBot: check here - https://github.com/jfrog/jfrog-cli-artifactory/pull/431/changes#diff-4cc35a6f447aa77c147813b00535a8148b82775e460ed8152e91c7b8878e3d53R178-R189

artifactory/commands/pnpm/install.go
fluxxBot: done

artifactory/commands/pnpm/install.go
fluxxBot: taken care

PR #432, opened 2026-04-27
RTECO-1015 - Implement jf skills list command

udaykb2:
### Overview
Supports listing skills from Artifactory or from local AI agent skill directories, with filtering and sorting options.

1. Implemented Flags:
--repo: list skills from an Artifactory skills repository.
--agent: list locally installed skills for a given AI agent. cursor / claude-code / github-copilot / windsurf
--project-dir: list project-scoped skills (e.g. .cursor/skills/) inside the given path. Use '.' for cwd. Does not fall back to global.
--limit: cap the number of results.
--sort-by: name, slug, versions, updated (default), created.
--sort-order: asc (default) or desc.
--format: table (default) or json.

2. Agent paths centralised in skills/common/agents.go
3. Handle YAML block scalar indicators (> >- | |-)
4. Hidden the list command from user

-----


cliutils/flagkit/flags.go
```diff
@@ -1166,6 +1175,11 @@ var flagsMap = map[string]components.Flag{
 	propSearch:          components.NewBoolFlag(propSearch, "Use Artifactory property search (skill.name) instead of Skills API search.", components.WithBoolDefaultValueFalse()),
 	skipScan:            components.NewBoolFlag(skipScan, "Skip Xray security scan after publish. Can also be set via JFROG_CLI_SKIP_SKILLS_SCAN=true.", components.WithBoolDefaultValueFalse()),
 	autoDeleteOnFailure: components.NewBoolFlag(autoDeleteOnFailure, "Automatically delete the artifact if Xray scan identifies it as malicious.", components.WithBoolDefaultValueFalse()),
+	skillsAgent:         components.NewStringFlag(skillsAgent, "AI agent name to list locally installed skills for. Supported: claude-code, cursor, github-copilot, windsurf.", components.SetMandatoryFalse()),
+	skillsProjectDir:    components.NewStringFlag("project-dir", "Path to the project root for project-scoped skills. Use '.' for the current directory. Falls back to global agent skills if not found.", components.SetMandatoryFalse()),
+	skillsLimit:         components.NewStringFlag("limit", "Maximum number of skills to return. Fetches all by default.", components.SetMandatoryFalse()),
```
itsmeleela: Why is it a stringflag? 

I feel skillsLimit should be intflag, not string. Using StringFlag defers type validation to inside the command handler, so jf skills list --limit foo will not fail at the flag-parse layer. Compare with how --count or similar numeric flags are defined elsewhere in this file.

skills/commands/list/list_test.go
```diff
@@ -0,0 +1,415 @@
...
+	_, _ = buf.ReadFrom(r)
+
+	var results []listResult
+	require.NoError(t, json.Unmarshal(buf.Bytes(), &results))
+	require.Len(t, results, 3)
+	assert.Equal(t, "zzz", results[0].Name)
+	assert.Equal(t, "mmm", results[1].Name)
+	assert.Equal(t, "aaa", results[2].Name)
+}
+
```
itsmeleela, line 265: Not sure if I overlooked, I didnt see a test for agent path resolution. 
May be as the --agent path reads from the local filesystem, this is typically easier to unit-test  Verify that a test covers: agent path resolution for each supported agent (cursor, claude-code, github-copilot, windsurf), --project-dir scoping, and the no-fallback behavior.

skills/cli/cli.go
```diff
@@ -4,13 +4,20 @@ import (
 	"github.com/jfrog/jfrog-cli-artifactory/cliutils/flagkit"
 	"github.com/jfrog/jfrog-cli-artifactory/skills/commands/delete"
 	"github.com/jfrog/jfrog-cli-artifactory/skills/commands/install"
+	skillslist "github.com/jfrog/jfrog-cli-artifactory/skills/commands/list"
 	"github.com/jfrog/jfrog-cli-artifactory/skills/commands/publish"
 	"github.com/jfrog/jfrog-cli-artifactory/skills/commands/search"
 	"github.com/jfrog/jfrog-cli-core/v2/plugins/components"
 )
```
itsmeleela, line 12: @bhanurp are we not adding e2e tests for Skills repo yet?

cliutils/flagkit/flags.go
udaykb2: If we see other flags, for example: maxWaitMinutes,limit,offset we are using there stringflag instead of intflag

Which means, earlier we are only consistently using either stringflag or boolflag, even when we need intflag, we were using stringflag only

So, Avoided using Intflag and went with StringFlag, to avoid inconsistencies in the code.

skills/commands/list/list_test.go
udaykb2: yes, we can write.

But my opinion is we are changing the logic of how we get paths for agents. We are planning to use a config file (highest priority), with a fall-back to use hard-coded values (current implementation). Since the logic changes afterwards, I believe if we cover these tests along with config file changes, we can avoid doing repeated work.

skills/cli/cli.go
udaykb2: no, we are not having e2e tests

cliutils/flagkit/flags.go
```diff
@@ -515,6 +516,11 @@ const (
 	skillsFormat        = "skills-" + Format
 	skipScan            = "skip-scan"
 	autoDeleteOnFailure = "auto-delete-on-failure"
+	skillsAgent         = "agent"
```
naveenku-jfrog: why "agent" only ? It should be "skills-agent".

skills/cli/cli.go
```diff
@@ -4,13 +4,20 @@ import (
...
 
 func GetCommands() []components.Command {
 	return []components.Command{
+		{
+			Name:        "list",
+			Flags:       flagkit.GetCommandFlags(flagkit.SkillsList),
+			Description: "List skills from an Artifactory repository (--repo) or from a local AI agent's skills directory (--agent). Exactly one of --repo or --agent must be specified.",
+			Action:      skillslist.RunList,
```
naveenku-jfrog, line 19: There is no Arguments here? Is this intentional?

skills/commands/list/list.go
```diff
@@ -0,0 +1,336 @@
...
+	if lc.limit > 0 && len(results) > lc.limit {
+		results = results[:lc.limit]
+	}
+
+	return lc.printResults(results)
+}
+
+func (lc *ListCommand) printResults(results []listResult) error {
+	if len(results) == 0 {
+		fmt.Println("No skills found.")
```
naveenku-jfrog: Use log.Info or log.Warn. Don't use fmt.Println() for printing.

skills/commands/list/list.go
```diff
@@ -0,0 +1,336 @@
...
+		return nil
+	}
+	if strings.EqualFold(lc.format, "json") {
+		data, err := json.MarshalIndent(results, "", "  ")
+		if err != nil {
+			return fmt.Errorf("failed to marshal results: %w", err)
+		}
+		fmt.Println(string(data))
```
naveenku-jfrog, line 196: Use log.Info or log.Warn. Don't use fmt.Println() for printing.

skills/commands/list/list.go
```diff
@@ -0,0 +1,336 @@
...
+			return fmt.Errorf("--limit must be a positive integer, got: %q", limitStr)
+		}
+	}
+
+	var sortBy string
+	var sortOrder string
+
+	if repoKey != "" {
+		// --repo: sort-by accepts updated (default) or downloads; sort-order not supported
+		if o := c.GetStringFlagValue("sort-order"); o != "" {
```
naveenku-jfrog: Can we use some meaningfull variable names here?

skills/commands/list/list.go
```diff
@@ -0,0 +1,336 @@
...
+			return fmt.Errorf("--sort-order is not supported with --repo")
+		}
+		sortBy = "updated"
+		if s := c.GetStringFlagValue("sort-by"); s != "" {
```
naveenku-jfrog: Can we use some meaningfull variable names here?

skills/commands/list/list.go
```diff
@@ -0,0 +1,336 @@
...
+			s = strings.ToLower(s)
+			if s != "updated" && s != "downloads" {
+				return fmt.Errorf("--sort-by for --repo accepts 'updated' or 'downloads', got: %q", s)
+			}
+			sortBy = s
+		}
+	} else {
+		// --agent / --project-dir: sort-by accepts name (default); sort-order asc/desc
+		sortBy = "name"
+		if s := c.GetStringFlagValue("sort-by"); s != "" {
```
naveenku-jfrog: Can we use some meaningfull variable names here?

skills/commands/list/list.go
```diff
@@ -0,0 +1,336 @@
...
+			s = strings.ToLower(s)
+			if s != "name" {
+				return fmt.Errorf("--sort-by for --agent only accepts 'name', got: %q", s)
+			}
+			sortBy = s
+		}
+		sortOrder = "asc"
+		if o := c.GetStringFlagValue("sort-order"); o != "" {
```
naveenku-jfrog: Can we use some meaningfull variable names here?

skills/commands/list/list.go
```diff
@@ -0,0 +1,336 @@
...
+			o = strings.ToLower(o)
+			if o != "asc" && o != "desc" {
+				return fmt.Errorf("--sort-order must be 'asc' or 'desc', got: %q", o)
+			}
+			sortOrder = o
+		}
+	}
+
+	projectDir := ""
+	if pd := c.GetStringFlagValue("project-dir"); pd != "" {
```
naveenku-jfrog: Can we use some meaningfull variable names here?

skills/common/skills_api.go
```diff
@@ -13,6 +13,31 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
+func ListSkills(serverDetails *config.ServerDetails, repoKey string, limit int, sortBy string) ([]services.SkillListItem, error) {
+	sm, err := utils.CreateServiceManager(serverDetails, 3, 0, false)
```
naveenku-jfrog, line 17: Why these 3,0, etc are hardcoded here? Can we use some constants?

naveenku-jfrog (changes requested):

cliutils/flagkit/flags.go
udaykb2: --agent flag is correct in this context

For below flags where we are using string like "skills-" + limit, this is to avoid conflict with other flags limit, sorBy, sortOrder.

skills/cli/cli.go
udaykb2: yes, "skills list" has no arguments

skills/common/skills_api.go
udaykb2: Thank you for pointing out this
here (3,0,false) represents (httpRetries, httpRetryWaitMilliSec, isDryRun)

If we see all other functions written below, we are passing hardcoded values only.

Will create a separate PR in which we can change every function that uses these hardcoded values.

cliutils/flagkit/flags.go
```diff
@@ -1166,6 +1175,11 @@ var flagsMap = map[string]components.Flag{
 	propSearch:          components.NewBoolFlag(propSearch, "Use Artifactory property search (skill.name) instead of Skills API search.", components.WithBoolDefaultValueFalse()),
 	skipScan:            components.NewBoolFlag(skipScan, "Skip Xray security scan after publish. Can also be set via JFROG_CLI_SKIP_SKILLS_SCAN=true.", components.WithBoolDefaultValueFalse()),
 	autoDeleteOnFailure: components.NewBoolFlag(autoDeleteOnFailure, "Automatically delete the artifact if Xray scan identifies it as malicious.", components.WithBoolDefaultValueFalse()),
+	agent:               components.NewStringFlag(agent, "AI agent name to list locally installed skills for. Supported: claude-code, cursor, github-copilot, windsurf.", components.SetMandatoryFalse()),
+	projectDir:          components.NewStringFlag(projectDir, "Path to the project root for project-scoped skills. Use '.' for the current directory. Falls back to global agent skills if not found.", components.SetMandatoryFalse()),
```
bhanurp: if path not found falls back to global? what if it is a typo? may be we shouldnt fallback but rather complain path doesnt exist?

skills/commands/list/list.go
```diff
@@ -0,0 +1,295 @@
...
+	if len(results) == 0 {
+		log.Info("No skills found.")
+		return nil
+	}
+	if strings.EqualFold(lc.format, "json") {
+		data, err := json.MarshalIndent(results, "", "  ")
+		if err != nil {
+			return fmt.Errorf("failed to marshal results: %w", err)
+		}
+		log.Info("\n" + string(data))
```
bhanurp: instead of log use writeoutput function from jfrog-cli-core so that ouput wont have [INFO] prefix and easy to parse

skills/commands/publish/skillmeta.go
```diff
@@ -69,6 +70,20 @@ func parseFrontmatter(content string) (*SkillMeta, error) {
 		key := strings.TrimSpace(line[:colonIdx])
 		value := stripQuotes(strings.TrimSpace(line[colonIdx+1:]))
 
+		// Handle YAML block scalar indicators (> >- | |-): collect indented continuation lines.
+		if value == ">" || value == ">-" || value == "|" || value == "|-" {
+			var parts []string
+			for i+1 < len(lines) {
+				next := lines[i+1]
+				if next != "" && !strings.HasPrefix(next, " ") && !strings.HasPrefix(next, "\t") {
```
bhanurp, line 78: shouldnt this also skip new lines?

skills/commands/list/list.go
```diff
@@ -0,0 +1,295 @@
...
+	return path
+}
+
+func RunList(c *components.Context) error {
+	repoKey := c.GetStringFlagValue("repo")
+	agentName := c.GetStringFlagValue("agent")
+
+	if repoKey == "" && agentName == "" {
+		return fmt.Errorf("either --repo <repository> or --agent <agent-name> must be specified.\nSupported agents: %s", common.SupportedAgentsList())
+	}
```
bhanurp: duplicate code with list.go 87-92

skills/commands/list/list.go
```diff
@@ -0,0 +1,295 @@
...
+			Name:        name,
+			Version:     latestVersion,
+			Description: item.Summary,
+			Source:      "Repo: " + lc.repoKey,
+		})
+	}
+	return lc.printResults(results)
+}
+
+func (lc *ListCommand) listLocalSkills() error {
```
bhanurp, line 124: refactor the fucntion

skills/commands/list/list.go
```diff
@@ -0,0 +1,295 @@
...
+		// Project-scoped: only look in <projectDir>/<agent-relative-path>, no fallback
+		dir = filepath.Join(lc.projectDir, agent.ProjectDir)
+	} else {
+		// Global scope
+		dir = expandHome(agent.GlobalDir)
+	}
+
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		if os.IsNotExist(err) {
```
bhanurp: Use errors.Is(err, os.ErrNotExist)

```suggestion
		if errors.Is(err, os.ErrNotExist) {
```

skills/common/agents.go
```diff
@@ -0,0 +1,26 @@
...
+	"windsurf":       {GlobalDir: "~/.windsurf/skills", ProjectDir: ".windsurf/skills"},
+}
+
+// SupportedAgentsList returns a human-readable comma-separated list of supported agent names.
+func SupportedAgentsList() string {
+	names := make([]string, 0, len(Agents))
+	for name := range Agents {
+		names = append(names, name)
+	}
+	return strings.Join(names, ", ")
```
bhanurp, line 29: sorting might be a good idea to have a deterministic output after multiple runs

skills/commands/list/list.go
```diff
@@ -0,0 +1,295 @@
...
+	if lc.limit > 0 && len(results) > lc.limit {
+		results = results[:lc.limit]
+	}
+
+	return lc.printResults(results)
+}
+
+func (lc *ListCommand) printResults(results []listResult) error {
+	if len(results) == 0 {
+		log.Info("No skills found.")
```
bhanurp: may be when format is json it should be only []. WDYT?

skills/commands/publish/skillmeta.go
udaykb2: Line 60: we are splitting using \n, so we will never see new lines

skills/commands/list/list.go
udaykb2: This function changes in upcoming improvements. So this change will be brough with that improvement

skills/commands/list/list.go
udaykb2: This change will be brought as part of upcoming list improvement changes

skills/common/agents.go
udaykb2: added sorting

skills/commands/list/list.go
udaykb2: Will update in upcoming improvement task. Actually we are using same method in search also. So will update both places once.

skills/commands/list/list.go
udaykb2: surem, name changed

skills/commands/list/list.go
udaykb2: done

skills/commands/list/list.go
udaykb2: done

cliutils/flagkit/flags.go
udaykb2: yes, if --project-dir is not given we will by default use global scope of the agent

skills/commands/list/list.go
udaykb2: done, removed duplicate code

skills/commands/list/list.go
udaykb2: done, now we are returning [] if format is json

cliutils/flagkit/flags.go
bhanurp: what if --project-dir has a path which cannot be resolved?

bhanurp (approved):

cliutils/flagkit/flags.go
udaykb2: [Info] No skills directory found for agent "cursor"

naveenku-jfrog (approved):

PR #434, opened 2026-04-28
RTECO-1016 - Implement skills update command

udaykb2:
### Overview

**Implement skill update command.**

The changes implemented are
1. Update an installed skill to the latest (if no version is provided) or a specific version using --version
2. Validates --path exists and that the skill is already installed before contacting Artifactory
3. Resolves target version from Artifactory; prompts interactively when the requested version is not found (error in quiet mode)
4. Skips re-download when already at the target version unless --force is set
5. Supports --dry-run to preview the update without making changes
6. Unit tests for path validation, installed-version reading, and version selection logic
7. Update logic:
    a. Create a backup by renaming our skill
    b. Install new version in the folder path
    c. If install successful, delete backup, else restore backup

new version still stays in the directory. So we are now deleting the old version and copying the new version during update
8. Hidden the update command from user

-----


skills/commands/install/install.go
```diff
@@ -115,6 +121,11 @@ func (ic *InstallCommand) Run() error {
 	}
 
 	destDir := ic.getDestDir()
+	if ic.removeExisting {
+		if err := os.RemoveAll(destDir); err != nil {
```
bhanurp: why are we removing all directly? a backup should be created and only after success backup should be removed, otherwise restore backup.

skills/commands/update/update.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/jfrog/jfrog-client-go/utils/log"
+)
+
+// RunUpdate is the CLI action for `jf skills update`.
+func RunUpdate(c *components.Context) error {
+	if c.GetNumberOfArgs() < 1 {
+		return fmt.Errorf("usage: jf skills update <slug> [--path <dir>] [--repo <repo>] [options]")
+	}
+
+	slug := c.GetArgumentAt(0)
```
bhanurp, line 27: publish.ValidateSlug(slug), should be used

skills/commands/update/update.go
```diff
@@ -0,0 +1,163 @@
...
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return fmt.Errorf("path '%s' does not exist", path)
+	}
+	return nil
+}
+
+func readInstalledVersion(skillDir string) (string, error) {
+	meta, err := publish.ParseSkillMeta(skillDir)
+	if err != nil {
+		if os.IsNotExist(err) || strings.Contains(err.Error(), "failed to read SKILL.md") {
```
bhanurp: errors.Is(err, fs.ErrNotExist) is a good enough check string comparision is not required.

skills/commands/update/update.go
```diff
@@ -0,0 +1,163 @@
...
+			return "", fmt.Errorf(
+				"skill '%s' is not installed at '%s'.\n"+
+					"To install it, run: jf skills install %s --path %s",
+				filepath.Base(skillDir), filepath.Dir(skillDir),
+				filepath.Base(skillDir), filepath.Dir(skillDir),
+			)
+		}
+		log.Warn(fmt.Sprintf("Could not read current version from %s/SKILL.md: %s", skillDir, err.Error()))
+		return "", nil
```
bhanurp: any other error apart from IsNotExist is failing silently here but rather it sholdnt for example permissions denied error is sent as warn with installedVersion as empty and error as nil how is it treated by caller?

skills/commands/update/update.go
```diff
@@ -0,0 +1,163 @@
...
+
+	log.Warn(fmt.Sprintf("Version '%s' not found. Please select from the available versions below.", requested))
+
+	options := make([]prompt.Suggest, len(available))
+	for i, v := range available {
+		options[i] = prompt.Suggest{Text: v}
+	}
+	selected := ioutils.AskFromListWithMismatchConfirmation(
+		"Select a version:",
+		fmt.Sprintf("'%%s' is not in the list of available versions."),
```
bhanurp:
```suggestion
		fmt.Sprintf("'%s' is not in the list of available versions."),
```

skills/commands/update/update.go
```diff
@@ -0,0 +1,163 @@
...
+		)
+	}
+
+	log.Warn(fmt.Sprintf("Version '%s' not found. Please select from the available versions below.", requested))
+
+	options := make([]prompt.Suggest, len(available))
+	for i, v := range available {
+		options[i] = prompt.Suggest{Text: v}
+	}
+	selected := ioutils.AskFromListWithMismatchConfirmation(
```
bhanurp, line 205: if a version selected is having a typo or version which is not present are we in a loop until correct version is selected?

skills/commands/install/install.go
```diff
@@ -115,6 +121,11 @@ func (ic *InstallCommand) Run() error {
 	}
 
 	destDir := ic.getDestDir()
+	if ic.removeExisting {
```
bhanurp: if this is only updation lets move it to update instead of install that way install logic stays as is

skills/commands/update/update.go
```diff
@@ -0,0 +1,163 @@
...
+		SetVersion(targetVersion).
+		SetInstallPath(installBase).
+		SetQuiet(quiet).
+		SetRemoveExisting(true)
+
+	return cmd.Run()
+}
+
+func validateInstallBase(path string) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
bhanurp: check if its a directory rather than only stat that way filepath.join works always

bhanurp (changes requested):

skills/commands/update/update.go
udaykb2: Only 2 cases happens when user gives --version some_wrong_version
1. Either we ask him to pick from a list of available version interactively
2. If he don't select any version from the list interactively and press enter... we will download the latest version automatically

skills/commands/update/update.go
udaykb2: added

skills/commands/update/update.go
udaykb2: added change, now it checks whether the path is directory or not

skills/commands/update/update.go
udaykb2: added, now we are using errors.Is(err, fs.ErrNotExist)

skills/commands/update/update.go
udaykb2: updated. Seems this kind of error is already present in skills_api.go file as well, will update that during Refactoring task

skills/commands/update/update.go
udaykb2: updated, now we are returning error

skills/commands/install/install.go
udaykb2: removed changes in install.go

skills/commands/install/install.go
udaykb2: moved to update.go

skills/commands/update/update.go
bhanurp: does it mean empty version is latest? worthy mention in documents.

bhanurp (dismissed):

skills/commands/update/update.go
udaykb2: sure, noted

naveenku-jfrog (dismissed):

bhanurp (approved):

agrasth (approved):

agrasth (approved):

itsmeleela (approved):

PR #435, opened 2026-04-28
Add --format flag support to artifactory commands

ehl-jf:
## Summary

Adds `--format` flag support across `rt` commands so output can be rendered as `json`, `table`, or the default human-readable text. The flag plumbs through to the framework's shared format machinery rather than each command rolling its own.

## Changes

**Commands with new `--format` support**

- Generic: `search`, `ping`, `upload`, `download`, `move`, `copy`, `delete`, `direct-download`, `set-props`, `delete-props`, `git-lfs-clean`
- Build: `build-publish`, `build-promote`, `build-discard`, `build-add-dependencies`
- Docker / Podman: `docker-promote`, `podman-push`, `podman-pull`
- Repo / Replication: `repo-create`, `repo-update`, `replication-create`
- NuGet: `nuget-deps-tree`

**`build-publish` specifics**

- `BuildPublishCommand` now exposes `GetBuildInfoUiUrl()`, `SetSuppressOutput()`, and `SetCollectSha256()` so the CLI layer can render the UI URL (and sha256) when `--format` is set, without double-printing JSON from `Run()`.

**Tests**

- New `artifactory/cli/cli_test.go` (~1.3k lines) covering format rendering across the affected commands.

---

### DIFF
--- a/artifactory/cli/cli.go
+++ b/artifactory/cli/cli.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -12,7 +13,6 @@ import (
 
 	ioutils "github.com/jfrog/gofrog/io"
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/buildinfo"
-	"github.com/jfrog/jfrog-cli-artifactory/artifactory/formats"
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/container"
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/curl"
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/dotnet"
@@ -1898,9 +1898,10 @@ func buildPublishCmd(c *components.Context) error {
 	cmd.SetDepExcludeScopes(c.GetStringsArrFlagValue("dep-exclude-scopes"))
 
 	// When --format is set, suppress the internal logJsonOutput call so that the
-	// CLI layer can render the URL itself.
+	// CLI layer can render the URL itself, and collect sha256.
 	if outputFormat != coreformat.None {
 		cmd.SetSuppressOutput(true)
+		cmd.SetCollectSha256(true)
 	}
 
 	err = commands.Exec(cmd)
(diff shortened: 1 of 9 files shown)


vitaliil-jfrog (approved):

artifactory/cli/cli.go
```diff
@@ -482,9 +524,46 @@ func dockerPromoteCmd(c *components.Context) error {
...
+	// The client layer discards the body, so we pass nil and let the helper
+	// synthesize {"status_code": 200, "message": "OK"}.
+	if outputFormat != coreformat.None {
+		printStatusJSON(200, "OK")
+	}
+	return nil
+}
+
+// printStatusJSON emits a synthetic JSON response with the given HTTP status code and message.
+func printStatusJSON(statusCode int, message string) {
```
agrasth: This helper is void while every sibling (printPingJSON, printBuildPublishJSON, etc.) returns error. The json.Marshal result is silently discarded with `_`.

artifactory/cli/cli.go
```diff
@@ -1454,7 +2235,33 @@ func gitLfsCleanCmd(c *components.Context) error {
...
+
+// printGitLfsCleanResponse renders the git-lfs-clean result in the requested output format.
+func printGitLfsCleanResponse(succeeded, failed int, outputFormat coreformat.OutputFormat, w io.Writer, originalErr error) error {
+	switch outputFormat {
+	case coreformat.Json:
+		if err := printSummaryJSON(succeeded, failed, false, originalErr); err != nil {
+			return err
+		}
+		return originalErr
+	case coreformat.Table:
```
agrasth, line 2160: `originalErr` is silently dropped here. Every other count-based response function (printMoveResponse, printCopyResponse, printDeleteResponse, etc.) calls `common.GetCliError(originalErr, succeeded, failed, failNoOp)` in the table branch.
Without it, a partial failure won't produce a non-zero exit code when --format table is used.

artifactory/cli/cli.go
```diff
@@ -991,10 +1448,38 @@ func moveCmd(c *components.Context) error {
...
+		return fmtErr
+	}
+	if outputFormat == coreformat.None {
+		return printBriefSummaryAndGetError(result.SuccessCount(), result.FailCount(), common.IsFailNoOp(c), err)
+	}
+	return printMoveResponse(result.SuccessCount(), result.FailCount(), outputFormat, os.Stdout, common.IsFailNoOp(c), err)
+}
+
+// printMoveResponse renders the move result in the requested output format.
+func printMoveResponse(succeeded, failed int, outputFormat coreformat.OutputFormat, w io.Writer, failNoOp bool, originalErr error) error {
```
agrasth: printMoveResponse, printCopyResponse, printDeleteResponse, printSetPropsResponse, printDeletePropsResponse, and printBuildAddDependenciesResponse are structurally identical — same switch body, only the command name in the error string differs.    
Consider extracting:                                                                
```
  func printCountBasedResponse(cmdName string, succeeded, failed int, outputFormat    
  coreformat.OutputFormat, w io.Writer, failNoOp bool, originalErr error) error {
      switch outputFormat {                                                           
      case coreformat.Json:
          if err := printSummaryJSON(succeeded, failed, failNoOp, originalErr); err !=
   nil {                                                                              
              return err                                        
          }                                                                           
          return common.GetCliError(originalErr, succeeded, failed, failNoOp)
      case coreformat.Table:
          if err := printCountsTable(succeeded, failed, w); err != nil {              
              return err                                        
          }                                                                           
          return common.GetCliError(originalErr, succeeded, failed, failNoOp)
      default:
          return errorutils.CheckErrorf("unsupported format '%s' for rt %s. Acceptable
   values are: json, table", outputFormat, cmdName)                                   
      }                                                         
  }                                                                                   
```

artifactory/cli/cli.go
```diff
@@ -830,12 +1133,110 @@ func downloadCmd(c *components.Context) error {
...
+			return err
+		}
+		return common.GetCliError(originalErr, result.SuccessCount(), result.FailCount(), failNoOp)
+	default:
+		return errorutils.CheckErrorf("unsupported format '%s' for rt direct-download. Acceptable values are: json, table", outputFormat)
+	}
+}
+
+// directDownloadTableRow is a table-printable representation of a directly downloaded file.
+type directDownloadTableRow struct {
```
agrasth: directDownloadTableRow is identical to downloadTableRow (same SOURCE/TARGET fields and col-name tags). One can be removed and the other reused.

agrasth (approved):

artifactory/cli/cli.go
```diff
@@ -86,94 +95,105 @@ const (
...
-			Category:    filesCategory,
+			Name:             "search",
+			Flags:            flagkit.GetCommandFlags(flagkit.Search),
+			Aliases:          []string{"s"},
+			Description:      search.GetDescription(),
+			Arguments:        search.GetArguments(),
+			Action:           searchCmd,
+			Category:         filesCategory,
+			SupportedFormats: []coreformat.OutputFormat{coreformat.Json, coreformat.Table},
+			DefaultFormat:    coreformat.Json,
```
reshmifrog, line 166: why only for search and nuget-deps-tree do we have the `DefaultFormat`

reshmifrog (changes requested):

reshmifrog:
Please update the description , It needs to have all the information regarding the changes you have done as part of this PR.

artifactory/cli/cli.go
ehl-jf: The idea was not to break, the current behavior.
The rule I applied is putting as default what the command already output without the `--format`.
So if a command did not output anything before it will have no default.

reshmifrog (approved):

PR #437, opened 2026-04-29
Add UV package manager setup support

sverdlov93:
## Summary

Implement UV-specific setup in the `SetupCommand`, enabling persistent configuration of UV to use Artifactory PyPI repositories. Follows the same architecture as pip/poetry/twine setup.

### Setup flow (`configureUV`)
1. **Credentials** — stored via `uv auth login <domain> --username <user> --password <token>` using UV's native credential store. Skipped for anonymous access.
2. **Index configuration** — writes a `[[index]]` entry to the user-level `uv.toml` (`~/.config/uv/uv.toml`) pointing to the Artifactory PyPI simple API, set as default.

### TOML utilities (`uvsetup.go`)
- `ConfigureUVIndex` / `RemoveUVIndex` / `GetConfiguredUVIndexURL` — read/write `uv.toml` while preserving all unrelated settings (uses `map[string]any` to avoid clobbering unknown keys)
- `RunUVAuthLogin` / `RunUVAuthLogout` — thin wrappers around UV CLI commands
- `getUserUVConfigPath` — resolves config path respecting `UV_CONFIG_FILE`, `XDG_CONFIG_HOME`, `%APPDATA%`

### Why not reuse existing Python PM utilities?
Each Python PM has unique config mechanisms: pip uses `pip config set`, poetry uses `poetry config`, twine uses `.pypirc` (INI). UV uses TOML files + a separate auth CLI. No shared abstraction fits.

## Testing
- **16 unit tests** (`uvsetup_test.go`) — config path resolution, TOML load/save, index add/update/remove, settings preservation
- **3 setup integration tests** (`setup_test.go`) — token auth, basic auth, anonymous access (skipped when `uv` binary not in PATH)

## Depends on
- jfrog/jfrog-cli-core#1549


artifactory/commands/python/uv.go
```diff
@@ -75,3 +82,239 @@ func (pc *UvCommand) GetStdWriter() io.WriteCloser {
 func (pc *UvCommand) GetErrWriter() io.WriteCloser {
 	return nil
 }
+
+// ── Setup helpers ────────────────────────────────────────────────────────────
+
+const (
+	uvIndexName = "jfrog-pypi"
```
yahavi, line 92: **Index name collision across JFrog instances.** Hardcoding `jfrog-pypi` means a user who runs `jf setup uv` against two different JFrog servers (e.g. via `jf c add`) will silently overwrite the first server's entry on the second run. Consider deriving the name from the server ID or repo name (e.g. `jfrog-pypi-<serverId>`), or at minimum document in the function comment that re-running setup replaces the previous JFrog index.

artifactory/commands/python/uv.go
```diff
@@ -75,3 +82,239 @@ func (pc *UvCommand) GetStdWriter() io.WriteCloser {
...
+)
+
+type uvIndex struct {
+	Name    string `toml:"name"`
+	URL     string `toml:"url"`
+	Default bool   `toml:"default,omitempty"`
+}
```
yahavi: **Duplicate of `uvIndexEntry` in `native_uv.go` (line 416).** Same package, same concept, two structs that differ only by `Default bool`. Suggest extending the existing one and dropping this:

```go
// in native_uv.go
type uvIndexEntry struct {
    Name    string `toml:"name"`
    URL     string `toml:"url"`
    Default bool   `toml:"default,omitempty"`
}
```

`omitempty` keeps `native_uv.go`'s pyproject.toml reads byte-identical (the field stays zero there). Pure refactor, no behavior change.

artifactory/commands/python/uv.go
```diff
@@ -75,3 +82,239 @@ func (pc *UvCommand) GetStdWriter() io.WriteCloser {
...
+
+// RunUVAuthLogin stores credentials in UV's native credential store.
+//
+//	uv auth login <service-url> --username <user> --password <password>
+func RunUVAuthLogin(serviceURL, username, password string) error {
+	log.Debug("Running uv auth login for", serviceURL)
+	cmd := exec.Command("uv", "auth", "login", serviceURL, "--username", username, "--password", password)
```
yahavi, line 106: **Two notes on this `exec.Command` line:**

1. **Don't switch to `--token`.** I tested with uv 0.11.8: `--username + --token` is rejected (mutually exclusive), and `--token` alone stores `username = "__token__"` (a PyPI convention Artifactory doesn't recognize). The current `--username + --password` is the only combination that works for both basic auth and access tokens. Worth a one-line comment here so the next reader doesn't try to "fix" it.
2. **`--password <pwd>` exposes the secret in `ps` / `/proc` for the duration of the call.** uv 0.11 doesn't yet support `--password-stdin`, so this is the only documented path today. Add a `// TODO: switch to --password-stdin once supported by uv` so it's revisited.

Also: consider reusing/introducing a tiny shared helper instead of inlining `exec.Command("uv", ...)` here and in `RunUVAuthLogout`. `native_uv.go` has `runUvBinary` for streaming stdio; you want a captured-output sibling:

```go
func runUvBinaryCaptured(args ...string) (string, error) {
    out, err := exec.Command("uv", args...).CombinedOutput()
    return string(out), err
}
```

Then the `"uv"` literal lives in one place, which makes a future `UV_BIN` env-var override trivial.

artifactory/commands/python/uv.go
```diff
@@ -75,3 +82,239 @@ func (pc *UvCommand) GetStdWriter() io.WriteCloser {
...
+	if username != "" {
+		args = append(args, "--username", username)
+	}
+	cmd := exec.Command("uv", args...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return errorutils.CheckErrorf("uv auth logout failed: %s\n%s", err, string(output))
+	}
+	return nil
+}
```
yahavi, line 130: **`RunUVAuthLogout`, `RemoveUVIndex` (line 168), and `GetConfiguredUVIndexURL` (line 197) are exported but never called.** Greppable confirmation:

```
$ rg -n 'RunUVAuthLogout|RemoveUVIndex|GetConfiguredUVIndexURL' --type go
artifactory/commands/python/uv.go: <definitions only>
artifactory/commands/python/uv_test.go: <test calls only>
```

If a follow-up PR will wire these into a logout / `jf c rm` / re-setup flow, please link the issue in the PR description. Otherwise either unexport them or remove them — speculative public API ages badly and forces future signature compatibility.

artifactory/commands/python/uv.go
```diff
@@ -75,3 +82,239 @@ func (pc *UvCommand) GetStdWriter() io.WriteCloser {
...
+			found = true
+			break
+		}
+	}
+	if !found {
+		indexes = append(indexes, uvIndex{
+			Name:    uvIndexName,
+			URL:     indexURL,
+			Default: true,
+		})
```
yahavi, line 163: **`Default: true` may stomp the user's other default indexes.** When the JFrog index is added/updated, no other index in `uv.toml` gets demoted. uv allows multiple `default = true` entries in the TOML but only one is honored at runtime (last-wins in current uv), so the user's previous default may silently stop working.

Safer policy: walk all entries and clear `Default` on everything except the JFrog one when claiming default. Something like:

```go
for i := range indexes {
    indexes[i].Default = (indexes[i].Name == uvIndexName)
}
```

artifactory/commands/python/uv.go
```diff
@@ -75,3 +82,239 @@ func (pc *UvCommand) GetStdWriter() io.WriteCloser {
...
+func getUserUVConfigPath() (string, error) {
+	if configFile := os.Getenv("UV_CONFIG_FILE"); configFile != "" {
+		return configFile, nil
+	}
+
+	var configDir string
+	if runtime.GOOS == "windows" {
+		configDir = os.Getenv("APPDATA")
+		if configDir == "" {
+			return "", errorutils.CheckErrorf("%%APPDATA%% not set")
```
yahavi: **Bug: `%%APPDATA%%` will be printed verbatim, not as `%APPDATA%`.** `errorutils.CheckErrorf` falls back to `errors.New(format)` when called with no args (see `jfrog-client-go/utils/errorutils/errorutils.go`). It does NOT run `fmt.Sprintf`, so `%%` is not unescaped. The user sees `%%APPDATA%% not set`.

Fix either way:
```go
return "", errorutils.CheckErrorf("%%APPDATA%% not set", "") // pass any arg to force fmt.Errorf path
// or just
return "", errorutils.CheckErrorf("%APPDATA% not set")     // single-percent literal works in errors.New
```

yahavi (commented):
Solid PR — architecture follows the existing pip/poetry/twine setup pattern cleanly. Inline comments below cover three correctness/safety bugs that should block merge, plus a handful of quality and reuse suggestions.

**Verified locally** on this commit with `uv 0.11.8`:
- `go vet` and all unit tests pass.
- `TestSetupCommand_UV` passes but actually pollutes the developer's real `~/.local/share/uv/credentials/credentials.toml` — confirmed by inspecting the file after the run. The `uv auth logout` cleanup line is a no-op when multiple entries share a service URL (uv requires `--username`). See the inline comment on `setup_test.go`.
- Empirically tested `uv auth login` flag combinations: `--username + --token` is rejected by uv (mutually exclusive), and `--token` alone stores `username = "__token__"` (a PyPI-ism Artifactory likely doesn't accept). So the current `--username + --password` design is the right call — see the comment on `RunUVAuthLogin`.

artifactory/commands/python/uv.go
```diff
@@ -75,3 +82,239 @@ func (pc *UvCommand) GetStdWriter() io.WriteCloser {
...
+	if _, err := toml.Decode(string(data), &fullCfg); err != nil {
+		return nil, nil, errorutils.CheckErrorf("failed to parse uv config at %s: %w", path, err)
+	}
+
+	var indexes []uvIndex
+	if rawIndexes, ok := fullCfg["index"]; ok {
+		indexSlice, ok := rawIndexes.([]map[string]any)
+		if !ok {
+			return fullCfg, nil, nil
+		}
```
yahavi, line 271: **Silent data loss when `index` field has an unexpected shape.** If the user has a single `[index]` table (object) instead of `[[index]]` (array of tables), or if BurntSushi returns `[]any` (it can, depending on heterogeneous content) instead of `[]map[string]any`, this assertion fails and the function returns `(fullCfg, nil, nil)`. The `index` key is still in `fullCfg`, but the caller now thinks there are zero indexes.

Then on the next write, `writeUVConfig` does:
```go
if len(indexMaps) > 0 {
    fullCfg["index"] = indexMaps
} else {
    delete(fullCfg, "index")  // ← user's malformed but valid index block is destroyed
}
```

At minimum, return an error here so we don't silently rewrite the user's config and lose data. Ideally also handle the `[]any` branch.

artifactory/commands/setup/setup.go
```diff
@@ -494,6 +497,42 @@ func (sc *SetupCommand) configureGradle() error {
...
+		return fmt.Errorf("failed to get PyPI repository URL with credentials: %w", err)
+	}
+
+	// Store credentials only when authentication is configured (skip for anonymous access)
+	if username != "" && password != "" {
+		serviceURL := repoUrl.Scheme + "://" + repoUrl.Host
+		if err := python.RunUVAuthLogin(serviceURL, username, password); err != nil {
+			return fmt.Errorf("failed to store UV credentials: %w", err)
+		}
+	}
```
yahavi, line 531: **Stale credentials remain in uv's credential store after switching to anonymous.** A user who first ran `jf setup uv` with a token, then re-runs against an anonymous repo (or re-points to a different repo whose service URL collides), keeps using the old credentials silently — uv resolves the service URL and finds the leftover entry in `~/.local/share/uv/credentials/credentials.toml`.

Suggest: when entering anonymous mode, attempt `uv auth logout <serviceURL>` (best-effort, ignore "no credentials" errors) before writing the new index. This is also a use-case for the unused `RunUVAuthLogout` function.

artifactory/commands/setup/setup_test.go
```diff
@@ -659,6 +659,46 @@ func TestSetupCommand_Twine(t *testing.T) {
...
+			uvConfigContent := string(uvConfigContentBytes)
+
+			assert.Contains(t, uvConfigContent, `name = "jfrog-pypi"`)
+			assert.Contains(t, uvConfigContent, "acme.jfrog.io/artifactory/api/pypi/test-repo/simple")
+			assert.Contains(t, uvConfigContent, "default = true")
+			assert.Contains(t, uvConfigContent, `publish-url = "https://acme.jfrog.io/artifactory/api/pypi/test-repo"`)
+
+			// Clean up: remove stored credentials and config file
+			_ = exec.Command("uv", "auth", "logout", "https://acme.jfrog.io").Run()
+			_ = os.Remove(uvConfigPath)
```
yahavi: **This cleanup doesn't actually clean up — verified locally.** I ran this test on this branch with `uv 0.11.8` and confirmed two leftover entries in my real `~/.local/share/uv/credentials/credentials.toml`:

```toml
[[credential]]
service = "https://acme.jfrog.io/"
username = "test-user"
scheme = "basic"
password = "<fake JWT>"

[[credential]]
service = "https://acme.jfrog.io/"
username = "myUser"
scheme = "basic"
password = "myPassword"
```

`uv auth logout <service>` without `--username` is a no-op when multiple entries exist for the same service — uv will not pick one arbitrarily. Every CI/dev test run accumulates entries.

**Fix: isolate uv's credentials directory itself, not just the config file.** uv reads the dir from `XDG_DATA_HOME` / `%APPDATA%`. Add at the top of the test:

```go
t.Setenv("XDG_DATA_HOME", t.TempDir()) // unix
t.Setenv("APPDATA", t.TempDir())       // windows
```

Then these two cleanup lines (`uv auth logout` and `os.Remove`) become unnecessary — `t.TempDir` cleans itself, and the credential file lives entirely inside the sandbox.

.gitignore
```diff
@@ -21,3 +21,4 @@
 
 .gradle
 **/gradle*/**/bin
+.fly/
```
yahavi, line 24: Unrelated to UV setup. Either split into a tiny prep PR or call it out explicitly in the PR description so reviewers don't wonder why a Fly-related ignore appears in a UV feature.

yahavi (dismissed):

artifactory/commands/python/uv.go
yahavi: **Follow-up: empirically tested the `ps`-leak surface, downgrading the priority.**

I ran a known-marker token through every uv command path I could think of, with a tight `ps -ww` polling loop. Results on uv 0.11.8:

| Scenario | Token leaks in output |
|---|---|
| `uv sync` | 0 |
| `uv sync --verbose` (DEBUG logs) | 0 |
| `uv pip install --dry-run --verbose requests` | 0 |
| `uv lock --verbose` | 0 |
| `uv auth login ... --bogus-flag` (error path that prints usage) | 0 |
| `uv auth token <service>` (the intentional reveal command, no stored creds) | 0 |
| **`ps -ww` polling during `uv auth login`** | **1 — caught on first iteration** |

uv has good credential hygiene in normal operation — even logs `DEBUG Received a cross-origin redirect. Removing sensitive headers.` So once setup completes, all subsequent `jf uv sync` / `uv install` / `uv publish` invocations are safe.

The only confirmed leak is the sub-second `ps` race during `uv auth login` itself. **Real, but bounded.**

**One bonus uv-upstream issue surfaced during testing:** uv writes the credentials file with mode `0644` (world-readable!) and the lock file with `0666` (world-writable). On a shared host any other local user can read the stored token at rest:

```
-rw-r--r--  1 user  staff  ... credentials.toml
-rw-rw-rw-  1 user  staff  ... credentials.toml.lock
```

That's outside this PR's control, but it strengthens the case for option A (write the credentials file directly): it would let the PR set `0600` itself and also kill the `ps` race. Not a blocker — just context for the trade-off.

**Updated recommendation: downgrade the `--password-stdin` TODO from "should do" to "nice to have, document and move on".** Empirical evidence of bounded leak, no leaks in steady-state operation.

yahavi (dismissed):

artifactory/commands/python/uv.go
```diff
@@ -75,3 +82,249 @@ func (pc *UvCommand) GetStdWriter() io.WriteCloser {
...
+		if idx.Name == uvIndexName {
+			indexes[i].URL = indexURL
+			indexes[i].Default = true
+			found = true
+		} else {
+			indexes[i].Default = false
+		}
+	}
+	if !found {
+		for i := range indexes {
```
agrasth: This inner loop is redundant. When `found == false`, the first loop (lines 149–157) already ran `indexes[i].Default = false` in the else branch for every entry (since none matched uvIndexName). The second pass over the same slice does nothing new.

go.mod
```diff
@@ -198,7 +198,7 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
-// replace github.com/jfrog/jfrog-cli-core/v2 => github.com/agrasth/jfrog-cli-core/v2 v2.0.0-20260428093456-2bc01db3b153
+replace github.com/jfrog/jfrog-cli-core/v2 => github.com/jfrog/jfrog-cli-core/v2 v2.60.1-0.20260429145410-f5a17911e10d
```
agrasth: before merging you can directly updated the version in require.

agrasth (dismissed):

agrasth (approved):

naveenku-jfrog (approved):

PR #440, opened 2026-05-05
Rteco 1069 skills search surface errors

bhanurp:
-----
Improved logging when jf skills search fails with error message.


skills/commands/search/search.go
```diff
@@ -93,6 +99,11 @@ func (sc *SearchCommand) runSkillsAPISearch() error {
 		}
 	}
 
+	if len(results) == 0 && len(failedRepos) == len(repos) {
+		return fmt.Errorf("skills search for %q failed in %s: %w",
```
itsmeleela, line 103: means if 4/5 repos fail but 1 succeeds, the user gets partial results with no indication that most repos errored. Is that what we are expecting?

itsmeleela: Worth adding a post-loop summary warn so users know the result set may be incomplete: log.Warn(fmt.Sprintf("Skills search returned partial results; failed repos: %s", strings.Join(failedRepos, ", ")))

itsmeleela (approved):

fluxxBot (approved):

PR #442, opened 2026-05-07
add nix package manager

reshmifrog:
-----
Added the support for nix package manager


Balasubramanyamkosuri (approved):
## Review summary

Solid `NixCommand` implementation — clear dispatch by native tool and appropriate netrc auth with cleanup.

### Suggestions (non-blocking)

1. Document which subcommands collect build-info vs passthrough (`nix-channel` is passthrough only).
2. Add an IT or doc example for `nix copy` URL format matching Artifactory Nix repo Set Me Up (`api/nix/<repo>/`).
3. Clear error when `nix` / `nix-env` / `nix-channel` is not on PATH.
4. After build-info-go#378 merges (with mergeArtifacts fix), pin released build-info-go version in go.mod.

**Blocked on:** build-info-go#378 mergeArtifacts fix before final dependency bump.

Structured review: `docs/reviews/nix-pr-review.md` in team workspace.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+
+// runNixBuild executes "nix-build" with args, then collects build-info from output store path.
+func (c *NixCommand) runNixBuild() error {
+	log.Info("Running nix-build")
+	cmd := exec.Command("nix-build", c.args...)
+	cmd.Env = c.buildEnv()
+	cmd.Stderr = os.Stderr // Show build progress to user
+	output, err := cmd.Output()
+	if err != nil {
+		return fmt.Errorf("nix-build failed: %s: %w", string(output), err)
```
agrasth: `fmt.Errorf("nix-build failed: %s: %w", string(output), err) — output` is stdout (the store path), not stderr. On failure stdout is likely empty; the useful error was already streamed to `os.Stderr`. The error message will just say` "nix-build failed: : exit status 1"`. Remove %s or capture stderr separately.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	if c.buildConfiguration != nil && len(storePaths) > 0 {
+		return c.collectBuildInfoFromStorePaths(storePaths)
+	}
+	return nil
+}
+
+// runNixFlakeBuild executes "nix build" (flake-style) with args, then collects build-info
+// from the ./result symlink. Unlike nix-build which prints store paths to stdout,
+// "nix build" creates a ./result symlink pointing to the output store path.
+func (c *NixCommand) runNixFlakeBuild() error {
```
agrasth, line 196: `runNixFlakeBuild` only follows the ./result symlink. nix build can create multiple outputs (`./result`, `./result-bin`, `./result-man`, etc.). 
Should glob result* symlinks in the working directory?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+				}
+			}
+		}
+	}
+	return ""
+}
+
+// resolveDefaultDeploymentRepo queries Artifactory REST API to find the defaultDeploymentRepo
+// for a virtual repo. Returns empty if not virtual or no default deployment repo.
+func (c *NixCommand) resolveDefaultDeploymentRepo(repoName string) string {
```
agrasth, line 666: Uses a raw `http.Client` instead of the jfrog-client-go HTTP client. This bypasses the server's proxy, TLS, and retry configuration. Use `utils.CreateServiceManager` + the repositories API instead.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	url := strings.TrimSuffix(c.serverDetails.ArtifactoryUrl, "/") + "/api/repositories/" + repoName
+	client := &http.Client{Timeout: 10 * time.Second}
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return ""
+	}
+	if c.serverDetails.User != "" {
+		password := c.serverDetails.Password
+		if password == "" {
+			password = c.serverDetails.AccessToken
```
agrasth: `password = c.serverDetails.AccessToken` is used as a Basic-auth password. That's JFrog convention, but it's worth checking if the REST call in `resolveDefaultDeploymentRepo` should use Bearer auth when an access token is set (no User field configured).

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+		if item.Path == "." {
+			pathInRepo = item.Name
+		}
+		return item.Name, pathInRepo
+	}
+	return "", ""
+}
+
+// getArtifactChecksums fetches sha1/sha256/md5 for a file in Artifactory.
+func (c *NixCommand) getArtifactChecksums(repo, pathInRepo string) entities.Checksum {
```
agrasth: Uses filepath.Dir / filepath.Base for AQL path components. On Windows these produce backslash separators, which will break the AQL query. Use path.Dir / path.Base from the path package.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+		return fmt.Errorf("create build: %w", err)
+	}
+	if err := buildInstance.SaveBuildInfo(buildInfo); err != nil {
+		return fmt.Errorf("save build info: %w", err)
+	}
+	return nil
+}
+
+// Ensure unused imports don't cause compile errors
+var _ = json.Marshal
```
agrasth: this is a code smell indicating json is imported but not used in some paths. Remove the import if unused, or refactor the code so it's genuinely needed.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	buildNumber, err := c.buildConfiguration.GetBuildNumber()
+	if err != nil || buildNumber == "" {
+		return "", "", fmt.Errorf("build number not configured")
+	}
+	return buildName, buildNumber, nil
+}
+
+// GetCmd returns the exec.Cmd for gofrogcmd.RunCmd interface.
+func (c *NixCommand) GetCmd() *exec.Cmd {
+	return exec.Command("nix", append([]string{c.nativeTool}, c.args...)...)
```
agrasth: For nativeTool = "nix-channel", this would run nix nix-channel ... instead of the nix-channel binary. GetCmd() / GetEnv() / GetStdWriter() / GetErrWriter() look like they implement gofrogcmd.RunCmd, but commands.Exec probably calls Run() directly. If these interface methods are unused/dead, remove them to avoid confusion.

artifactory/commands/nix/command.go
agrasth, line 1: consider adding tests for this file.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	if err != nil {
+		return
+	}
+	if _, err = tmpFile.WriteString(netrcContent); err != nil {
+		return
+	}
+	if err = tmpFile.Close(); err != nil {
+		return
+	}
+	c.netrcPath = tmpFile.Name()
```
naveenku-jfrog, line 791: Two issues here:
1. If WriteString (line 610) or Close (line 613) fails, the function
   returns early but tmpFile is never closed and the temp file is never
   deleted — OS fd + disk leak. Add `_ = tmpFile.Close(); _ = os.Remove(tmpFile.Name())`
   in each early-return error path.
2. The function has no return value, so auth failures are invisible to
   the caller. Callers in Run() (line 80) just continue silently and Nix
   runs unauthenticated, producing cryptic "access denied" errors downstream.
   Change the signature to `createNetrcFile() error` and return fmt.Errorf
   on each failure.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+		DefaultDeploymentRepo string `json:"defaultDeploymentRepo"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&repoConfig); err != nil {
+		return ""
+	}
+
+	if repoConfig.Rclass == "virtual" && repoConfig.DefaultDeploymentRepo != "" {
+		return repoConfig.DefaultDeploymentRepo
+	}
+	return ""
```
naveenku-jfrog: This raw http.Client bypasses all the jfrog-client-go transport layer, which means:
1. InsecureTls / custom CA certs in ServerDetails are ignored.
2. Token-only auth is silently skipped — line 489 checks `c.serverDetails.User != ""`
   but modern JFrog Platform setups often use access tokens with no User field,
   so no Authorization header is set at all.
   Fix: if User is empty but AccessToken is set, use:
     req.Header.Set("Authorization", "Bearer "+c.serverDetails.AccessToken)
3. Consider replacing this with the existing Artifactory service manager /
   jfrog-client-go repository API to get retries and proper TLS handling for free.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+				// Find the .nar.xz file — the actual dependency binary
+				narXzName, narXzPath := c.findNarFile(searchRepo, dirPath)
+				if narXzName != "" {
+					artChecksum := c.getArtifactChecksums(searchRepo, narXzPath)
+					if artChecksum.Sha1 != "" || artChecksum.Sha256 != "" {
+						buildInfo.Modules[0].Dependencies[i].Checksum = artChecksum
+						resolved++
+					}
+				}
+			}
```
naveenku-jfrog: findNarFile() (line 327/398) and getArtifactChecksums() (line 329/413)
each call utils.CreateServiceManager internally. For a Nix closure with
50–200 store paths this creates hundreds of ServiceManager instances and
HTTP connection pools in a tight loop.
Create a single ServiceManager once before the loop and thread it through,
or promote it to a field on NixCommand set during Run().

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	var pkgAttr string
+	for i := len(c.args) - 1; i >= 0; i-- {
+		if !strings.HasPrefix(c.args[i], "-") && strings.Contains(c.args[i], ".") {
+			pkgAttr = c.args[i]
+			break
+		}
+	}
+	if pkgAttr == "" {
+		return nil
+	}
```
naveenku-jfrog, line 283: The dot-check on line 248 means `nix-env -i hello` silently skips
build-info collection with no warning. Package names without a channel
prefix (e.g. plain "hello", "git") are completely ignored.
At minimum add a log.Warn here so the user knows why no build-info
was collected. Ideally document in the method comment which invocation
patterns are supported.

artifactory/commands/nix/command.go
naveenku-jfrog: cmd.Output() captures stdout; nix-build writes store paths to stdout,
not errors. Errors go to stderr which is already forwarded to os.Stderr
(line 140), so string(output) in the error message will be the store
path (or empty), not a diagnostic.
Use cmd.CombinedOutput() or just drop string(output) from the
fmt.Errorf call.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+		CommonParams: &specutils.CommonParams{
+			Aql: specutils.Aql{ItemsFind: searchQuery},
+		},
+	}
+	reader, err := servicesManager.SearchFiles(searchParams)
+	if err != nil {
+		return
+	}
+	_, _ = servicesManager.SetProps(services.PropsParams{Reader: reader, Props: props})
+}
```
naveenku-jfrog: Every error path in this function returns without logging or surfacing
anything. A failed SetProps call is completely invisible to the user.
The function should at minimum log.Warn on each failure, and ideally
return error so the caller (tagUploadedArtifacts) can decide whether
to abort or continue.

naveenku-jfrog (changes requested):

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	case "nix-channel":
+		return c.runNixChannel()
+	case "nix-env":
+		return c.runNixEnv()
+	case "nix-build":
+		return c.runNixBuild()
+	case "build":
+		return c.runNixFlakeBuild()
+	case "copy":
+		return c.runNixCopy()
```
fluxxBot, line 127: do we actually need to capture all this commands? what about nixEnv, will capturing this modify our build-info?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	log.Info("Running nix-channel")
+	cmd := exec.Command("nix-channel", c.args...)
+	cmd.Env = c.buildEnv()
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("nix-channel failed: %w", err)
+	}
+	return nil
+}
```
fluxxBot: why can't this be a simple pass through?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+// runNixEnv executes "nix-env" with args, then collects build-info from runtime closure.
+func (c *NixCommand) runNixEnv() error {
+	log.Info("Running nix-env")
+	cmd := exec.Command("nix-env", c.args...)
+	cmd.Env = c.buildEnv()
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("nix-env failed: %w", err)
+	}
```
fluxxBot, line 160: this line is being constantly repeated, consider making this a utility function or a mere pass through

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+		deployRepo := c.resolveDefaultDeploymentRepo(c.repo)
+		if deployRepo != "" && deployRepo != c.repo {
+			log.Info(fmt.Sprintf("Resolved default deployment repo: %s → %s", c.repo, deployRepo))
+			c.replaceRepoInToArg(c.repo, deployRepo)
+			c.repo = deployRepo
+			// Add --refresh so nix re-checks the LOCAL repo (which is empty)
+			// instead of using cached knowledge from previous virtual repo checks
+			c.args = append([]string{"--refresh"}, c.args...)
+		}
+	}
```
fluxxBot, line 239: check if a similar utility method exists, also what happens if defaultdDeploymentRepo is not present

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	cmd.Env = c.buildEnv()
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("nix %s failed: %w", c.nativeTool, err)
+	}
+	return nil
+}
+
+// collectDepsFromEnvArgs resolves the store path from nix-env args (e.g., "nixpkgs.hello").
```
fluxxBot: why do we have so many different ways and logic across build-info-go and here to resolve storePaths?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	parts := strings.SplitN(pkgAttr, ".", 2)
+	if len(parts) != 2 {
+		return nil
+	}
+
+	// Resolve store path: nix-build '<channel>' -A attr --no-out-link
+	cmd := exec.Command("nix-build", fmt.Sprintf("<%s>", parts[0]), "-A", parts[1], "--no-out-link")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Warn(fmt.Sprintf("Could not resolve store path for %s: %s", pkgAttr, string(output)))
```
fluxxBot, line 295: why warn if we're returning error? also check across entire inmplementation, sometimes the errors are recorded as part of output and err only contains `exit code 1` string

artifactory/commands/nix/command.go
fluxxBot: why do we need this?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+
+func (c *NixCommand) GetStdWriter() io.WriteCloser { return nil }
+func (c *NixCommand) GetErrWriter() io.WriteCloser { return nil }
+func (c *NixCommand) CommandName() string           { return "rt_nix" }
+
+func (c *NixCommand) ServerDetails() (*config.ServerDetails, error) {
+	return c.serverDetails, nil
+}
+
+func saveBuildInfoLocally(buildInfo *entities.BuildInfo, projectKey string) error {
```
fluxxBot, line 927: why a function is required out of this? shouldn't we save only after the build-info is collected?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	}
+	return buildName, buildNumber, nil
+}
+
+// GetCmd returns the exec.Cmd for gofrogcmd.RunCmd interface.
+func (c *NixCommand) GetCmd() *exec.Cmd {
+	return exec.Command("nix", append([]string{c.nativeTool}, c.args...)...)
+}
+
+func (c *NixCommand) GetEnv() map[string]string {
```
fluxxBot: name should be a little more specific, don't you think?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+		return entities.Checksum{
+			Sha1:   item.Actual_Sha1,
+			Sha256: item.Sha256,
+			Md5:    item.Actual_Md5,
+		}
+	}
+	return entities.Checksum{}
+}
+
+func (c *NixCommand) getBuildNameAndNumber() (string, string, error) {
```
fluxxBot, line 906: a common utility might also be present, also are we not collecting project info from flag?

artifactory/commands/nix/command.go
fluxxBot: why are we making a call per file basis? let's group them per repo and make a single shot call to get all the details

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+	if c.serverDetails == nil {
+		return
+	}
+	servicesManager, err := utils.CreateServiceManager(c.serverDetails, -1, 0, false)
+	if err != nil {
+		return
+	}
+
+	props := fmt.Sprintf("build.name=%s;build.number=%s;build.timestamp=%s", buildName, buildNumber, timestamp)
+	searchQuery := fmt.Sprintf(`{"repo": "%s", "$or": [{"$and":[{"path": "%s","name": {"$match": "%s"}}]}]}`, repo, dirPath, namePattern)
```
fluxxBot: https://github.com/jfrog/jfrog-cli-artifactory/pull/442/changes#diff-9ef932742f174167bd4f2969755f537daa50c6562bacad7cdd71b66ee31701d9R653
findNarFiles function might already be giving you this info, consider caching it and reusing to reduce network calls

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+					return repo
+				}
+			}
+		}
+	}
+	return ""
+}
+
+// createNetrcFile creates a temporary netrc file for Nix authentication.
+func (c *NixCommand) createNetrcFile() {
```
fluxxBot: why are we doing this? are we not considering user's config files?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,768 @@
...
+		return fmt.Errorf("nix path-info failed: %w", err)
+	}
+
+	closurePaths := strings.Fields(strings.TrimSpace(string(output)))
+	log.Info(fmt.Sprintf("Found %d store path(s) in closure", len(closurePaths)))
+
+	timestamp := strconv.FormatInt(time.Now().UnixMilli(), 10)
+	var artifacts []entities.Artifact
+
+	for _, sp := range closurePaths {
```
fluxxBot: let's try to combine them in a single network call, check for better efficiency

fluxxBot (changes requested):

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+	c.nativeTool = tool
+	return c
+}
+
+func (c *NixCommand) SetArgs(args []string) *NixCommand {
+	c.args = args
+	return c
+}
+
+func (c *NixCommand) SetServerDetails(details *config.ServerDetails) *NixCommand {
```
naveenku-jfrog: Nit-pick:
serverDetails should be used in place of details.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+
+		if searchRepo != "" {
+			resolved := 0
+			for i, dep := range buildInfo.Modules[0].Dependencies {
+				storePath, ok := deps[dep.Id]
+				if !ok {
+					continue
+				}
+				storeHash := nixpkg.ExtractStoreHash(storePath)
+				dirPath := "binary-cache/" + storeHash
```
naveenku-jfrog: can we store this "binary-cache/" into some constant?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+		return nil, nil, fmt.Errorf("nix path-info failed: %w", err)
+	}
+
+	var pathInfoMap map[string]nixStorePathInfo
+	if err := json.Unmarshal(output, &pathInfoMap); err != nil {
+		return nil, nil, fmt.Errorf("parse nix path-info output: %w", err)
+	}
+
+	rootIDs := make(map[string]bool, len(rootPaths))
+	for _, sp := range rootPaths {
```
naveenku-jfrog: sp should be more descriptive.

artifactory/commands/nix/command.go
naveenku-jfrog: sp name should be more descriptive.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+
+	closurePaths := strings.Fields(strings.TrimSpace(string(output)))
+	log.Info(fmt.Sprintf("Found %d store path(s) in closure", len(closurePaths)))
+
+	timestamp := strconv.FormatInt(time.Now().UnixMilli(), 10)
+	var artifacts []entities.Artifact
+
+	for _, sp := range closurePaths {
+		storeHash := nixpkg.ExtractStoreHash(sp)
+		dirPath := "binary-cache/" + storeHash
```
naveenku-jfrog: use some constant: "binary-cache/"

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+			})
+		}
+
+		// narinfo artifact
+		narinfoName := storeHash + ".narinfo"
+		narinfoPath := dirPath + "/" + narinfoName
+		narinfoChecksum := c.getArtifactChecksums(c.repo, narinfoPath)
+		artifacts = append(artifacts, entities.Artifact{
+			Name:                   narinfoName,
+			Type:                   "narinfo",
```
naveenku-jfrog: If you are hard coding this "narinfo" for every artifact, then it is redundant.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+		log.Info(fmt.Sprintf("Tagged %d artifact(s) with build properties", len(artifacts)))
+	}
+
+	return nil
+}
+
+// findStorePathFromArgs finds a store path or ./result from the command args.
+func (c *NixCommand) findStorePathFromArgs() string {
+	for _, arg := range c.args {
+		if strings.HasPrefix(arg, "/nix/store/") {
```
naveenku-jfrog: constant should be used for "/nix/store/"

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+			if err == nil {
+				return resolved
+			}
+		}
+		// Check if it's a path that resolves to a store path
+		if !strings.HasPrefix(arg, "-") && !strings.HasPrefix(arg, "http") {
+			resolved, err := filepath.Abs(arg)
+			if err == nil {
+				link, err := os.Readlink(resolved)
+				if err == nil && strings.HasPrefix(link, "/nix/store/") {
```
naveenku-jfrog: constant should be used for "/nix/store/"

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+	return nil
+}
+
+// findStorePathFromArgs finds a store path or ./result from the command args.
+func (c *NixCommand) findStorePathFromArgs() string {
+	for _, arg := range c.args {
+		if strings.HasPrefix(arg, "/nix/store/") {
+			return arg
+		}
+		if arg == "./result" || arg == "result" {
```
naveenku-jfrog: constant should be used for "./result"

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+	if password == "" {
+		password = c.serverDetails.AccessToken
+	}
+	if user == "" || password == "" {
+		return fmt.Errorf("no credentials configured (need user+password or user+access-token)")
+	}
+
+	host := c.serverDetails.ArtifactoryUrl
+	if idx := strings.Index(host, "://"); idx != -1 {
+		host = host[idx+3:]
```
naveenku-jfrog: there is no check for length to be greater than 3.

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+func buildNixBuildInfo(buildName, buildNumber, projectName string, deps map[string]string, depGraph map[string][]string) *entities.BuildInfo {
+	requestedBy := make(map[string][]string)
+	for parent, children := range depGraph {
+		for _, child := range children {
+			requestedBy[child] = append(requestedBy[child], parent)
+		}
+	}
+
+	module := entities.Module{
+		Id:   projectName,
```
naveenku-jfrog, line 456: We are not using module flag value here?

artifactory/commands/nix/command.go
```diff
@@ -0,0 +1,824 @@
...
+		return c.runNixFlakeBuild()
+	case "copy":
+		return c.runNixCopy()
+	default:
+		return c.runPassthrough()
+	}
+}
+
+// runNixChannel executes "nix-channel" with all args. No build-info.
+func (c *NixCommand) runNixChannel() error {
```
naveenku-jfrog: Why there are different functions executing same type of statements?
1. runNixChannel
2. runNixEnv
3. runNixBuild
4. runNixFlakeBuild
5. runNixCopy
6. runPassthrough
All of the functions have following statements:
```
log.Info("Running nix-channel")
	cmd := exec.Command("nix-channel", c.args...)
	cmd.Env = c.buildEnv()
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
		return fmt.Errorf("nix-channel failed: %w", err)
	}
	return nil
```
Make one generic function and Just pass the command name in the function for executing separate commands.

naveenku-jfrog (changes requested):

naveenku-jfrog (dismissed):

fluxxBot (approved):

agrasth (approved):

naveenku-jfrog (approved):

PR #448, opened 2026-05-12
RTECO-1011 - Improve skill commands

udaykb2:
### Overview
Improve "skills list" and "skills update" command

**Improvement in SKILLS LIST**
- For every install inside skills, a file is created ./jfrog/skill-info.json, which contains information of (repo, version, slug, etc)
- For --check-updates, cli will check skill-info.json version to get the version info, repo & calls that particular repo to get the latest version

**Improvement in SKILLS UPDATE**
- supports --agent, --project-dir/--global
- Supports --path <direct_path> , when --path is given no need to provide --agent, --project-dir/--global
- Loads meta-data from skill-info.json to get version, fall back to SKILL.md if skill-info.json is not present
- multi-agent update (--agent takes multiple agents seperated by comma (,))

**Improvement in SKILLS Install**
- During installation of skill, all its meta data will be stored in .jfrog/skill-info.json , this file will be used by list, update, to know REPO from which it is downloaded, VERSION (no need to write version number inside SKILL.md) & other info

**RTECO-1011 is parent task of RTECO-1077 & RTECO-1078**
Since changes of both RTECO-1078 & 1077 are present in this PR, used RTECO-1011 in PR title

-----

### DIFF
--- a/skills/commands/update/update.go
+++ b/skills/commands/update/update.go
@@ -14,11 +14,11 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
-type preflight struct {
-	target       common.AgentTarget
-	installedVer string
-	upToDate     bool
-	failure      string
+type preUpdate struct {
+	agentTarget            common.AgentTarget
+	installedVersion       string
+	alreadyAtTargetVersion bool
+	failureReason          string
 }
 
 // RunUpdate is the CLI action for `jf skills update`.
@@ -65,7 +65,7 @@ func RunUpdate(c *components.Context) error {
 		return err
 	}
 
-	checks := preflightTargets(targets, targetVersion, force, quiet)
+	checks := preUpdateTargets(targets, targetVersion, force, quiet)
 	results, updatable := initialResultsAndUpdatable(checks, targetVersion)
 
 	if dryRun {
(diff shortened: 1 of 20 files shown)


skills/commands/update/update.go
```diff
@@ -14,135 +14,245 @@ import (
...
-func validateInstallBase(path string) error {
-	if err := common.ValidateExistingDir(path); err != nil {
-		return fmt.Errorf("install path %q: %w", path, err)
+func finalError(results []install.SummaryRow) error {
+	if len(results) == 0 {
+		return nil
+	}
+	for _, result := range results {
+		if result.Status != install.SummaryStatusFailed {
+			return nil
```
itsmeleela, line 237: This is intentional, partial success exits 0. But I think worth documenting explicitly in the command description (if possible) so CI users who rely on exit codes know to check the summary output. 

Also confirm - should a run where all targets are Skipped (none actually updated) exit 0 silently? Currently it does.

udaykb2: 1. All agents's skill is updated successfully -> No error
2. Few agesnts successful, few agents skipped(already up-to-date) -> No error
3. We will only return error, if update failed for at least one skill (eg: skill not found), but we will still update  the skills that are updatable.

udaykb2: Skipped -> only happens when already versions matched, so exiting silently.

Only if any one of the skill updation fails, then only we are returning error. (but we still update skill that can be updatable)

bhanurp (approved):

naveenku-jfrog (approved):

PR #449, opened 2026-05-13
RTECO-992 - Restore build-info for jf uv build, clear artifacts only

agrasth:
## Problem

PR #438 fixed duplicate dependency entries by skipping build-info entirely for `jf uv build`. This was too aggressive — it broke three tests that expected at least a module to exist:

- `TestUvBuild` — expects 1 module, 0 artifacts
- `TestUvArtifactTypeIsExtension` — expects module exists
- `TestUvCustomModule` — expects module with custom ID

## Fix

Restore build-info collection for `jf uv build`, following Maven's behavior for non-deploy goals:

- **Module**: recorded (enables `--module`, `jf rt bp`, custom naming)
- **Dependencies**: recorded (what was resolved to produce the wheel/sdist)
- **Artifacts**: cleared — `uv build` uploads nothing to Artifactory; artifacts are captured by `jf uv publish`

This matches how `jf mvn compile/package` works: dependencies are tracked, artifacts are only recorded when something is actually deployed.

## Changes

- `native_uv.go Run()`: removed the `!= "build"` guard so `uvGetBuildInfo` is called for `build`
- `native_uv.go uvGetBuildInfo()`: added `case "build"` block that clears `Artifacts` (keeps dependencies)

## Related

- Fixes issue introduced by #438
- Companion test fix in jfrog/jfrog-cli (TestUvArtifactTypeIsExtension updated to verify artifact types via `jf uv publish`)

### DIFF
--- a/artifactory/commands/python/native_uv.go
+++ b/artifactory/commands/python/native_uv.go
@@ -714,13 +714,18 @@ func uvGetBuildInfo(workingDir string, buildConfiguration *buildUtils.BuildConfi
 
 	// publish only records artifacts — deps come from the sync/install partial and
 	// must not be duplicated here without enrichment.
+	// TODO: loop over all modules if workspace support is added
 	if cmdName == "publish" && len(bi.Modules) > 0 {
 		bi.Modules[0].Dependencies = nil
 	}
-	// build creates local dist/ files but uploads nothing to Artifactory — record
-	// dependencies (what was resolved) but no artifacts, matching Maven's behavior
-	// for non-deploy goals.
+	// build creates local dist/ files but uploads nothing to Artifactory.
+	// deps are cleared to avoid duplicates when sync ran first under the same build
+	// name — jf rt bp merges by key(Id+Sha1+Md5+Scopes), so an unenriched build
+	// dep and an enriched sync dep produce two separate entries for the same package.
+	// artifacts are cleared because nothing is uploaded to Artifactory by build.
+	// TODO: loop over all modules if workspace support is added (same as publish block below)
 	if cmdName == "build" && len(bi.Modules) > 0 {
+		bi.Modules[0].Dependencies = nil
 		bi.Modules[0].Artifacts = nil
 	}
 
--- a/artifactory/commands/python/native_uv.go
+++ b/artifactory/commands/python/native_uv.go
@@ -127,10 +127,7 @@ func (c *NativeUVCommand) Run() error {
 				if biErr := uvGetScriptBuildInfo(scriptPath, c.buildConfiguration, serverDetails); biErr != nil {
 					log.Warn("Failed to collect UV script build info: " + biErr.Error())
 				}
-			} else if c.commandName != "build" {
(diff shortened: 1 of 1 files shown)


bhanurp (dismissed):

artifactory/commands/python/native_uv.go
```diff
@@ -127,10 +127,7 @@ func (c *NativeUVCommand) Run() error {
 				if biErr := uvGetScriptBuildInfo(scriptPath, c.buildConfiguration, serverDetails); biErr != nil {
 					log.Warn("Failed to collect UV script build info: " + biErr.Error())
 				}
-			} else if c.commandName != "build" {
-				// "build" only creates local dist/ files — nothing is uploaded to Artifactory,
-				// so it must not save a partial build-info. Saving unenriched deps from "build"
-				// would create duplicate dependency entries when merged with a prior "sync" partial.
+			} else {
```
itsmeleela, line 130: The removed comment flagged a dep-duplication risk for workflows that run jf uv sync then jf uv build under the same build name. Now that build enters the collection path, does jf rt bp deduplicate deps across partials? If yes, this is fine - please add a comment confirming. If no, we may want bi.Modules[0].Dependencies = nil in the build case too (same as publish clears deps, build would clear artifacts and deps).

artifactory/commands/python/native_uv.go
```diff
@@ -720,6 +717,12 @@ func uvGetBuildInfo(workingDir string, buildConfiguration *buildUtils.BuildConfi
 	if cmdName == "publish" && len(bi.Modules) > 0 {
 		bi.Modules[0].Dependencies = nil
 	}
+	// build creates local dist/ files but uploads nothing to Artifactory — record
+	// dependencies (what was resolved) but no artifacts, matching Maven's behavior
+	// for non-deploy goals.
+	if cmdName == "build" && len(bi.Modules) > 0 {
+		bi.Modules[0].Artifacts = nil
+	}
```
itsmeleela, line 730: Same [0] pattern as the publish case above, fine for now since UV produces a single module, but if we ever support workspace builds this will need a range loop. Consider a // TODO: loop over all modules if workspace support is added comment to match the implicit contract of the publish block.

itsmeleela (changes requested):

artifactory/commands/python/native_uv.go
agrasth: `jf rt bp` does NOT deduplicate — it keys by `Id+Sha1+Md5+Scopes`, so enriched (from sync) and unenriched (from build) deps are treated as different entries. Added Dependencies = nil to the build block to prevent duplicates.

artifactory/commands/python/native_uv.go
agrasth: Added // TODO: loop over all modules if workspace support is added, will create a ticket for this.

naveenku-jfrog (approved):
Please address comments.

bhanurp (approved):

itsmeleela (approved):

PR #452, opened 2026-05-14
Added logic for fat manifest file

naveenku-jfrog:
-----
What: Added logic for fat manifest file

### DIFF
--- a/artifactory/commands/ocicontainer/layerextract.go
+++ b/artifactory/commands/ocicontainer/layerextract.go
@@ -51,8 +51,9 @@ func ExtractLayersFromManifestData(candidateLayers map[string]*utils.ResultItem,
 
 // SearchLayersForDetailedSummary searches for container image layers in Artifactory
 // without using build info builders, returning layers for detailed summary display.
-// This function searches for layers, extracts them from the manifest, and returns them
-// in a format suitable for displaying a detailed summary.
+// This function searches for layers, extracts them from the manifest (single-platform
+// or fat / multi-platform manifest), and returns them in a format suitable for
+// displaying a detailed summary.
 func SearchLayersForDetailedSummary(image *Image, repo string, serviceManager artifactory.ArtifactoryServicesManager, imageSha256 string) (*[]utils.ResultItem, error) {
 	// Get repository details to determine searchable repo
 	repoDetails := &services.RepositoryDetails{}
@@ -79,6 +80,8 @@ func SearchLayersForDetailedSummary(image *Image, repo string, serviceManager ar
 
 	var resultMap map[string]*utils.ResultItem
 	var imageManifest *manifest
+	var foundResultMap map[string]*utils.ResultItem
+	isFatManifest := false
 
 	// Search for manifest and layers
 	for _, searchPath := range manifestPathsCandidates {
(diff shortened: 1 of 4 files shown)


artifactory/commands/ocicontainer/layerextract.go
```diff
@@ -139,10 +156,93 @@ func SearchLayersForDetailedSummary(image *Image, repo string, serviceManager ar
...
 
+	imageLayers := aggregateFatManifestLayers(fatManifestResult, fatManifest, multiPlatformImages)
 	return &imageLayers, nil
 }
+
+// aggregateFatManifestLayers collects every artifact that should be reported as
+// part of a multi-platform image: the fat manifest result item itself, plus all
+// layers belonging to each platform-specific manifest digest. Layers shared
+// across platforms are returned once.
+func aggregateFatManifestLayers(fatManifestResult *utils.ResultItem, fatManifest *FatManifest, multiPlatformImages map[string][]*utils.ResultItem) []utils.ResultItem {
```
fluxxBot, line 208: I think there's already a common method for this, can you make use of that? you might find it as part of the docker build workflow

fluxxBot (changes requested):

artifactory/commands/ocicontainer/layerextract.go
fluxxBot, line 1: also please take of the repo key extraction logic

fluxxBot (changes requested):

artifactory/commands/ocicontainer/layerextract.go
naveenku-jfrog: I did look at createMultiPlatformBuildInfo in buildinfo.go — that's the closest existing logic for fat-manifest layer aggregation in the docker build workflow. However, it can't be reused as-is from the detailed-summary path because it does more than just aggregate layers:
1. It mutates builder.imageLayers as a side-effect on the build-info builder.
2. It constructs per-platform buildinfo.Module artifacts and a *buildinfo.BuildInfo return value.
3. It calls setBuildProperties to write build properties back to Artifactory.

The detailed-summary flow only needs the layer list (no builder, no build-info, no prop writes), so calling it directly would pull in unwanted side-effects and the wrong return type.

artifactory/commands/ocicontainer/layerextract.go
naveenku-jfrog: This will be handled in https://jfrog-int.atlassian.net/browse/RTECO-1295

artifactory/commands/ocicontainer/layerextract.go
```diff
@@ -139,10 +156,93 @@ func SearchLayersForDetailedSummary(image *Image, repo string, serviceManager ar
...
+		if strings.Contains(err.Error(), "download blocking policy configured in Xray") {
+			log.Info("Artifact download blocked by Xray policy. Returning basic summary with available files.")
+			basicSummary := collectBasicSummary(resultMap)
+			return &basicSummary, nil
+		}
+		return nil, err
+	}
+	if fatManifest == nil {
+		return nil, errorutils.CheckErrorf("could not parse fat manifest in Artifactory")
+	}
```
fluxxBot, line 190: this part is already done when collecting build-info, why are we doing again? let's use the cache data and avoid expensive aql calls

fluxxBot (changes requested):

artifactory/commands/ocicontainer/layerextract.go
fluxxBot: such calls are very expensive in docker

naveenku-jfrog: Build info collection comes after this, not before this. This is how exactly non-fat manifest file process works, so implemented the same way.

fluxxBot (approved):

reshmifrog (approved):

PR #458, opened 2026-05-25
Updated logic for extracting repository key in case of reverse proxy scenario

naveenku-jfrog:
-----
What: Updated logic for extracting repository key so that reverse proxy scenarios can also be handled.

### DIFF
--- a/artifactory/commands/container/push.go
+++ b/artifactory/commands/container/push.go
@@ -113,7 +113,7 @@ func (pc *PushCommand) Run() error {
 		return err
 	}
 	if !toCollect && pc.IsDetailedSummary() {
-		repo, err := pc.image.ExtractArtifactoryRepoKey()
+		repo, err := pc.GetRepo()
 		if err != nil {
 			return err
 		}
--- a/artifactory/commands/container/push_test.go
+++ b/artifactory/commands/container/push_test.go
@@ -1,0 +1,208 @@
+package container_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"sync/atomic"
+	"testing"
+
+	container "github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/container"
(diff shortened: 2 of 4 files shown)


itsmeleela (dismissed):

itsmeleela:
@naveenku-jfrog Commit message should explicitly talk about reverse proxy - please update the same

artifactory/commands/container/push.go
```diff
@@ -113,7 +113,7 @@ func (pc *PushCommand) Run() error {
 		return err
 	}
 	if !toCollect && pc.IsDetailedSummary() {
-		repo, err := pc.image.ExtractArtifactoryRepoKey()
```
fluxxBot, line 116: please remove this function, this is misleading for any future dev

artifactory/commands/container/push.go
```diff
@@ -113,7 +113,7 @@ func (pc *PushCommand) Run() error {
...
+		repo, err := pc.GetRepo()
```
fluxxBot, line 116: also please replace this function usage to other places as well, probably pull.go and such other files

fluxxBot (changes requested):

artifactory/commands/container/push.go
itsmeleela, line 116: If we dont have tests to cover reverse proxy scenario - we need to add them

naveenku-jfrog: Addressed it.

artifactory/commands/container/push.go
naveenku-jfrog: Addressed it.

artifactory/commands/container/push.go
naveenku-jfrog: Addressed it.

fluxxBot (approved):

agrasth (approved):

PR #460, opened 2026-05-26
RTECO-1352 - Fix jf setup docker login when using --url

Edenb2131:
- [x] All [tests](https://github.com/jfrog/jfrog-cli-core#tests) passed. New table-driven tests added for the affected helper.
- [x] All [static analysis checks](https://github.com/jfrog/jfrog-cli-core/actions/workflows/analysis.yml) passed locally (`gofmt`, `go vet`).
- [x] This pull request is on the dev branch.

-----

## Summary

`jf setup docker` (and `jf setup podman`) fail immediately when invoked with `--url`. The CLI calls `docker login ""`, the daemon resolves the empty registry to Docker Hub, and the user sees `failed to login to container registry: docker login failed for: .`

Reproduces on **both SaaS and self-hosted** — CLI-only defect, independent of Artifactory version. Tracked as **RTECO-1352**.

## Reproduction

```bash
jf setup docker --url https://<tenant>.jfrog.io/artifactory \
  --user <user> --access-token <token> --repo <docker-virtual-repo>
```

## Root cause

`configureContainer` derived the registry host from `sc.serverDetails.GetUrl()`. But `createServerDetailsFromFlags` (jfrog-cli) clears `details.Url` for the Rt command domain after copying it into `ArtifactoryUrl`:

```go
case commonCliUtils.Rt:
    details.ArtifactoryUrl = details.Url
...
details.Url = ""
```

So on the `--url` path, `GetUrl()` returns `""`, `url.Parse("")` yields an empty URL, and `ContainerManagerLogin` is called with an empty registry — the daemon then defaults to Docker Hub.

The bug only manifests on `--url`. The `--server-id` path worked because saved configs populate `Url` from disk.

## Fix

Extract a `deriveContainerRegistryHost` helper that:
- Reads `GetArtifactoryUrl()` first (populated on the `--url` path), falls back to `GetUrl()` (populated on `--server-id`).
- Returns `parsedURL.Host` directly — `docker login` takes only `host[:port]`.
- Returns an explicit `"server URL is empty; provide --url or --server-id"` when both are empty, instead of silently redirecting to Docker Hub.

## Tests

12 table-driven subtests in `TestDeriveContainerRegistryHost` covering: `--url` path, `--server-id` fallback, both-set precedence, ports, http scheme, self-hosted IP (with and without port), IPv6, subdomain registry method, embedded-credentials guard, no-scheme typo, both-empty. All pass.

## Backward compatibility

No public API or flag changes. `--server-id` success path unchanged. The only user-visible change is a clearer error in the both-URLs-empty case.

Depends on:
1. jfrog-cli - https://github.com/jfrog/jfrog-cli/pull/3517


artifactory/commands/setup/setup.go
```diff
@@ -449,6 +447,31 @@ func (sc *SetupCommand) configureContainer() error {
...
+	rawUrl := artifactoryUrl
+	if rawUrl == "" {
+		rawUrl = platformUrl
+	}
+	parsedUrl, err := url.Parse(rawUrl)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse server URL: %w", err)
+	}
+	if parsedUrl.Host == "" {
+		return "", errorutils.CheckErrorf("server URL is empty; provide --url or --server-id")
```
fluxxBot: When the URL is missing a scheme (e.g. `acme.jfrog.io/artifactory/`), `url.Parse` succeeds but treats the entire string as a path, so `parsedUrl.Host` ends up empty. The error then says "server URL is empty" even though the user passed a non-empty value, which is misleading.

Could we detect the no-scheme case explicitly and surface a clearer message? Something like:

```go
if parsedUrl.Scheme == "" {
    return "", errorutils.CheckErrorf("server URL %q is missing a scheme; expected http:// or https://", rawUrl)
}
if parsedUrl.Host == "" {
    return "", errorutils.CheckErrorf("server URL is empty; provide --url or --server-id")
}
```

This also lets the existing `"URL without scheme returns explicit error"` test case assert against a more accurate message.

artifactory/commands/setup/setup.go
fluxxBot, line 1: please add a e2e test related to this jfrog-cli and replace jfrog-cli-artifactory directive in jfrog-cli go mod to run the tests and check the outcome

fluxxBot (changes requested):

fluxxBot (dismissed):

fluxxBot (approved):

bhanurp (approved):

PR #463, opened 2026-05-28
RTECO-1272 - Implement agent plugins install command

udaykb2:
Implement agent plugins install command

- If --version is not provided, we will use \<AGENT\>-marketplace.json to download the plugin
- If --version is provided, we will directly download plugin from repository
- --harness takes only single agent name
- Currently we are supporting claude, cursor, codex only
- If user want to include new agent, he can edit ./jfrog/agents/agent-config.json under plugins-agents key and add new agents
- we are supporting --path (raw_path) where user can point to a folder to install directly into that folder. Here since --harness is not provided, we will directly download that plugin from repo
- we are storing some metadata under .jfrog/plugins-info.json of each plugin, so we will have information of repo, version, harness info, which will be used by list & update command to check for updates & to update all if they are behind the latest version
- Common code between agents & plugins moved inside /agent/common to reduce redundency

-----


agent/common/agent_config.go
```diff
@@ -0,0 +1,67 @@
...
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
+)
+
+// Keys recognized inside ~/.jfrog/agents/agent-config.json.
```
bhanurp: is it matching ~/.jfrog/agents/skills-config.json

agent/common/agent_config.go
```diff
@@ -0,0 +1,67 @@
...
+		return "", err
+	}
+	return filepath.Join(home, agentsConfigSubdir, agentsConfigFile), nil
+}
+
+// LoadAgentConfigSection reads ~/.jfrog/agents/agent-config.json and returns the
+// raw JSON for the requested top-level key. When the file or the key is missing
+// it returns (nil, path, nil). The resolved path is always returned so callers
+// can include it in error messages.
+func LoadAgentConfigSection(key string) (json.RawMessage, string, error) {
```
bhanurp, line 51: what will happen incase agents-config.json or skills-config.json is not present do throw error? Incase user wants to continue without any config what will be the UX?

agent/plugins/common/metadata.go
```diff
@@ -24,17 +24,62 @@ const manifestJSONIndent = "    "
...
+	var fromConfig []string
+	if err := json.Unmarshal(section, &fromConfig); err != nil {
+		return nil, fmt.Errorf("failed to parse %q in %s: %w", agentcommon.PluginManifestPathsKey, path, err)
+	}
+	return mergePluginManifestPaths(fromConfig), nil
+}
+
+// mergePluginManifestPaths prepends config paths, then appends built-in defaults once each.
+func mergePluginManifestPaths(fromConfig []string) []string {
+	seen := make(map[string]struct{})
```
bhanurp: What is seen? What exactly it holds? May be name should be changed to what exactly it is holding?

agent/plugins/common/metadata.go
```diff
@@ -24,17 +24,62 @@ const manifestJSONIndent = "    "
...
+	merged := make([]string, 0, len(fromConfig)+len(KnownManifestRelPaths))
```
bhanurp: similar to seen, even this should be named something meaningful!

agent/common/agents.go
```diff
@@ -0,0 +1,155 @@
...
+	for name := range registry {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	return strings.Join(names, ", ")
+}
+
+// AgentRegistryHelp lists agents/paths and how to edit agent-config.json.
+func AgentRegistryHelp(registry map[string]AgentSpec, helpExample AgentRegistryHelpExample) string {
+	configPath, _ := AgentConfigPath()
```
bhanurp: why is the error ignored here? but considered in other invocation.

agent/common/install_fs.go
```diff
@@ -0,0 +1,282 @@
...
+
+const (
+	// PackageDownloadThreads is the parallel download count for package and marketplace files.
+	PackageDownloadThreads = 1
+	// PackageDownloadRetries is the HTTP retry count for package and marketplace downloads.
+	PackageDownloadRetries = 3
+)
+
+// InstallInfoManifestSchemaVersion is bumped when the JSON shape changes incompatibly.
+const InstallInfoManifestSchemaVersion = 1
```
bhanurp: cant we group all constants 
```
const (
    jfrogInstallDirName = ".jfrog"
    // PackageDownloadThreads is the parallel download count for package and marketplace files.
	PackageDownloadThreads = 1
	// PackageDownloadRetries is the HTTP retry count for package and marketplace downloads.
	PackageDownloadRetries = 3
    // InstallInfoManifestSchemaVersion is bumped when the JSON shape changes incompatibly.
    const InstallInfoManifestSchemaVersion = 1
)
```

agent/common/install_fs.go
```diff
@@ -0,0 +1,282 @@
...
+	Repo             string `json:"repo"`
+	Slug             string `json:"slug"`
+	InstalledVersion string `json:"installedVersion"`
+	Scope            string `json:"scope"`
+	Agent            string `json:"agent"`
+	ProjectDir       string `json:"projectDir,omitempty"`
+}
+
+// InstallInfoManifestPath returns the path to the manifest under an install directory.
+func InstallInfoManifestPath(installDir, manifestFileName string) string {
```
bhanurp: does it has to be exposed? cant it be `installInfoManifestPath` since usage is only in this file?
Apply where ever needed functions should be exposed only when needed outside the package

agent/common/install_fs.go
```diff
@@ -0,0 +1,282 @@
...
+
+		if info.IsDir() {
+			return os.MkdirAll(destPath, info.Mode())
+		}
+
+		return copyFile(path, destPath)
+	})
+}
+
+func copyFile(src, dst string) error {
```
bhanurp: better to move file operations into a separate file under common like file_operations.go
which can have create, copy, move, delete etc

agent/common/install_fs.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+
+		if err := extractZipEntry(entry, entryPath); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func extractZipEntry(entry *zip.File, outputPath string) error {
```
bhanurp: lets move zip related changed to handle_zip.go

agent/common/install_fs.go
```diff
@@ -0,0 +1,282 @@
...
+			return fmt.Errorf(
+				"failed to create install destination %q: %w. "+
+					"Create the directory at that path (including parent folders if needed), then run the command again",
+				dest, mkErr,
+			)
+		}
+		return nil
+	default:
+		return fmt.Errorf("install destination %q is not accessible: %w", dest, err)
+	}
```
bhanurp:
```suggestion
	switch {
        case err == nil:
                if !info.IsDir() {
                        return fmt.Errorf("install destination %q exists and 
  dest)
                }
                return nil
        case errors.Is(err, fs.ErrNotExist):
                // #nosec G301 -- install files need to be readable across th
  tools.
                if mkErr := os.MkdirAll(dest, InstallDirMode); mkErr != nil {
                        return fmt.Errorf(
                                "failed to create install destination %q: %w.
                                        "Create the directory at that path (i
  then run the command again",
                                dest, mkErr,
                        )
                }
                return nil
        default:
                return fmt.Errorf("install destination %q is not accessible: 
  err)
        }
```
this approach is lot cleaner

agent/common/path.go
```diff
@@ -0,0 +1,124 @@
...
+
+// ValidateExistingDir requires path to exist and be a directory (after filepath.Abs).
+func ValidateExistingDir(path string) error {
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return fmt.Errorf("invalid path %q: %w", path, err)
+	}
+	info, err := os.Stat(abs)
+	if err != nil {
+		return fmt.Errorf("path %q: %w", path, err)
```
bhanurp: if the err is description enough you can return directly instead of wrapping else add more descriptive message and then wrap
indicating only path in error is not sufficient enough

agent/common/path.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// PathAgentName is the synthetic agent name used when --path selects the install target.
+const PathAgentName = "(path)"
```
bhanurp: why is file name path? you want to create something like file_operations.go?

agent/plugins/commands/install/install.go
```diff
@@ -0,0 +1,323 @@
...
+//   - --version "" + path-only → ListPluginVersions on Artifactory, pick latest semver (same as skills)
+func (ic *InstallCommand) resolveVersion() (string, error) {
+	requested := strings.TrimSpace(ic.version)
+	if requested != "" && requested != "latest" {
+		return requested, nil
+	}
+	if requested == "latest" || ic.agent.Name == "" {
+		return resolveLatestPluginVersion(ic.serverDetails, ic.repoKey, ic.slug)
+	}
+	// Empty --version with harness: resolve from marketplace.
```
bhanurp: Empty version with harness is always latest. Why do we need to resolve? harness-plugin.json will serve the latest version, is this checking latest version?

agent/plugins/commands/install/install.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// resolveVersion picks the version to install based on the requested --version and --harness.
+//
+//   - --version 1.0.0 (exact) → use directly
+//   - --version latest         → ListPluginVersions on Artifactory, pick latest semver
+//   - --version "" + harness   → download <harness>-marketplace.json, look up slug, use that version,
+//     then delete marketplace.json (deferred cleanup)
+//   - --version "" + path-only → ListPluginVersions on Artifactory, pick latest semver (same as skills)
+func (ic *InstallCommand) resolveVersion() (string, error) {
```
bhanurp, line 168: when harness is not give since it is not mandatory flag and we are in interactive terminal does jfrog-cli asks to select harness?

What will happen if CI=true?

agent/common/install_fs.go
```diff
@@ -0,0 +1,282 @@
...
+		return fmt.Errorf("illegal file path: %s", outputPath)
+	}
+
+	entryReader, err := entry.Open()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		// Best-effort close after zip entry read.
+		_ = entryReader.Close()
```
fluxxBot: unnamed err

Suggested fix:
 ```
 defer func() {
      if cerr := outputFile.Close(); cerr != nil && err == nil {
          err = fmt.Errorf("close extracted file: %w", cerr)
      }
  }() 
```
  (Named return err needed, or return close error in a second pass.)

File: agent/common/install_fs.go [hunk 16]
```diff
@@ -0,0 +1,282 @@
...
+
+func copyFile(src, dst string) error {
+	// #nosec G304 -- src comes from a vetted unzip temp directory
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		// Best-effort close after copy read.
+		_ = in.Close()
```
fluxxBot: defer func() {
      if cerr := out.Close(); cerr != nil && err == nil {
          err = fmt.Errorf("close copy destination %s: %w", dst, cerr)
      }
  }()

agent/common/agent_config.go
```diff
@@ -0,0 +1,67 @@
+package common
```
bhanurp, line 1: instead of common lets change it to utils which is in synchronous with other packages. WDYT?

agent/plugins/common/install_targets.go
```diff
@@ -0,0 +1,90 @@
...
+	ScopeGlobal  = agentcommon.InstallScopeGlobal
+	ScopePath    = agentcommon.InstallScopePath
+)
+
+type AgentTarget = agentcommon.InstallTarget
+
+// ValidateInstallFlags validates `--path | (--harness [, --project-dir | --global])` for plugins install.
+// absoluteInstallBaseDir is non-empty when --path was supplied; otherwise spec is resolved from --harness.
+// Unlike skills install, --harness accepts exactly one agent name.
+func ValidateInstallFlags(c *components.Context) (absoluteInstallBaseDir string, spec AgentSpec, projectDirAbs string, isGlobal bool, err error) {
```
fluxxBot: nit: I would suggest to return a struct instead of these many values from this function

agent/plugins/common/agents.go
```diff
@@ -0,0 +1,40 @@
...
+// RegistryHelp configures agent-config.json help text for plugins harness resolution.
+var RegistryHelp = agentcommon.AgentRegistryHelpExample{
+	ConfigSectionKey:  agentcommon.PluginsAgentsKey,
+	ExampleProjectDir: ".my-agent/plugins",
+	ExampleGlobalDir:  "~/.my-agent/plugins",
+}
+
+// ParseSingleHarness parses a single harness name from --harness. Comma-separated lists
+// are rejected because plugins install targets exactly one agent.
+func ParseSingleHarness(raw string) (string, error) {
```
bhanurp: since flag supports multiple harness how are we installing in all given harness target paths when we return only one from here.

agent/common/agent_config_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func withJfrogHome(t *testing.T) string {
+	t.Helper()
+	dir := t.TempDir()
+	t.Setenv(coreutils.HomeDir, dir)
+	return dir
+}
```
fluxxBot: we must already be having this function defined in the codebase, try to use that

agent/plugins/common/agents_test.go
```diff
@@ -0,0 +1,139 @@
...
+	t.Setenv(coreutils.HomeDir, dir)
+	return dir
+}
+
+func writeAgentConfig(t *testing.T, home, body string) {
+	t.Helper()
+	path := filepath.Join(home, "agents", "agent-config.json")
+	require.NoError(t, os.MkdirAll(filepath.Dir(path), 0o755))
+	require.NoError(t, os.WriteFile(path, []byte(body), 0o644))
+}
```
fluxxBot: duplicated functions

fluxxBot (changes requested):

fluxxBot (approved):

fluxxBot (dismissed):

agent/common/install_fs.go
udaykb2: Sure, but these changes have to be done for all commands (not just install), so I will push a separate PR for refactoring, which apply this change for other commands as well.

agent/common/install_fs.go
udaykb2: same, will be covered in Refactoring PR, for every command

agent/plugins/common/install_targets.go
udaykb2: done

agent/common/agent_config.go
udaykb2: Earlier file name is agent-config.json
I am reusing the same file for plugins also.

File struct looks like

{
skills-agents:{
     cursor:{projectDir:"somepath", globalDir:"somepath"}
},
plugins-agents:{
     cursor:{projectDir:"somepath", globalDir:"somepath"}
},
plugin-manifest-paths:[ "/path/to/plugin.json" ]
}

agent/common/agent_config.go
udaykb2: We do not error if ~/.jfrog/agents/agent-config.json is missing, empty, or lacks skills-agents / plugins-agents / plugin-manifest-paths. LoadAgentConfigSection returns (nil, path, nil) and callers fall back to built-ins.
UX without config: jf agent plugins install / jf agent skills install work out of the box with default harness paths; no file is required.

We only error when the file exists but is unreadable/unparseable, or when a present section has invalid content.

agent/plugins/commands/install/install.go
udaykb2: --harness cursor (no --version is given)

we will only install if that plugin is present in cursor-marketplace.json, else we will ask user to use --version to download anyway

agent/plugins/commands/install/install.go
udaykb2: If neither --path nor --harness is passed, we error in both interactive and CI... no harness prompt. --harness supports multiple agents via comma-separated names, so we don’t prompt for a single choice; the user must specify the target mode on the command line.

agent/common/agent_config.go
udaykb2: Sure, as @agrasth  is also working no this code base, I will make sure to take a separate PR dedicated to remove redundancy, nit picking, adding DPs.

udaykb2:
> Screenshot 2026-05-29 at 4 56 11 PM Screenshot 2026-05-29 at 4 56 27 PM Screenshot 2026-05-29 at 4 56 50 PM

Thank you for pointing this out.

As @agrasth is also working on plugins in parallel, making changes to these might affect his PRs as well, plus already this PR became too large

I would like to take a separate PR, dedicated to removing redundancy, adding DPs, using templates, making code consistent between skills & plugins

agent/plugins/common/metadata.go
udaykb2: updated

agent/plugins/common/agents.go
udaykb2: handling multi-harnesses now

agent/common/path.go
udaykb2: done

agrasth (approved):

agent/common/agent_config.go
udaykb2: going with other file, make it give name as agent-plugins-config.json for plugins & agent-config.json for skills. So I have chosen the single file approach

bhanurp (approved):

PR #464, opened 2026-05-29
Add delete Subcommand to jf agent plugins for Removing Plugin Versions from Artifactory

agrasth:
-----
## Summary

  - Add `jf agent plugins delete <slug> --version <version>` command to delete a specific agent plugin version from Artifactory
  - Extract shared HTTP DELETE logic into `agentcommon.DeleteVersion` in `agent/common/storage.go` so both skills and plugins reuse it
  - Refactor `DeleteSkillVersion` in `agent/skills/common/xray_gate.go` to delegate to the shared utility

  ## Usage

  ```bash
  # Delete a specific plugin version
  jf agent plugins delete <slug> --version <version> --repo <repo>

  # Preview what would be deleted without making any changes
  jf agent plugins delete <slug> --version <version> --repo <repo> --dry-run
```
  Repo can also be set via JFROG_AGENT_PLUGINS_REPO env var.

### DIFF
--- a/agent/common/storage.go
+++ b/agent/common/storage.go
@@ -9,7 +9,28 @@ import (
 
 	"github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
+	clientutils "github.com/jfrog/jfrog-client-go/utils"
 )
+
+// DeleteVersion deletes the entire version directory for a package (plugin, skill, etc.)
+// at {artURL}{repoKey}/{slug}/{version}/ using an HTTP DELETE request.
+func DeleteVersion(serverDetails *config.ServerDetails, repoKey, slug, version string) error {
+	sm, err := utils.CreateServiceManager(serverDetails, 3, 0, false)
+	if err != nil {
+		return fmt.Errorf("failed to create service manager for deletion: %w", err)
+	}
+	artURL := clientutils.AddTrailingSlashIfNeeded(sm.GetConfig().GetServiceDetails().GetUrl())
+	deletePath := fmt.Sprintf("%s%s/%s/%s/", artURL, repoKey, slug, version)
+	httpDetails := sm.GetConfig().GetServiceDetails().CreateHttpClientDetails()
+	resp, body, err := sm.Client().SendDelete(deletePath, nil, &httpDetails)
+	if err != nil {
+		return fmt.Errorf("failed to delete %s/%s/%s: %w", repoKey, slug, version, err)
+	}
+	if resp.StatusCode >= 400 {
+		return fmt.Errorf("failed to delete %s/%s/%s: HTTP %d — %s", repoKey, slug, version, resp.StatusCode, string(body))
+	}
+	return nil
+}
 
 // ErrVersionExistenceUnknown indicates the version path could not be checked (e.g. network
 // error or an Artifactory response whose HTTP status could not be read). Callers should not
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: agent/common/storage.go [hunk 1]
```diff
@@ -9,8 +9,29 @@ import (
 
 	"github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
+	clientutils "github.com/jfrog/jfrog-client-go/utils"
 )
 
+// DeleteVersion deletes the entire version directory for a package (plugin, skill, etc.)
+// at {artURL}{repoKey}/{slug}/{version}/ using an HTTP DELETE request.
+func DeleteVersion(serverDetails *config.ServerDetails, repoKey, slug, version string) error {
```
udaykb2, line 17: Is this function doing the same work of deleteSkillVersionWithManager?  If yes, how about removing redundant function in xray_gate.go and directly call this function?

cliutils/flagkit/flags.go
```diff
@@ -892,6 +893,9 @@ var commandFlags = map[string][]string{
 	SkillsDelete: {
 		url, user, password, accessToken, serverId, repo, version, dryRun,
 	},
+	AgentPluginsDelete: {
```
udaykb2: Lets group plugins together & skills together.

Lets move this declaration under AgentPluginsPublish: {}

agent/skills/common/xray_gate.go
```diff
@@ -167,13 +167,8 @@ func handleBlocked(sm artifactory.ArtifactoryServicesManager, params XrayGatePar
...
 
 // DeleteSkillVersion deletes the entire version directory for a skill.
-// Creates its own service manager. For callers that already have one, use deleteSkillVersionWithManager.
 func DeleteSkillVersion(serverDetails *config.ServerDetails, repoKey, slug, version string) error {
-	sm, err := utils.CreateServiceManager(serverDetails, 3, 0, false)
-	if err != nil {
-		return fmt.Errorf("failed to create service manager for deletion: %w", err)
-	}
-	return deleteSkillVersionWithManager(sm, repoKey, slug, version)
+	return agentcommon.DeleteVersion(serverDetails, repoKey, slug, version)
```
udaykb2, line 170: This function is just passing exact variables to other function, how about using agentcommon.DeleteVersion directly in the place where this function is getting called?

cliutils/flagkit/flags.go
agrasth: make sense, doing it.

agent/common/storage.go
agrasth: I think `deleteSkillVersionWithManager` serves a distinct purpose, it's an internal helper for the Xray auto-delete path that already has a service manager in `scope. agentcommon.DeleteVersion` is the public API for callers that only have serverDetails. 
So it looked like it's not doing the same job from the same context. 

I can add a comment to deleteSkillVersionWithManager making this explicit, WDYT?
```
  // deleteSkillVersionWithManager is used by the Xray gate path which already holds a
  // service manager; reusing it avoids creating a second connection. For callers with
  // only serverDetails, use agentcommon.DeleteVersion instead.
  func deleteSkillVersionWithManager(sm artifactory.ArtifactoryServicesManager, ...) error {
```

udaykb2 (dismissed):

bhanurp (approved):

udaykb2 (approved):

PR #471, opened 2026-06-01
RTECO-1079 - Remove updating plugins during publish

udaykb2:
-----

### Summary:
Earlier we used to update plugin.json with version that we are publishing.
We should not do it, as there can be many plugin.json files, which cause inconsistent values in different files


naveenku-jfrog (approved):

agrasth (approved):
Approved.

PR #473, opened 2026-06-01
RTECO-1079 - Consistent plugin identification

udaykb2:
-----
### Summary:

In plugins update
 - for --all, if we don't find plugins.json, plugin-info.json, we will skip that folder & inform the use that we couldn't find plugin.json, so we are not considering that folder as plugin
 - for --slug <slug_name> , same as above

- for both list & update, we are maintaining consistent behaviour, first we check .jfrog/plugin-info.json, fall back to plugin.json

### DIFF
--- a/agent/plugins/common/plugin_info_meta.go
+++ b/agent/plugins/common/plugin_info_meta.go
@@ -66,6 +66,7 @@ func pluginSlugsFromInstallDirEntries(installDir string, entries []os.DirEntry) 
 		}
 		pluginDir := filepath.Join(installDir, name)
 		if _, err := ReadInstalledPluginVersion(pluginDir); err != nil {
+			log.Warn(fmt.Sprintf("Skipping plugin '%s': %s", name, err.Error()))
 			continue
 		}
 		slugs = append(slugs, name)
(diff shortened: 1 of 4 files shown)


agent/plugins/common/plugin_info_meta.go
```diff
@@ -66,6 +66,7 @@ func pluginSlugsFromInstallDirEntries(installDir string, entries []os.DirEntry)
 		}
 		pluginDir := filepath.Join(installDir, name)
 		if _, err := ReadInstalledPluginVersion(pluginDir); err != nil {
+			log.Warn(fmt.Sprintf("Skipping plugin '%s': %s", name, err.Error()))
```
naveenku-jfrog, line 69: Avoid sprintf inside log.warn.

naveenku-jfrog (approved):

agent/plugins/common/plugin_info_meta.go
udaykb2: As this is common across agent/ , will be updating all of them at once in refactoring PR

agrasth (approved):
the `preUpdateTargets` change removes the `!quiet` gate, meaning warnings now always print even in quiet mode. Worth checking if that's intentional, but given it's a log.Warn and quiet mode typically suppresses log.Info not log.Warn, this is likely fine.

PR #484, opened 2026-06-07
XRAY-139364 - Automatically set VSC properties from local git on uploaded artifacts

attiasas:
## `feat(artifactory): attach local git VCS properties to published artifacts (XRAY-139364)`

Following:
* https://github.com/jfrog/jfrog-cli-artifactory/pull/348

## Summary
Extends artifact VCS property collection beyond CI environment variables. When CI VCS props are missing or incomplete, the CLI now discovers local git metadata (`vcs.url`, `vcs.revision`, `vcs.branch`) from the working directory and attaches it to artifacts during upload/publish flows. User-provided properties still take precedence over both CI and local git values.

## Changes
- **Core VCS utilities** (`artifactory/utils/civcs/`):
  - Add local git discovery via upstream `.git` search
  - Add `MergeWithUserProps` and `SetCIVcsPropsToConfig` with precedence: user props > CI props > local git props
  - Refactor VCS property keys to shared constants; add `DeriveSearchDirFromFileSpec` for path pattern resolution
- **`artifactory/utils/vcs.go`**: Add `FindDotGit` to locate a git repo root from a given directory
- **Publish/upload commands wired to local VCS fallback**


artifactory/utils/civcs/civcs.go
```diff
@@ -132,31 +146,81 @@ func hasProp(props, key string) bool {
...
-	if ciVcsInfo.Branch != "" && !vConfig.IsSet(VcsBranchKey) {
-		vConfig.Set(VcsBranchKey, ciVcsInfo.Branch)
+	if info.Repo != "" && !vConfig.IsSet(VcsRepoKey) {
+		vConfig.Set(VcsRepoKey, info.Repo)
 	}
 }
+
+func DeriveSearchDirFromFileSpec(pattern string, isRegexp bool) string {
+	if isRegexp {
+		return "."
```
reshmifrog, line 194: what does this mean , root directory i'm guessing?

reshmifrog (changes requested):

artifactory/utils/civcs/civcs.go
attiasas: Assume root looks like:
`/a/b`
`/c`

Let's break it down to components:
`/` -> root
`/a` -> in (a)
`.` -> THIS dir path
`/a/./` -> still in /a
`/a/./b` -> in /a/b
`..` -> go "up" one level
`/a/./b/..` -> /a/b/.. -> /a
`/a/./b/../..` -> /a/.. -> /
`/a/./b/../../c` -> /c

So every directory has an implicit '.' directory that refers to itself, and an implicit '..' directory that refers to its parent.

artifactory/utils/civcs/civcs.go
```diff
@@ -132,31 +146,81 @@ func hasProp(props, key string) bool {
...
+	return detectedLocalGitProps(hasProp(props, VcsUrlKey), hasProp(props, VcsRevisionKey), hasProp(props, VcsBranchKey))
+}
+
+func configIncludeAllLocalGitProps(vConfig *viper.Viper) bool {
+	return detectedLocalGitProps(vConfig.IsSet(VcsUrlKey), vConfig.IsSet(VcsRevisionKey), vConfig.IsSet(VcsBranchKey))
+}
+
+func detectedLocalGitProps(hasUrl bool, hasRevision bool, hasBranch bool) bool {
+	return hasUrl && hasRevision && hasBranch
+}
```
agrasth, line 222: `detectedLocalGitProps` adds an extra call frame and 3 unnamed bool parameters for what is just &&. Inline it. 
When a 4th local-git prop is added later, a developer has to find and update 3 sites. Inline the condition directly in each of the two callers, it's one clear line each. WDYT?

artifactory/utils/vcs.go
```diff
@@ -132,6 +148,58 @@ func GetDotGit(providedDotGitPath string) (string, error) {
...
+		if info, statErr := os.Stat(absPath); statErr == nil && !info.IsDir() {
+			absPath = filepath.Dir(absPath)
+		}
+	}
+	// Save the current working directory.
+	origWd, err := os.Getwd()
+	if err != nil {
+		return "", errorutils.CheckError(err)
+	}
+	defer func() {
```
agrasth: `os.Chdir` changes the process-wide working directory. If two goroutines call `FindDotGit` concurrently (e.g., parallel upload workers in `jfrog-cli-core`), they race on CWD: goroutine A saves CWD, goroutine B saves CWD=A's dir, A Chdirs to dirA, B Chdirs to dirB, A's deferred restore puts it back to A's dir — but B's deferred restore now restores to A's dir, not B's original dir. Net result: CWD corruption, wrong .git discovery, or relative-path breakage for all subsequent operations.

Upload is currently sequential per file spec, so this doesn't fire today. But it's a trap for any future parallelism in the callers. The right fix is to walk upward using string operations (loop `filepath.Dir()`) without any `Chdir`.

agrasth (dismissed):

artifactory/utils/civcs/civcs.go
attiasas: done

artifactory/utils/vcs.go
attiasas: changed, also in GetDotGit func

agrasth (approved):

reshmifrog (approved):

PR #485, opened 2026-06-07
Expend vsc detection

attiasas:
-----

Depends on:
* https://github.com/jfrog/jfrog-cli-artifactory/pull/484

## Changes
  - `rt upload` (generic upload)
  - `rt bp` (buildinfo publish)
  - Conan upload, Python twine/uv, Nix copy
  - Flexpack (Gradle/Maven), npm/pnpm publish, Go publish
  - Gradle/Maven extractors


artifactory/commands/conan/artifacts.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
+var mergeVcsPropsForConan = civcs.MergeWithUserProps
+
```
reshmifrog, line 21: we need to have integration tests for all the changes done here in jfrog-cli , please add it

artifactory/commands/pnpm/publish.go
```diff
@@ -569,30 +570,33 @@ func (ppc *PnpmPublishCommand) resolveVirtualRepos(publishRepos map[string]strin
...
+// tagPublishedProperties sets build and detected VCS properties on published artifacts
+// in a single SetProps call. Constructs artifact paths from package name/version and
+// registry config, avoiding a SearchFiles API call.
+func (ppc *PnpmPublishCommand) tagPublishedProperties(published []publishedPackage, publishRepos map[string]string, fallbackRepos registryMap) {
 	if ppc.serverDetails == nil {
-		log.Debug("No server details configured. Skipping build property tagging.")
+		log.Debug("No server details configured. Skipping property tagging.")
 		return
 	}
-	servicesManager, err := artCoreUtils.CreateServiceManager(ppc.serverDetails, -1, 0, false)
```
reshmifrog, line 580: why did we remove this , where we creating the servicemanger?

reshmifrog (changes requested):

reshmifrog:
the testcases are failing please take a look at them

artifactory/commands/pnpm/publish.go
attiasas: did not remove, refactor code to a method (`setPropsOnPublishedItems`) this is the only place you actually need it

artifactory/commands/conan/artifacts.go
attiasas: see:
* https://github.com/jfrog/jfrog-cli/pull/3542

artifactory/commands/buildinfo/civcs_props.go
```diff
@@ -42,28 +42,28 @@ func extractArtifactPathsWithWarnings(buildInfo *buildinfo.BuildInfo) ([]string,
...
 			return artifact.OriginalDeploymentRepo + "/" + artifact.Name
 		}
 	}
 
-	// Fallback: Use Path directly - it might be a complete path or might fail
-	// If it fails, setPropsOnArtifacts will warn and move on
+	// Fallback: path-only from build-info — use wildcard repo for search-based SetProps
 	if artifact.Path != "" {
-		return artifact.Path
+		return "*/" + strings.TrimPrefix(artifact.Path, "/")
```
agrasth, line 62: When the Gradle extractor omits `OriginalDeploymentRepo`, the path `*/com/example/foo/1.0/foo.jar` is used for `SetItemProperties`. Artifactory will match this pattern across all repos — local deployment repos, virtual repos that shadow the same path, remote repo caches, etc. This silently stamps VCS metadata on artifacts the user never intended to tag.

In practice this is mostly harmless (virtual repos are usually read-only for property writes), but if an org has a virtual repo backed by multiple locals and the same artifact was released at the same path in a different local repo, both get tagged.

We can look for a way to restrict the search to local repos only in the AQL or SetProps call when using */? If not, at minimum log a warning when the wildcard fallback is used.

agrasth (dismissed):

artifactory/commands/buildinfo/civcs_props.go
attiasas: added a log and a call to get all local repositories

agrasth (approved):

reshmifrog (approved):

PR #486, opened 2026-06-07
Expend vsc detection for container

attiasas:
-----

Depends on:
* https://github.com/jfrog/jfrog-cli-artifactory/pull/484

## Changes
  - Docker push/publish (OCI container)
  - Helm push, Hugging Face upload
  - Terraform publish


artifactory/commands/container/containermanagerbase.go
```diff
@@ -21,6 +23,7 @@ type ContainerCommandBase struct {
 	buildConfiguration *build.BuildConfiguration
 	serverDetails      *config.ServerDetails
 	validateSha        bool
+	workingDirectory   string
 }
 
 func (ccb *ContainerCommandBase) ImageTag() string {
```
agrasth, line 29: For every `docker push / docker pull / build-docker-create` invocation where the user specifies the repo (the normal case), `init()` exits early and `ccb.workingDirectory` stays `""`. This empty string is then passed to `NewLocalAgentBuildInfoBuilder / NewRemoteAgentBuildInfoBuilde`. Git detection falls back to process CWD, which could be /tmp, a workspace root, or a CI runner directory — not necessarily the project containing `.git`.

`BuildDockerCreateCommand` is worst-case: it has no other code path that populates `workingDirectory` (unlike `BuildCommand`, which calls `ResolveWorkingDirectoryFromDockerArgs`). So build-docker-create always passes "".

We can try to move the `os.Getwd()` block above the if `ccb.repo != ""` early-return, so workingDirectory is always initialized

agrasth (changes requested):

artifactory/commands/container/containermanagerbase.go
attiasas: done

artifactory/utils/docker_context.go
```diff
@@ -0,0 +1,38 @@
...
+
+// ExtractDockerBuildContextFromArgs returns the docker build context directory (last non-flag arg).
+// Falls back to "." when not found.
+func ExtractDockerBuildContextFromArgs(args []string) (string, error) {
+	var last string
+	for _, arg := range args {
+		if arg == "" || arg[0] == '-' {
+			continue
+		}
+		last = arg
```
agrasth: Major 🟠 — Functionality: This "pick the last non-flag token" heuristic breaks whenever any flag (or its value) follows the build context on the command line, e.g. `docker build ./service-a --build-arg FOO=bar -t org/service-a:1.0` — a valid and common invocation since Docker's cobra-based parser allows interspersed flags. Here, `org/service-a:1.0` (the `-t` value) wins instead of `./service-a`. Confirmed via a local repro: `ResolveWorkingDirectoryFromDockerArgs` returns the CWD instead of the intended subdirectory in this scenario, because the wrong candidate fails the `os.Stat` dir check and it falls back to `os.Getwd()`.

**Impact:** For monorepo-style builds (`docker build <subdir> [flags...]`), the newly added VCS property detection silently attaches the wrong (CWD-based) git metadata instead of the actual build-context repo/subdirectory metadata — undermining the exact feature this PR adds.

**Suggested fix:** Don't rely on token position. Docker build takes exactly one positional PATH|URL|- argument; instead, maintain a small set of known docker-build flags that consume a following value (`-t/--tag`, `-f/--file`, `--build-arg`, `--label`, `--target`, `--platform`, `--output`, `--secret`, `--ssh`, `--cache-from`, `--cache-to`, `--iidfile`, `--metadata-file`, `--network`, `--add-host`, etc.) and skip the flag together with its value, then take the single remaining non-flag token as context. Add a test case where the context precedes trailing flags (e.g. `["build", "./service-a", "-t", "img:tag"]`) to lock in the fix.

artifactory/commands/ocicontainer/docker_build.go
```diff
@@ -116,6 +118,12 @@ func (dbib *DockerBuildInfoBuilder) applyBuildProps(items []utils.ResultItem, pu
 	if err != nil {
 		return
 	}
+	searchDir, wdErr := artutils.ResolveWorkingDirectoryFromDockerArgs(dbib.cmdArgs)
+	if wdErr != nil {
+		log.Warn("Failed to resolve docker build context for VCS props:", wdErr.Error())
+	} else {
+		props = civcs.MergeWithUserProps(props, searchDir)
+	}
 	if pushedRepo == "" {
```
agrasth, line 121: Minor 🔵 — Performance/Maintainability: The new VCS resolution above (lines 121-126), including a `git` lookup via `civcs.MergeWithUserProps`, runs before this early-return check. When `pushedRepo == ""` (or the `len(filteredLayers) == 0` check just below also returns), the computed `props` are discarded entirely, so the git/filesystem work was wasted. Consider moving the VCS-resolution block after these guard clauses so it only runs when `props` will actually be used.

artifactory/commands/helm/buildinfo.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // CollectHelmBuildInfoWithFlexPack collects Helm build info using FlexPack
-func CollectHelmBuildInfoWithFlexPack(workingDir, buildName, buildNumber, project, commandName string, helmArgs []string, serverDetails *config.ServerDetails) error {
+func CollectHelmBuildInfoWithFlexPack(workingDir, buildName, buildNumber, project, commandName string, helmArgs []string, serverDetails *config.ServerDetails, workingDirectory string) error {
```
agrasth: Nit 🟢 — Maintainability: `workingDir` and the newly added `workingDirectory` are both plain strings on this signature and, per the only call site (`helmcommand.go:207`), are always passed the same value (`hc.workingDirectory`). Having two similarly-named parameters that mean the same thing here is confusing for future readers — consider reusing `workingDir` for the VCS lookup instead of adding a second parameter.

agrasth (changes requested):
## Review Summary

Reviewed the diff extending VCS-detection to Docker/OCI push-build, Helm push, Hugging Face upload, and Terraform publish (jfrog-cli-artifactory#486). Build and package-level tests pass (`go build ./...`, plus targeted `go test` on all touched packages).

### Major 🟠
1. **Docker build-context detection breaks for a very common CLI arg order.** `ExtractDockerBuildContextFromArgs` (`artifactory/utils/docker_context.go:12-19`) picks the *last* token that doesn't start with `-` and assumes it's the build context. Docker's own CLI parser (cobra) allows flags to be interspersed anywhere, so `docker build ./service-a --build-arg FOO=bar -t org/service-a:1.0` is a completely valid, common invocation. In that case the function returns the `-t` flag's *value* (`org/service-a:1.0`) instead of `./service-a`. I reproduced this locally: calling `ResolveWorkingDirectoryFromDockerArgs([]string{"build", "./service-a", "--build-arg", "FOO=bar", "-t", "org/service-a:1.0"})` returns the current working directory instead of `./service-a`, because the mis-picked candidate fails the `os.Stat` directory check and silently falls back to `os.Getwd()`. This means the new VCS-detection feature this PR adds will silently attribute the wrong (or merely CWD-based) VCS metadata whenever a user builds a subdirectory context and puts any flag after the context path — which is one of the two most common Docker invocation styles. Neither of the two new tests in `docker_context_test.go` covers this ordering (both put the context as the literal last token), so CI does not catch it.

### Minor 🔵
2. **Wasted work / misleading warning on the early-return path.** In `applyBuildProps` (`artifactory/commands/ocicontainer/docker_build.go:121-130`), the new VCS resolution (`ResolveWorkingDirectoryFromDockerArgs` + `civcs.MergeWithUserProps`, which shells out to `git`) runs unconditionally before the pre-existing early-return checks for `pushedRepo == ""` and "no eligible layers". When those checks trigger, the computed `props` (including the freshly merged VCS properties) are discarded entirely, so the git lookup was pure overhead. Moving the VCS-resolution block after those guards avoids the wasted `git`/filesystem work on the common short-circuit paths.

### Nit 🟢
3. **Confusing duplicate parameter naming.** `CollectHelmBuildInfoWithFlexPack` (`artifactory/commands/helm/buildinfo.go:14`) now takes both `workingDir` and `workingDirectory` as separate string parameters, and its only caller (`helmcommand.go:207`) passes `hc.workingDirectory` for both. They always carry the same value here, so the second parameter is redundant/confusing — worth renaming or dropping in favor of reusing the existing `workingDir` parameter for the new VCS lookup.

## Verdict
Request changes on finding #1 — the core VCS-detection heuristic this PR introduces for Docker builds silently produces wrong results for a very common Docker CLI argument order.

artifactory/utils/docker_context.go
attiasas: implemented suggested fix

artifactory/commands/helm/buildinfo.go
attiasas: done

artifactory/commands/ocicontainer/docker_build.go
attiasas: Moved ResolveWorkingDirectoryFromDockerArgs + civcs.MergeWithUserProps to after the pushedRepo == "" and len(filteredLayers) == 0 guards, so git lookup only runs when props will be applied.

agrasth (approved):

fluxxBot (approved):

PR #489, opened 2026-06-16
fix poetry publish command not respecting the repo flag

reshmifrog:
-----
Problem: `jf poetry publish -r <deploy-repo>` failed with Repository  <name> is not defined. Two bugs in poetry.go: (1) pc.repository always held the resolver repo from poetry.yaml, the user's -r flag was ignored, so
  credentials got registered on the wrong repo; (2) the deploy repo was appended as a single argv element "-r "+repo, which Poetry parsed as a value with a leading space.
Solution: Read the deploy repo from the user's -r/--repository flag and apply it before credentials are configured, so credentials are registered against the actual deploy repo. Then pass the repository to Poetry as two separate arguments
  instead of one combined string, and skip adding it altogether when the user already supplied the flag, avoiding both the leading-space parsing error and a duplicate flag.

### DIFF
--- a/artifactory/commands/python/poetry.go
+++ b/artifactory/commands/python/poetry.go
@@ -58,9 +58,8 @@ func (pc *PoetryCommand) Run() (err error) {
 	}()
 	// For publish, the deploy repository comes from the user's -r/--repository flag rather than the
 	// resolver repo configured in .jfrog/projects/poetry.yaml. Honor it before credential setup so
-	// Poetry credentials are registered under the deploy repo. If the user didn't pass one, append the
-	// resolver repo so Poetry still has a publish target. Either way pc.args then carries the repo, so
-	// publish() can pass it through as-is.
+	// Poetry credentials are registered under the deploy repo. The flag is already in pc.args, so
+	// publish() passes it through to Poetry as-is.
 	if pc.commandName == "publish" {
 		var repo string
 		if _, _, repo, err = coreutils.FindFlagFirstMatch([]string{"--repository", "-r"}, pc.args); err != nil {
@@ -68,8 +67,6 @@ func (pc *PoetryCommand) Run() (err error) {
 		}
 		if repo != "" {
 			pc.repository = repo
-		} else {
-			pc.args = append(pc.args, "-r", pc.repository)
 		}
 	}
 	err = pc.SetPypiRepoUrlWithCredentials()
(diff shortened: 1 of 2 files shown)


agrasth (dismissed):

artifactory/commands/python/poetry.go
```diff
@@ -40,6 +40,19 @@ func NewPoetryCommand() *PoetryCommand {
 	}
 }
 
+// extractRepoFromArgs returns the value of the -r / --repository flag from args
+func extractRepoFromArgs(args []string) string {
```
fluxxBot: there are existing helper functions to extract in jfrog-cli-core, please use them

artifactory/commands/python/poetry.go
```diff
@@ -96,7 +117,13 @@ func (pc *PoetryCommand) install(buildConfiguration *buildUtils.BuildConfigurati
...
 
 func (pc *PoetryCommand) publish(buildConfiguration *buildUtils.BuildConfiguration, pythonBuildInfo *build.Build) error {
-	publishCmdArgs := append(slices.Clone(pc.args), "-r "+pc.repository)
+	// Pass the repository as two separate argv elements ("-r", repo). Appending "-r "+pc.repository as
+	// a single element makes Poetry read the value with a leading space ("Repository  <name> is not
+	// defined"). Only append when the user did not already supply -r/--repository themselves.
+	publishCmdArgs := slices.Clone(pc.args)
+	if extractRepoFromArgs(publishCmdArgs) == "" {
+		publishCmdArgs = append(publishCmdArgs, "-r", pc.repository)
+	}
```
fluxxBot: no need to extract and append again, simply pass it

fluxxBot (changes requested):

artifactory/commands/python/poetry.go
```diff
@@ -56,6 +56,22 @@ func (pc *PoetryCommand) Run() (err error) {
...
+	// publish() can pass it through as-is.
+	if pc.commandName == "publish" {
+		var repo string
+		if _, _, repo, err = coreutils.FindFlagFirstMatch([]string{"--repository", "-r"}, pc.args); err != nil {
+			return err
+		}
+		if repo != "" {
+			pc.repository = repo
+		} else {
+			pc.args = append(pc.args, "-r", pc.repository)
```
fluxxBot: I don't think you need to append again, please check

fluxxBot (changes requested):

fluxxBot (approved):

agrasth (approved):

PR #491, opened 2026-06-17
Zero Touch Remediation (ZTR) for npm — opt-in

attiasas:
-----

Depends on:
* https://github.com/jfrog/jfrog-client-go/pull/1355
* https://github.com/jfrog/jfrog-cli-core/pull/1575

### What

Adds Zero Touch Remediation (ZTR) to the npm `install` / `ci` / `publish` flows. Before the build tool runs, the CLI discovers the project's lockfiles, sends each one to Xray, and writes back the remediated lockfile. If the subsequent npm command fails, the original lockfiles are restored. After a remediated `install`, the CLI runs `npm ci` so the remediated lockfile's integrity values are honored.

### Enabling it

The feature is **disabled by default**. To turn it on:

```
JFROG_CLI_ZTR_COMPONENTS_ENABLED=true
```

Requires Xray 3.154.0 or later — older versions log a warning and skip remediation.

### Notes

* Feature code lives in `artifactory/zerotouchremediation` (with `/npm` for npm-specific project discovery, argument parsing, and lockfile bootstrapping).
* Best-effort by design: a missing resolver repo, an unreachable Xray, or an unsupported Xray version logs and continues rather than failing the build.

## Summary by CodeRabbit

* **New Features**
  * Added zero-touch remediation for npm installs and publishes, including lockfile preparation and restoration.
  * Improved support for npm workspaces, prefixes, publish paths, and project-root discovery.
  * Added AI editor extension token retrieval.
* **Bug Fixes**
  * Remediation failures now safely skip remediation instead of interrupting npm commands.
  * Added safeguards against unsafe or symlinked lockfile paths.
  * Ensured failed lockfile updates can be rolled back cleanly.
* **Tests**
  * Expanded coverage for npm argument parsing, workspace handling, lockfiles, remediation, and rollback scenarios.


artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,112 @@
...
+	return err
+}
+
+func isSinglePackageInstall(npmArgs []string) bool {
+	for _, arg := range npmArgs {
+		if !strings.HasPrefix(arg, "-") {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
attiasas: Verified. `-w` is `--workspace` (value required) but was grouped with `--workspaces` (boolean), so `npm install -w app` dropped `app` from bootstrap args. Independently, `isSinglePackageInstall` treated any non-dash token as a package, so `--workspace app` skipped ZTR.

`parseNpmCLIArgs` now consumes `-w`/`-w=` like `--workspace` and records positional package operands. `isSinglePackageInstall` uses that result.

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+	}
+	if len(toWrite) == 0 {
+		return noopRestore, false, nil
+	}
+	log.Debug("Applying", len(toWrite), "healed lockfile(s)...")
+	restore, err = ApplyLockfiles(projectRoot, toWrite, bootstrapped)
+	if err != nil {
+		return noopRestore, false, err
+	}
```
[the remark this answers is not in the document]
attiasas: Verified. On a later `MkdirAll`/`WriteFile` failure, `ApplyLockfiles` returned without restoring already-written files, and `RunIfEnabled` handed the caller `noopRestore`.

`ApplyLockfiles` now restores only successfully written files before returning the original error (joined with restore errors if those fail). Covered by `TestApplyLockfiles_RollsBackOnLaterWriteFailure`.

Jordanh1996 (changes requested):
The repo organizes by technology — commands/<tech>/, 29 of them — and Curation is the closest precedent for what ZTR is doing: a cross-cutting, Xray-driven concern that needs per-package-manager handling. It lives inside each tech package (commands/npm, commands/python, commands/golang, commands/setup), not in a curation/ tree with per-tech subdirs. zerotouchremediation/npm is the only per-technology directory outside commands/, and zerotouchremediation is the only feature-named dir at the artifactory/ top level.

Two concrete costs:

- Name collision. commands/npm (package npm) has to import this as cnpm, so we read cnpm.NewBuildToolWithArgs(...) from inside package npm. Aliasing internal packages is rare here.
- npm knowledge in two places. commands/npm already parses npm args (filterFlags, ExtractNpmOptionsFromArgs) and owns npmrc and the executable path. This adds a second npm arg parser next to it — which is where the --registry-swallows-the-value bug comes from.

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	var projectKey string
+	if ca.buildConfiguration != nil {
+		projectKey = ca.buildConfiguration.GetProject()
+	}
+	xrayManager, xrayErr := xray.CreateXrayServiceManager(ca.serverDetails, xray.WithScopedProjectKey(projectKey))
+	if xrayErr != nil {
+		return zerotouchremediation.SkipRemediation("Zero Touch Remediation skipped: could not create Xray service manager: ", xrayErr)
+	}
+	return zerotouchremediation.RunIfEnabled(ctx, xrayManager, resolverRepo, tool, command, workingDir, ca.npmBootstrapRunner(), bootstrapArgs...)
```
Jordanh1996, line 46: **The bootstrap shells out to npm before the Artifactory .npmrc exists, so it resolves against the developer's own registry.**

This runs at NpmCommand.Run:354, but CreateTempNpmrc is not called until installHandler.Install() at :366 → artifactoryinstall.go:23. So when EnsureLockfiles shells out to npm install --package-lock-only, npm is still using the user's own .npmrc.

I ran this against a local stack: the log prints "generating package-lock.json" before "Creating temporary .npmrc file", and the lockfile it produced resolved lodash to https://registry.npmjs.org/... rather than through Artifactory. That means the bootstrap bypasses the proxy and the Curation gate — and the lockfile we then hand Xray describes packages fetched from outside the platform.

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,200 @@
...
+				backup.content = data
+			} else if !os.IsNotExist(readErr) {
+				return fail(readErr)
+			}
+		}
+
+		if err = os.MkdirAll(filepath.Dir(fullPath), 0755); err != nil {
+			return fail(err)
+		}
+		if err = os.WriteFile(fullPath, lf.Content, 0644); err != nil {
```
Jordanh1996: **A failed write destroys the user's lockfile with no way back.**

os.WriteFile opens with O_TRUNC, so the file is emptied the moment it opens — but backups = append(backups, backup) is three lines later, only on success. If the write fails after the open, fail() rewinds every lockfile except the one it just truncated.

Reproduced on a full volume: a valid 403-byte package-lock.json became 0 bytes and the original was unrecoverable. A CI runner out of disk is enough to trigger it.

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// RunIfEnabled ensures lockfiles exist, discovers them, calls Xray once per file, writes remediated lockfiles when changes returned.
+// Returns a restore function to revert lockfile writes if the subsequent build-tool command fails, and remediated=true when at least one lockfile was updated.
+func RunIfEnabled(ctx context.Context, client ComponentResolutionClient, repo string, tool BuildTool, command, workingDir string, runner CommandRunner, bootstrapArgs ...string) (restore func() error, remediated bool, err error) {
+	if !IsComponentResolutionEnabled() || !IsRelevantCommand(tool, command) {
+		log.Debug("Zero Touch Remediation disabled or not relevant command: ", command)
+		return noopRestore, false, nil
+	}
+	version, err := client.GetVersion()
```
Jordanh1996, line 75: **"Best-effort" holds for the pre-flight only; six exits inside RunIfEnabled fail the build.**

The PR description says an unreachable Xray "logs and continues rather than failing the build", but GetVersion() is the first call here and its error returns straight out to NpmCommand.Run, which returns before npm is invoked. Same for ProjectRoot, EnsureLockfiles, DiscoverLockfiles, the remediate call and ApplyLockfiles — six hard exits against one skipRemediationWarn.

Verified: with --xray-url pointed at a closed port, jf npm install exits 1 on connection refused and never runs npm. Since the feature is opt-in and advertised as best-effort, I think every error inside this function should become skipRemediationWarn — a remediation problem should never be why someone's install fails.

artifactory/commands/npm/npmcommand.go
```diff
@@ -348,6 +350,16 @@ func (nc *NpmCommand) Run() (err error) {
 	if err = nc.PreparePrerequisites(nc.repo); err != nil {
 		return
 	}
+	var restoreResolution func() error
+	restoreResolution, nc.remediatedLockfile, err = nc.runZeroTouchRemediation(context.Background(), nc.cmdName, nc.workingDirectory, nc.npmArgs)
+	if err != nil {
+		return err
+	}
+	defer func() {
```
Jordanh1996, line 367: **Defers run LIFO, so a failed npmrc restore reverts a successful remediation.**

This defer is registered before the RestoreNpmrc defer at :363, so it runs after it — by which point RestoreNpmrc may have joined its own error into err. A successful install whose npmrc cleanup fails therefore trips if err != nil and rolls the lockfile back.

Reproduced: npm ci succeeded ("added 1 package"), the restore failed, and the remediated lockfile was reverted while node_modules stayed installed from it — the two now disagree with nothing flagging it.

Capturing the install error in its own variable, and testing that rather than the shared named return, would keep the two concerns separate.

go.mod
```diff
@@ -199,7 +199,11 @@ require (
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
 
-// replace github.com/jfrog/jfrog-cli-core/v2 => github.com/jfrog/jfrog-cli-core/v2 v2.60.1-0.20260729061834-1c527b8abaa6
+// attiasas:xray_component_resolution
+replace github.com/jfrog/jfrog-client-go => github.com/attiasas/jfrog-client-go v0.0.0-20260827103045-03afa565ee4e
+
+// attiasas:expend_xray_manager
```
Jordanh1996: don't forget

artifactory/commands/npm/zerotouchremediation.go
attiasas: Verified: `EnsureLockfiles` ran at `NpmCommand.Run` before `CreateTempNpmrc` (inside `installHandler.Install()`), so `npm install --package-lock-only` used the user's registry.

Artifactory installs now write the temp `.npmrc` before ZTR, and `RestoreNpmrc` is deferred from the start of `Run` so that file is always restored. Native installs still use the user's npmrc (no temp file).

artifactory/zerotouchremediation/service.go
attiasas: Verified: `os.WriteFile` truncates on open, and the backup was appended only after a successful write, so `fail()` could not restore the file that had just been emptied.

`ApplyLockfiles` now records the backup before the write. `TestApplyLockfiles_RestoresWhenWriteTruncatesThenFails` truncates then returns an error and asserts the original bytes are restored.

artifactory/zerotouchremediation/service.go
attiasas: Verified: `GetVersion`, `ProjectRoot`, `EnsureLockfiles`, `DiscoverLockfiles`, the remediate call, and `ApplyLockfiles` returned errors that `NpmCommand.Run` treated as fatal, so an unreachable Xray aborted npm.

Those paths now go through `skipRemediationWarn` (log + continue). Covered by `TestRunIfEnabled_SkipsGetVersionError`, `TestRunIfEnabled_SkipsEnsureLockfilesError`, and `TestRunIfEnabled_SkipsRemediateError`.

artifactory/commands/npm/npmcommand.go
attiasas: Verified: LIFO meant `RestoreNpmrc` ran first and joined its error into `err`, so a successful install still triggered lockfile restore.

Restore now keys off `installErr` from `Install()`, not the aggregated named return. `RestoreNpmrc` errors still join into the returned error.

go.mod
attiasas: These replace directives stay until the client-go and cli-core PRs (jfrog-client-go#1355 / jfrog-cli-core#1575) land. They come out in the follow-up that bumps those modules to released versions.

attiasas:
On package layout: Curation living in commands/<tech> is the right precedent for npm-specific discovery and arg parsing, and the cnpm alias was a real cost.

artifactory/zerotouchremediation stays as the generic Xray apply/restore loop (Maven ZTR in #492 uses the same contract). npm-only code moved into artifactory/commands/npm (ztr_*.go), so that package no longer imports a second npm package.

Also skipped ZTR on npm publish when a tarball is already provided — remediating lockfiles cannot change the artifact being uploaded.

artifactory/zerotouchremediation/service.go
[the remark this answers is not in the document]
attiasas: Lockfile.Path comes from discovery, not Xray, and npm only returns constant basenames. ApplyLockfiles is still a generic writer, so it now rejects paths that leave projectRoot after Clean/Rel, and Lstat-rejects a symlink at the lockfile itself.

It does not walk ancestor components for symlinks: on macOS /var is a symlink to /private/var, so that check would refuse every write under a typical TempDir.

artifactory/commands/npm/ztr_args_parse.go
```diff
@@ -0,0 +1,95 @@
...
+			if i+1 < len(args) {
+				i++
+				out.bootstrapArgs = append(out.bootstrapArgs, arg, args[i])
+			}
+		case strings.HasPrefix(arg, "--workspace="):
+			out.bootstrapArgs = append(out.bootstrapArgs, arg)
+		case strings.HasPrefix(arg, "-w=") && len(arg) > 3:
+			out.bootstrapArgs = append(out.bootstrapArgs, arg)
+		case strings.HasPrefix(arg, "-"):
+			continue
```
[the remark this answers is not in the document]
attiasas: Verified. `--registry URL` and `--tag TAG` (space-separated) were classified as package operands, so `HasPackageOperands` skipped ZTR on package-free installs. The `=` forms were already ignored as flags.

`parseNpmCLIArgs` now consumes the following token for `--registry` and `--tag`. Covered by `TestHasPackageOperands` and `TestIsSinglePackageInstall`.

artifactory/commands/npm/ztr_args_parse.go
```diff
@@ -0,0 +1,99 @@
...
+		case arg == "--workspace" || arg == "-w":
+			if i+1 < len(args) {
+				i++
+				out.bootstrapArgs = append(out.bootstrapArgs, arg, args[i])
+			}
+		case strings.HasPrefix(arg, "--workspace="):
+			out.bootstrapArgs = append(out.bootstrapArgs, arg)
+		case strings.HasPrefix(arg, "-w=") && len(arg) > 3:
+			out.bootstrapArgs = append(out.bootstrapArgs, arg)
+		case arg == "--registry" || arg == "--tag":
```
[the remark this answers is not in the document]
attiasas: `--omit` takes a value (`dev`/`optional`/`peer`). The parser treated the space-separated form as a package operand, so `npm install --omit dev` skipped ZTR.

`parseNpmCLIArgs` now consumes the token after `--omit` the same way it does for `--registry`/`--tag`. `--omit=dev` was already ignored as a flag. Covered in `TestHasPackageOperands` and `TestIsSinglePackageInstall`.

artifactory/commands/npm/ztr_buildtool.go
```diff
@@ -0,0 +1,91 @@
...
+func (t BuildTool) DiscoverLockfiles(workingDir string) ([]zerotouchremediation.Lockfile, error) {
+	root, err := discoverProjectRootWithOptions(workingDir, t.opts)
+	if err != nil {
+		return nil, err
+	}
+	name, err := lockfileNameInDir(root)
+	if err != nil {
+		return nil, err
+	}
+	data, err := os.ReadFile(filepath.Join(root, name))
```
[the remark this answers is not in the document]
attiasas: `lockfileNameInDir` used `os.Stat`, so a symlinked `package-lock.json` was accepted and `DiscoverLockfiles` read the target into the Xray request. `ApplyLockfiles` already refused the write, but the content had already been sent.

Discovery now uses `os.Lstat` and skips symlink lockfiles. `TestNpmBuildTool_DiscoverLockfiles_RejectsSymlink` and `TestRunIfEnabled_SkipsSymlinkedLockfile` assert discovery errors and zero remediation requests.

artifactory/commands/cargo/command_test.go
fluxxBot: no changes are made here, please remove this diff

artifactory/commands/npm/common.go
```diff
@@ -14,6 +14,8 @@ type CommonArgs struct {
 	npmArgs            []string
 	serverDetails      *config.ServerDetails
 	useNative          bool
+	configFilePath     string
+	executablePath     string
 }
```
fluxxBot: I see this being part of NpmCommand and NpmModule struct, check if you can resuse them instead of creating duplicates here

Jordanh1996 (approved):

attiasas:
Split into smaller functional PRs as requested:

1. #543 — generic Zero Touch Remediation service
2. #544 — opt-in ZTR for npm install/ci (depends on #543)
3. #545 — ZTR for npm publish (depends on #543 and #544)

The unrelated cargo/setup/token churn was left out. Closing this combined PR in favor of the split series.

PR #495, opened 2026-06-18
RTECO-1074 - Remove redundant functions in agent namespace(1/2)

udaykb2:
-----
### Summary
**Removes wrapper functions that were duplicating logic between agent/plugins/common and agent/skills/common. Specifically:**

- Moved ResolveAgentTargets, CopyExtractedToTargets, and related helpers up to agent/common (the single source of truth)
- Deleted the per-namespace copies in plugins/common and skills/common
- Updated all callers in install.go, update.go, list.go to call agentcommon.* directly
- Moved tests to agent/common

### DIFF
--- a/agent/common/agents.go
+++ b/agent/common/agents.go
@@ -140,6 +140,28 @@ func ResolveAgentInstallDir(spec AgentSpec, projectDir string, global bool) (str
 	return filepath.Abs(filepath.Join(projectDir, spec.Config.ProjectDir))
 }
 
+// ParseHarnessList parses comma-separated harness names (trim, lowercase, reject empty/duplicates).
+func ParseHarnessList(raw string) ([]string, error) {
+	if strings.TrimSpace(raw) == "" {
+		return nil, fmt.Errorf("--harness is required (comma-separated list of harness names)")
+	}
+
+	seen := make(map[string]struct{})
+	var result []string
+	for _, part := range strings.Split(raw, ",") {
+		name := strings.ToLower(strings.TrimSpace(part))
+		if name == "" {
+			return nil, fmt.Errorf("--harness contains an empty name in %q", raw)
+		}
+		if _, dup := seen[name]; dup {
+			return nil, fmt.Errorf("--harness lists %q more than once", name)
+		}
+		seen[name] = struct{}{}
+		result = append(result, name)
+	}
+	return result, nil
+}
+
 // SupportedAgentsList is comma-separated names from registry or built-ins.
 func SupportedAgentsList(builtIns map[string]AgentConfig, configSectionKey string) string {
 	registry, err := LoadAgentRegistry(builtIns, configSectionKey)
--- a/agent/common/agents_test.go
+++ b/agent/common/agents_test.go
@@ -1,0 +1,27 @@
+package common
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseHarnessList(t *testing.T) {
+	got, err := ParseHarnessList("cursor, Claude")
+	require.NoError(t, err)
+	assert.Equal(t, []string{"cursor", "claude"}, got)
+}
+
+func TestParseHarnessList_EmptyAndDuplicates(t *testing.T) {
+	_, err := ParseHarnessList("")
+	require.Error(t, err)
+
+	_, err = ParseHarnessList("cursor,,claude")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "empty name")
+
+	_, err = ParseHarnessList("cursor,cursor")
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "more than once")
+}
--- a/agent/common/agents_test.go
+++ b/agent/common/agents_test.go
@@ -7,12 +7,17 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestParseHarnessList verifies that a comma-separated harness list is split,
+// trimmed, and lowercased correctly.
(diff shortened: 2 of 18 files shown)


agent/common/agents.go
```diff
@@ -140,6 +140,28 @@ func ResolveAgentInstallDir(spec AgentSpec, projectDir string, global bool) (str
...
+
+	seen := make(map[string]struct{})
+	var result []string
+	for _, part := range strings.Split(raw, ",") {
+		name := strings.ToLower(strings.TrimSpace(part))
+		if name == "" {
+			return nil, fmt.Errorf("--harness contains an empty name in %q", raw)
+		}
+		if _, dup := seen[name]; dup {
+			return nil, fmt.Errorf("--harness lists %q more than once", name)
```
reshmifrog, line 157: Nit pick : why %q why not %s , it is a string right or will it be having quotes also?

agent/common/agents_test.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// TestParseHarnessList_EmptyAndDuplicates verifies that an empty string, a
+// segment that resolves to an empty name after trimming, and a repeated harness
+// name are all rejected with descriptive errors.
+func TestParseHarnessList_EmptyAndDuplicates(t *testing.T) {
+	_, err := ParseHarnessList("")
+	require.Error(t, err)
+
+	_, err = ParseHarnessList("cursor,,claude")
```
reshmifrog, line 25: it is not empty then why are we throwing empty name error?

agent/skills/commands/install/install.go
```diff
@@ -247,13 +247,13 @@ func (ic *InstallCommand) resolveAgentTargetDirectories() ([]common.AgentTarget,
 		return ic.explicitTargets, nil
 	}
 	if ic.installPath != "" {
-		return common.ResolveAgentTargets(ic.slug, ic.installPath, nil, "", false)
+		return agentcommon.ResolveAgentTargets(ic.slug, ic.installPath, nil, "", false)
```
reshmifrog, line 252: why are we sending empty string here?

agent/skills/commands/install/install.go
```diff
@@ -247,13 +247,13 @@ func (ic *InstallCommand) resolveAgentTargetDirectories() ([]common.AgentTarget,
...
 	}
 	if ic.scope == agentcommon.InstallScopeProject && ic.projectDir == "" {
 		return nil, fmt.Errorf("project directory is required for project-scoped install")
 	}
 	isGlobal := ic.scope == agentcommon.InstallScopeGlobal
-	return common.ResolveAgentTargets(ic.slug, "", ic.agents, ic.projectDir, isGlobal)
+	return agentcommon.ResolveAgentTargets(ic.slug, "", ic.agents, ic.projectDir, isGlobal)
```
reshmifrog, line 260: here as well again empty string , it is not understandable what exactly it is, can you please declare a variable and assign empty string value to it so that one can understand what it is we are sending empty as.

reshmifrog (changes requested):

agent/skills/commands/install/install.go
udaykb2: There are 2 ways where we use install command
1. Using --harness cursor or --harness claude
2. Other, we provide direct raw path using --path

so if user provides --path, there is no agent name & there is no directory path, as we are passing install path directly through --path (line 250)

Where as if user uses --harness, we use agent name + directory path (see line 256), that time we don't have --path so 2nd param is empty.

agent/skills/commands/install/install.go
udaykb2: There are 2 ways where we use install command

Using --harness cursor or --harness claude
Other, we provide direct raw path using --path
so if user provides --path, there is no agent name & there is no directory path, as we are passing install path directly through --path (line 250)

Where as if user uses --harness, we use agent name + directory path (see line 256), that time we don't have --path so 2nd param is empty.

agent/common/agents_test.go
udaykb2: When we split "cursor,,claude" by comma, it becomes ["cursor", "", "claude"], when we are parsing name by name here second name is empty

The output will say '--harness contains an empty name in "cursor,,claude"'

agent/common/agents.go
udaykb2: I have added it for more clarity

for example:
If %q is set : --harness contains an empty name in "claude,,cursor"

If %s is set : --harness contains an empty name in claude,,cursor

As these are comma separated values, grouping them under quotes makes more sense, isn't it?

Please share your opinion, I will update accordingly

agent/skills/commands/install/install.go
reshmifrog: can you please add a comments above it so that it's understandable for someone looking at the code and please add an example also to make it easier in understanding

agent/skills/commands/install/install.go
reshmifrog: I think it's not clear please add proper comments with an example it will be easy to understand the code for someone looking at it

agent/skills/commands/install/install.go
udaykb2: Done, added comments with example

Also update comments of ResolveAgentTargets() to make it clear to understand

agent/skills/commands/install/install.go
udaykb2: Done, added comments with example

Also update comments of ResolveAgentTargets() to make it clear to understand

reshmifrog (dismissed):

fluxxBot (dismissed):

reshmifrog (approved):

fluxxBot (approved):

PR #496, opened 2026-06-19
RTECO-945: Add Alpine APK command support (apk add/upload/config)

naveenku-jfrog:
## Summary

Implements the JFrog CLI Artifactory layer for Alpine APK integration: `jf apk add`, `jf apk upload`, and `jf apk config`. This layer sits between the JFrog CLI frontend (`jfrog-cli`) and the build-info library (`build-info-go`), handling credential injection, build-info orchestration, AQL checksum enrichment, and repository configuration.

Depends on: [build-info-go#393](https://github.com/jfrog/build-info-go/pull/393)

---

### New: `artifactory/commands/alpine/apkcommand.go`
The main `jf apk <subcommand>` wrapper:
- **`ApkCommand`** — wraps the native `apk` binary with credential injection and Build Info collection
- **Pre/post snapshot diffing** — takes a package snapshot before invoking native `apk`, diffs after to identify newly installed packages
- **`buildEnvWithHTTPAuth()`** — injects `HTTP_AUTH=basic:<host>:<user>:<pass>` into the subprocess environment so `apk-tools ≥ 2.12` can authenticate against Artifactory virtual repositories without touching `/etc/apk/repositories`
- **`collectBuildInfo()`** — two-phase collection: local checksums first (from APK DB `C:` field), then AQL enrichment for SHA-256/MD5 of packages that went through Artifactory
- **`enrichChecksumsFromAQL()`** — batched AQL query against the specified virtual repo to resolve SHA-256, SHA-1, and MD5 for any deps with missing checksums
- **`stripJFFlags()` / `extractPackageNames()`** — strips JFrog-specific flags (`--repo`, `--build-name`, etc.) before forwarding args to native `apk`
- **`filterSecretEnvVars()`** — respects `JFROG_CLI_ENV_EXCLUDE` to prevent secrets leaking into the subprocess environment or build-info records

### New: `artifactory/commands/alpine/apkupload.go`
Publishes a local `.apk` file to an Artifactory Alpine repository:
- Constructs the correct Artifactory path (`<repo>/<branch>/<arch>/<name>-<version>.apk`)
- Stamps CI/VCS properties on the uploaded artifact when running in a CI environment (GitHub Actions, Jenkins, GitLab CI, etc.)
- Supports `--branch`, `--arch` overrides

### New: `artifactory/commands/alpine/apkconfig.go`
Configures the local Alpine client to use Artifactory as the package source:
- Downloads the Artifactory RSA public key and writes it to `/etc/apk/keys/`
- Rewrites `/etc/apk/repositories` to replace CDN URLs with the Artifactory virtual repository URL
- `--apply` flag controls whether changes are written to disk or previewed
- Restores original `/etc/apk/repositories` on write failure to avoid leaving the system in a broken state

### New: `artifactory/commands/alpine/credentials.go`
Credential resolution helpers:
- `resolveHTTPAuthCredentials()` — merges server config, explicit `--user`/`--password` flags, and token-based auth
- `buildHTTPAuth()` — formats the `HTTP_AUTH=basic:<host>:<user>:<pass>` string for `apk-tools`

---

## Design Notes

**No routing change without `jf apk config`**: `jf apk add --repo=X` alone does not change `/etc/apk/repositories`. Packages are fetched from wherever `apk` is currently configured (CDN by default). The `--repo` flag affects HTTP_AUTH injection and AQL enrichment only. This is intentional — routing all traffic through Artifactory requires an explicit `jf apk config` step, matching the security model of the `apt` and `yum` integrations.

**`--repo` is optional for build-info**: Consistent with `jf npm install` and `jf pip install`, omitting `--repo` does not block build-info collection. Dependencies are still recorded with checksums from the APK DB. AQL enrichment (for SHA-256/MD5) is simply skipped.

**Unknown `--server-id` is a hard error**: If `--server-id` is explicitly provided and the server is not in the config, the command fails immediately — consistent with every other JFrog CLI package manager command.

**Dependency ID format**: `name:version` (e.g. `curl:8.14.1-r2`), consistent with npm, pip, and Maven modules.

## Summary by CodeRabbit

* **New Features**
  * Added Alpine APK package management through Artifactory, including authenticated installs and package handling.
  * Added APK uploads with metadata, dependency information, and optional build details.
  * Added Alpine repository configuration for local, remote, and virtual repositories.
  * Added repository validation, credential resolution, signing-key setup, and secure configuration updates.

* **Tests**
  * Added comprehensive coverage for APK commands, uploads, dependencies, credentials, repository setup, and security-related scenarios.


artifactory/commands/setup/setup.go
```diff
@@ -612,3 +626,358 @@ func (sc *SetupCommand) configureHelm() error {
...
+//
+// /etc/apk/repositories is fully replaced with the single Artifactory repo line so apk
+// resolves packages exclusively through Artifactory; public mirrors (e.g. dl-cdn) are
+// dropped. Use a virtual repo proxying the upstream Alpine mirror to still reach public packages.
+//
+// The repository URL is written with the Artifactory credentials embedded in its
+// userinfo component (https://<user>:<token>@host/...). This lets native apk
+// authenticate directly from /etc/apk/repositories without jf rebuilding HTTP_AUTH
+// on every command. Because the file then contains a secret, it is written 0600.
+func (sc *SetupCommand) configureApk() error {
```
agrasth, line 936: The PR description states that a `--apply` flag controls whether `jf setup apk` writes changes to disk vs. only previews them, but I don't see any such flag wired into `SetupCommand`/`configureApk` — it unconditionally calls `apkWriteSigningKey` and `apkUpdateRepositories` (which overwrites `/etc/apk/repositories` entirely). Is the `--apply`/preview gate meant to land in a follow-up, or should it be added here before this merges?

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,764 @@
...
+	}
+
+	// Build one AQL query that matches any of the missing .apk file names.
+	// Dep.Id format: "<name>-<version>.<arch>" → filename: "<name>-<version>.<arch>.apk"
+	names := make([]string, 0, len(missing))
+	for _, i := range missing {
+		id := deps[i].Id
+		names = append(names, fmt.Sprintf(`{"name":"%s.apk"}`, id))
+	}
+	aqlQuery := fmt.Sprintf(
```
agrasth, line 590: `aqlQuery` is built by splicing `deps[i].Id` (line 574, ultimately derived from package names/versions parsed off the apk DB or an index) and `apkCmd.repoKey` directly into the JSON query string via `fmt.Sprintf("%s")`, with no escaping. A package or repo name containing a `"` would break out of the query structure. Worth building this with the AQL/JSON builder helpers (or at least escaping quotes) instead of raw string interpolation.

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,764 @@
...
+}
+
+// filterSecretEnvVars removes environment variables whose names match common secret patterns
+// (password, secret, token, key) from the env slice so they are not passed to the apk subprocess
+// and are not captured in build-info environment records.
+func filterSecretEnvVars(env []string) []string {
+	// Respect the user-configured exclusion pattern from JFROG_CLI_ENV_EXCLUDE.
+	excludePattern := os.Getenv(coreutils.EnvExclude)
+	if excludePattern == "" {
+		excludePattern = "*password*;*secret*;*token*;*key*"
```
agrasth: This default pattern is semicolon-separated (`"*password*;*secret*;*token*;*key*"`) and split on `;` at line 717, but the established default for `JFROG_CLI_ENV_EXCLUDE` elsewhere in the codebase (`jfrog-cli-core/common/build/buildinfoproperties.go`) is comma-separated: `"*password*,*psw*,*secret*,*key*,*token*,*auth*"`. A user who sets this env var using the documented comma syntax (as they would for every other package manager) will have it parsed here as one unmatched pattern, silently disabling filtering. The hardcoded fallback is also missing `*psw*`/`*auth*` compared to the canonical default. Should this reuse the same default/split convention as the rest of the codebase?

artifactory/commands/alpine/apkupload.go
```diff
@@ -0,0 +1,726 @@
...
+		return "", entities.Checksum{}, fmt.Errorf("empty output from apk info %q", name)
+	}
+
+	tmpDir, err := os.MkdirTemp("", "apk-dep-chk-*")
+	if err != nil {
+		return id, entities.Checksum{}, nil
+	}
+	defer func() { _ = os.RemoveAll(tmpDir) }()
+
+	fetchCmd := exec.Command("apk", "fetch", "--output", tmpDir, name)
```
agrasth: `resolveDepInfo` shells out to `apk fetch` per dependency to download it just to compute a checksum — this is the "one API/network call per dependency" pattern the team generally wants to avoid (see the batched-AQL approach already used in `apkcommand.go`'s `enrichChecksumsFromAQL`). For uploads with many transitive deps this could be slow and hammer the package index/CDN. Could this batch through the same AQL/local-cache-first strategy instead?

artifactory/commands/alpine/apkupload.go
```diff
@@ -0,0 +1,726 @@
...
+	// If the resolved repo is virtual, resolve it to its default deployment (local) repo.
+	apkCmd.repoKey, err = resolveLocalUploadRepo(apkCmd.repoKey, apkCmd.serverDetails)
+	if err != nil {
+		return err
+	}
+
+	// Artifactory only indexes Alpine packages deployed under
+	// <repo>/<alpine-version>/<branch>/<arch>/<file>. Omitting the branch segment
+	// leaves the package outside every APKINDEX, so it is always included here.
+	target := fmt.Sprintf("%s/%s/%s/%s/%s", apkCmd.repoKey, apkCmd.alpineVersion, apkCmd.branch, apkCmd.arch, filename)
```
agrasth, line 223: `filename` goes through sanitization elsewhere, but `apkCmd.branch` (`--branch`, defaults to `"main"`) and `apkCmd.arch` (`--arch`) are spliced into this Artifactory path with no validation or escaping. If either comes from a raw user/CI-supplied string (e.g. a git branch name passed straight through as `--branch`), a value containing `../` could push the upload target outside the intended `<repo>/<version>/<branch>/<arch>/` path. Worth rejecting/escaping path separators in `branch`/`arch` before building `target` (same applies to the duplicate construction at lines 344 and 355).

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,764 @@
...
+		return
+	}
+
+	moduleID := apkCmd.buildConfiguration.GetModule()
+	if moduleID == "" {
+		moduleID = buildName + ":" + buildNumber
+	}
+	// Remove explicitly requested packages from the pre-snapshot so they always
+	// appear in the diff — even if they were already installed before this run.
+	effectivePreSnapshot := excludeRequestedPackages(preSnapshot, apkCmd.apkArgs)
```
agrasth: This passes the raw `apkCmd.apkArgs` (still containing flag values like `--build-name`'s value, `--project`, `--module`, `--alpine-version`) instead of the already-cleaned `nativeArgs`/`requestedPkgs` computed above at lines 155/157. If a flag value happens to match an already-installed package name, `excludeRequestedPackages` would incorrectly exclude that package from the pre-snapshot and corrupt the resulting diff.

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,764 @@
...
+			emitSignatureHint(exitErr)
+			return exitErr.ExitCode(), nil
+		}
+		return 1, err
+	}
+	return 0, nil
+}
+
+// emitSignatureHint prints a remediation hint when apk exits with an RSA signature error.
+func emitSignatureHint(exitErr *exec.ExitError) {
```
agrasth: This reads `exitErr.Stderr` to decide whether to print the signature-verification hint, but `Stderr` on `*exec.ExitError` is only populated by `cmd.Output()`, not by `cmd.Run()`/`Wait()` (which is what `gofrogcmd.RunCmd` uses here). As written, `exitErr.Stderr` should always be empty, so this hint likely never fires regardless of what `apk` printed — might need to capture stderr into a buffer explicitly to make this reachable.

artifactory/commands/setup/setup.go
```diff
@@ -612,3 +626,358 @@ func (sc *SetupCommand) configureHelm() error {
...
+	// Point apk exclusively at the Artifactory repository: replace the entire file with the
+	// single Artifactory line so all package resolution flows through Artifactory. Public
+	// mirrors (e.g. dl-cdn.alpinelinux.org) are intentionally dropped — use a virtual repo
+	// that proxies the upstream Alpine mirror if packages from public repos are still needed.
+	content := repoURL + "\n"
+
+	if err = apkWriteFile(apkRepositoriesFile, content, 0600); err != nil {
+		// Restore on failure so the system isn't left in a broken state.
+		if originalContent != "" {
+			_ = apkWriteFile(apkRepositoriesFile, originalContent, 0600)
```
agrasth: The restore-on-failure discards its own error (`_ = apkWriteFile(...)`). If the write failed for a reason that also breaks the restore (e.g. disk full, permissions changed mid-run), the user only sees "failed to write %s" and has no idea `/etc/apk/repositories` may now be missing or truncated. Worth surfacing the restore error too, e.g. joining it into the returned error.

artifactory/commands/setup/setup.go
naveenku-jfrog: The `--apply`/preview gate is intentionally not wired in this PR. `jf setup apk` applies configuration unconditionally, consistent with the existing `jf setup <pm>` commands (npm, pip, go, etc.), which all write their config directly rather than gating behind a preview flag. The `--apply` reference in the description is outdated — apk is handled the same way as the other package managers here instead of introducing a divergent flag.

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,785 @@
...
+	"github.com/jfrog/jfrog-client-go/artifactory/services"
+	specutils "github.com/jfrog/jfrog-client-go/artifactory/services/utils"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+)
+
+var buildInfoSubcmds = map[string]bool{
+	"add":     true,
+	"upgrade": true,
+}
```
reshmifrog, line 31: why is this needed?

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,785 @@
...
+
+// jfFlagSet is the set of jf-specific flags that must be stripped before forwarding
+// args to the native apk binary. Each flag may appear as "--flag value" or "--flag=value".
+var jfFlagSet = map[string]bool{
+	"--build-name":     true,
```
reshmifrog, line 36: No need to check jfFlagSet , we have a func named isBuildConfigurationSet , please use that.

agrasth (approved):

agrasth (dismissed):

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,787 @@
...
+	return fmt.Sprintf("%s:%s:%s", repoKey, arch, alpineVersion)
+}
+
+// shouldIsolateRepo reports whether this command should target its --repo on the selected server
+// via a temporary repositories file instead of /etc/apk/repositories. It isolates only when a
+// --repo is set AND the selected server (serverDetails.ServerId) differs from the default server
+// that /etc/apk/repositories is assumed to be configured with (via `jf setup apk`). This runs the
+// one-off command against the requested server while leaving the default server's persistent
+// configuration untouched.
+func (apkCmd *ApkCommand) shouldIsolateRepo() bool {
```
reshmifrog, line 118: why are we doing this?

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,787 @@
...
+			log.Warn("--no-cache prevents apk from keeping the downloaded archives, so checksums " +
+				"cannot be computed locally and will only be filled for packages found in the " +
+				"Artifactory repository. Drop --no-cache for complete Build Info.")
+		}
+		log.Debug("--no-cache detected: skipping temp cache dir; local checksums require a cached .apk.")
+	case userCacheDir != "":
+		cacheDir = userCacheDir
+		log.Debug("Reusing caller-provided --cache-dir for checksum collection:", cacheDir)
+	case needsBuildInfo:
+		cacheDir, err = createTempDir()
```
reshmifrog, line 210: why are we creating the cache Dir?

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,787 @@
...
+		downloadsDir = cacheDir
+	}
+	apkCmd.collectBuildInfo(preSnapshot, cacheDir, downloadsDir, requestedPkgs)
+	return nil
+}
+
+func unsupportedBuildInfoMessage(commandName string) string {
+	return fmt.Sprintf(
+		"Build Info flags were provided, but Build Info collection is not available for 'apk %s'. "+
+			"Only 'add', 'upgrade', and 'upload' support Build Info. The command was executed as a passthrough.",
```
reshmifrog, line 268: i think we should not support upgrade and add commands for build info collection right?

artifactory/commands/alpine/apkcommand.go
```diff
@@ -0,0 +1,787 @@
...
+		}
+	}
+	return ""
+}
+
+// extractPackageNames returns the non-flag tokens from the args list,
+// which represent the package names the user explicitly requested.
+// apkValueFlags are native apk flags that consume the following token as their value.
+// Their value tokens must not be mistaken for requested package names.
+var apkValueFlags = map[string]bool{
```
reshmifrog, line 713: why are we maintaining the list for apk flags?

artifactory/commands/alpine/apkupload.go
```diff
@@ -0,0 +1,769 @@
...
+			continue
+		}
+		seen[id] = struct{}{}
+		var requestedBy [][]string
+		if uploadedPkgID != "" {
+			requestedBy = [][]string{{uploadedPkgID}}
+		}
+		deps = append(deps, entities.Dependency{
+			Id:          id,
+			Scopes:      []string{bibuild.AlpineScopeProd},
```
reshmifrog, line 468: why are we explicitly setting the scope of the dependency?

artifactory/commands/alpine/apkupload.go
```diff
@@ -0,0 +1,769 @@
...
+	var missing []int
+	for i, dep := range deps {
+		if dep.Sha256 == "" && dep.Sha1 == "" {
+			missing = append(missing, i)
+		}
+	}
+	if len(missing) == 0 {
+		return deps
+	}
+	sm, err := artutils.CreateServiceManager(apkCmd.serverDetails, -1, 0, false)
```
reshmifrog: what is this -1 and 0 values ? also can you please use const for it?

reshmifrog (changes requested):

artifactory/commands/alpine/apkcommand.go
naveenku-jfrog: Needed so we only collect Build Info for install-side commands (add/upgrade). Other apk subcommands (update/del/info/…) are passthrough; if BI flags are present we warn and skip collection. Upload is handled separately by ApkUploadCommand.

artifactory/commands/alpine/apkcommand.go
naveenku-jfrog: That function doesn’t exist. We already gate collection with IsCollectBuildInfo(). jfFlagSet/stripJFFlags is a different concern: strip JF-only flags before forwarding to native apk. The CLI already extracts these via ExtractBuildDetailsFromArgs / ExtractStringOptionFromArgs; this is defense-in-depth if ApkCommand is called with raw args. Can’t replace one with the other.

artifactory/commands/alpine/apkcommand.go
naveenku-jfrog: So a one-off --repo/--server-id doesn’t mutate the default /etc/apk/repositories from jf setup apk. We write a temp --repositories-file only when the selected server ≠ default server. Covered by TestApk_NonDefaultServerDoesNotMutateRepositoriesFile.

artifactory/commands/alpine/apkcommand.go
naveenku-jfrog: apk deletes downloads with --no-cache. For BI we need the .apk archives to compute local checksums, so we create a temp --cache-dir when BI is on (reuse caller’s --cache-dir if set). Cleaned up after the run.

artifactory/commands/alpine/apkcommand.go
naveenku-jfrog: Intentional — install-side BI is a core part of this feature (same idea as jf npm install / jf yarn). add/upgrade use pre/post snapshot diffs; upload records artifacts. Integration tests (TestApkAdd_BasicBuildInfo, TestApkUpgrade_BuildInfo) cover this.

artifactory/commands/alpine/apkcommand.go
naveenku-jfrog: So extractPackageNames doesn’t treat flag values as packages. e.g. apk add --cache-dir /tmp/cache curl must yield ["curl"], not ["/tmp/cache","curl"]. Those tokens feed scopes/requestedBy.

artifactory/commands/alpine/apkupload.go
naveenku-jfrog: APK has no dev/test deps — first-level deps are prod (matches build-info-go’s AlpineScopeProd and the add/upgrade flow). Leaving scope empty would break Xray/BI consumers that expect it.

artifactory/commands/alpine/apkupload.go
naveenku-jfrog: Addressed it.

agrasth (approved):

udaykb2 (approved):

reshmifrog (approved):

PR #499, opened 2026-06-29
RTECO-1536 - Add native RubyGems/Bundler command with auth and build-info

agrasth:
## Summary

Implements the core business logic for the `jf ruby gem|bundle` native command in jfrog-cli-artifactory, covering authentication injection, repository discovery, native tool execution, and build-info collection.

**Depends on**: https://github.com/jfrog/build-info-go/pull/394

### What is included in this PR

- **RubyCommand dispatcher**: Routes to `gem` or `bundle` with full stdio passthrough
- **Authentication injection** (non-destructive, per-process env vars only):
  - Bundler: `BUNDLE_<HOST>` with full host-key normalization (. -> __, - -> ___, etc.)
  - RubyGems: `GEM_HOST_API_KEY` with proper `Basic base64(user:pass)` encoding
  - Embedded URL credentials for `gem install/fetch --source` (for index downloads)
  - Respects existing native credentials (skips injection if already set)
  - Host-match safety: warns and skips when source host differs from jf server host
  - Help bypass: no auth for help/-h/--help requests
  - Skips injection when no Artifactory source is discovered (even with --server-id)
- **`--repo` flag**: Constructs the full Artifactory gems API URL from server config + repo name, eliminating the need to pass full URLs. Auto-injects `--source`/`--host` for gem install/push.
- **Repository discovery** (precedence: --repo (constructed URL) > --source/--host arg > Gemfile source > gem sources --list)
- **Build-info collection**:
  - Dependencies: via RubygemsFlexPack for `bundle install/update/lock/add` and opportunistic `gem install/fetch`
  - Artifacts: `gem build` (local .gem checksums) and `gem push` (+ build property tagging in Artifactory)
  - AQL checksum enrichment: single batched query with $or clauses for all deps
  - Dependency repo path included in build-info output
- **Unit tests** covering auth, repo extraction, help detection, host matching, error handling, credential embedding, URL construction, and source arg injection

### Bug fixes in this update

- **Bug 1**: `GEM_HOST_API_KEY` now encodes as `Basic base64(user:pass)` — Artifactory expects this format
- **Bug 2**: `gem install/fetch` embeds credentials directly in `--source` URL for index downloads (`specs.4.8.gz`) since `GEM_HOST_API_KEY` is not used for those requests in RubyGems 3.x
- **Bug 3**: Explicit no-args error before help bypass — prevents silent fallthrough to `gem help`
- **Bug 4**: Skips credential injection when no Artifactory gem source is discovered (even with `--server-id`)
- **Dep path**: Build-info dependencies now include the Artifactory repository path

### Auth level

Server-level (host-level) — one credential covers all repos on the same Artifactory instance. Not repo-level.

### What will be in follow-up PRs

- **Future enhancements**: Local gem cache checksum scan, scope classification (prod/dev/test), `jf setup` integration for Ruby, paginated AQL, cached build-info reuse, multi-source Gemfile auth

## Test plan

- [x] `go test ./artifactory/commands/ruby/... -v` - all tests pass
- [x] `go build ./...` - compiles successfully (with local build-info-go replace)
- [x] `go vet ./artifactory/commands/ruby/...` - no issues
- [ ] End-to-end testing with real Artifactory gems repo

## Summary by CodeRabbit

* **New Features**
  * Added native RubyGems and Bundler command execution.
  * Added automatic repository discovery and secure authentication for configured repositories.
  * Added Ruby setup support for Bundler mirrors, credentials, and RubyGems sources.
  * Added optional build-info collection for published gems and dependencies.
* **Bug Fixes**
  * Improved credential cleanup and prevented credentials from being sent to unrelated registries.
  * Preserved existing Ruby and Bundler configuration while preventing duplicate entries.
* **Tests**
  * Added comprehensive coverage for Ruby commands, authentication, configuration, dependencies, and build metadata.


artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+// when build parameters are supplied, collects build info.
+func (rc *RubyCommand) Run() error {
+	if rc.nativeTool == "" {
+		rc.nativeTool = toolGem
+	}
+	if rc.nativeTool != toolGem && rc.nativeTool != toolBundle {
+		return fmt.Errorf("unsupported ruby tool %q: expected 'gem' or 'bundle'", rc.nativeTool)
+	}
+
+	// Bug 3 fix: explicit no-args check before help bypass so we don't
```
reshmifrog: remove this comment, wdym by bug 3 fix?

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+	// silently fall into gem help when no subcommand is given.
+	if len(rc.args) == 0 {
+		return fmt.Errorf("no subcommand provided for '%s'. Usage: jf ruby %s <subcommand> [args...]", rc.nativeTool, rc.nativeTool)
+	}
+
+	subCommand := rubySubCommand(rc.args)
+
+	// Help requests bypass auth injection entirely so credentials are never
+	// printed in help output (same rationale as the UV native command).
```
reshmifrog: this comment does not look good , refactor it and write exactly what this func is doing .

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+		}()
+		defer func() {
+			signal.Stop(signals)
+			close(signals)
+			runCleanup()
+		}()
+	}
+
+	log.Info(fmt.Sprintf("Running %s %s.", rc.nativeTool, subCommand))
+	// For gem install/fetch, capture stdout to parse "Successfully installed"/"Downloaded" lines.
```
reshmifrog: let's not parse the stdout , because the format can change and it would break our code

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+				log.Warn("Failed to collect Ruby build info: " + biErr.Error())
+			}
+		}
+	}
+	return nil
+}
+
+// rubyEmbedCredsInSourceArg rewrites --source/--host URL args to embed credentials
+// for gem install/fetch. RubyGems 3.x uses embedded URL credentials for index downloads
+// (specs.4.8.gz) but does NOT use GEM_HOST_API_KEY for those requests.
```
reshmifrog: this comment is very confusing , not able to understand what exactly this function is doing.

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+	// from it. Both names below are constants, and no caller-supplied value reaches them.
+	if !filepath.IsAbs(homeDir) {
+		return nil, fmt.Errorf("home directory %q is not an absolute path", homeDir)
+	}
+	gemDir := filepath.Clean(filepath.Join(homeDir, ".gem"))
+	credFile := filepath.Clean(filepath.Join(gemDir, "credentials"))
+
+	// Ensure ~/.gem directory exists.
+	if err := os.MkdirAll(gemDir, 0700); err != nil {
+		return nil, fmt.Errorf("could not create ~/.gem directory: %w", err)
```
reshmifrog, line 282: why are we creating this the setup command would create it right?

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+func (rc *RubyCommand) injectAuth(serverDetails *coreConfig.ServerDetails, sourceURL string) []string {
+	user, pass := rubyCredentials(serverDetails)
+	if !rubyHasCredentials(pass) {
+		log.Debug("Ruby auth: no username/password/token available in server config; relying on native configuration")
+		return nil
+	}
+
+	// Determine the host to authenticate. Prefer the discovered source URL host;
+	// otherwise fall back to the Artifactory server host.
+	host := rubyHostOf(sourceURL)
```
reshmifrog, line 605: it has to be always artifactory right? what do mean by the ` Prefer the discovered source URL host; otherwise fall back to the Artifactory server host.`

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+	buildName, err := rc.buildConfiguration.GetBuildName()
+	if err != nil {
+		return err
+	}
+	buildNumber, err := rc.buildConfiguration.GetBuildNumber()
+	if err != nil {
+		return err
+	}
+
+	// For gem install/fetch: parse stdout to determine exactly what was installed/fetched.
```
reshmifrog: we should not parse stdout anywhere , please check it everywhere and try ot not use it.

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+
+	if len(bi.Modules) > 0 && len(bi.Modules[0].Dependencies) > 0 && repoKey != "" && serverDetails != nil {
+		directURLDeps := collector.GetDirectURLDeps()
+		rubyEnrichDepsChecksums(bi.Modules[0].Dependencies, repoKey, directURLDeps, serverDetails)
+	} else if repoKey == "" {
+		log.Info("Ruby build-info: no Artifactory gems repo discovered — dependency checksum enrichment skipped. " +
+			"Point your Gemfile/gem source at an Artifactory gems repository or pass --server-id.")
+	}
+
+	if err := rubySaveBuildInfo(bi, rc.buildConfiguration); err != nil {
```
reshmifrog, line 1119: why we have rubySave BuildInfo? i think we already have a function named SaveBuidInfo why seperate for ruby?

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1820 @@
...
+		return fmt.Errorf("failed to save RubyGems build info: %w", err)
+	}
+	log.Info(fmt.Sprintf("RubyGems build info collected. Use '%s' to publish.", rubyPublishHint(rc.buildConfiguration, buildName, buildNumber)))
+	return nil
+}
+
+// collectGemBuildDependencies records the dependencies a `gem build` was built against,
+// read from Gemfile.lock. Without a lock file there is nothing resolved to record, and
+// saying so is better than accepting --build-name and silently producing nothing.
+func (rc *RubyCommand) collectGemBuildDependencies(workingDir, repoKey string, serverDetails *coreConfig.ServerDetails) error {
```
reshmifrog, line 1129: i see this same function repeating , remove the duplicate method.

reshmifrog (changes requested):

artifactory/commands/ruby/native_ruby_test.go
```diff
@@ -0,0 +1,615 @@
...
+	// -v without space (unusual but valid)
+	assert.Equal(t, "1.0.0", extractVersionFromArgs([]string{"install", "gem", "-v1.0.0"}))
+}
+
+func TestRubyWriteTempGemCredentials(t *testing.T) {
+	// Use a temporary home directory to avoid touching real ~/.gem/credentials
+	origHome := os.Getenv("HOME")
+	tmpHome := t.TempDir()
+	os.Setenv("HOME", tmpHome)
+	defer os.Setenv("HOME", origHome)
```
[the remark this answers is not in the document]
agrasth: Fixed — all three affected tests now also `t.Setenv("USERPROFILE", tmpHome)` alongside `HOME`.

artifactory/commands/setup/setup_test.go
```diff
@@ -923,3 +924,148 @@ func TestSetupCommand_MavenCorrupted(t *testing.T) {
...
+	existing := "BUNDLE_PATH: \"vendor/bundle\"\nBUNDLE_OTHERHOST__COM: other:creds\n"
+	require.NoError(t, os.WriteFile(filepath.Join(bundleDir, "config"), []byte(existing), 0600))
+
+	require.NoError(t, writeBundleConfig("my.jfrog.io", "admin", "secret"))
+
+	content, err := os.ReadFile(filepath.Join(bundleDir, "config"))
+	require.NoError(t, err)
+	assert.Contains(t, string(content), "BUNDLE_PATH: vendor/bundle")
+	assert.Contains(t, string(content), "BUNDLE_OTHERHOST__COM: other:creds")
+	assert.Contains(t, string(content), "BUNDLE_MY__JFROG__IO: admin:secret")
```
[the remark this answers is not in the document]
agrasth: This was marked addressed but the underlying gap actually remained: `writeBundleSettings`/`addGemrcSource` wrote with `os.WriteFile(path, out, 0600)`, which does not tighten the mode of a file that already exists (Go only applies the mode on creation). If `~/.bundle/config` or `~/.gemrc` pre-existed at a looser mode, the credentials we merged in would stay under it. No test caught this either — the closest one (`TestAddGemrcSource_PreservesUnrelatedKeys`) writes an existing 0644 file but only asserts content, not final mode. Now routed through `permissions.WriteFileOwnerOnly`, the existing helper for exactly this case (used by pip/uv), which chmods before and after the write.

artifactory/commands/ruby/native_ruby.go
```diff
@@ -0,0 +1,1808 @@
...
+
+		// Detect `end` closing a group block.
+		if line == "end" && currentGroups != nil {
+			if nested > 0 {
+				nested--
+			} else {
+				currentGroups = nil
+			}
+			continue
+		}
```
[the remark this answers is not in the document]
agrasth: Fixed — `opensRubyBlock` now also matches `... do |x|` (regex on the trailing `do\s*|...|`) and bare `if`/`unless`/`case`/`begin`/`while`/`until` openers, not just lines ending in the literal `do`.

artifactory/commands/ruby/native_ruby.go
agrasth: Fixed — reworded to describe what the check does instead of referencing an internal bug number with no context here.

artifactory/commands/ruby/native_ruby.go
agrasth: Reworded — it was pointing at "the UV native command" for the rationale instead of just stating it directly.

artifactory/commands/ruby/native_ruby.go
agrasth: Reworded to state directly why GEM_HOST_API_KEY doesn't cover this path (RubyGems 3.x reads the index via URL-embedded credentials for this specific request, not the env var).

artifactory/commands/ruby/native_ruby.go
agrasth: This is deliberate, not an oversight: `jf ruby` is a config-less flow (see the doc comment at the top of ruby.go) that does not assume `jf setup ruby` ran first, so `gem push` needs to be able to create `~/.gem` itself on a machine where setup was never run.

artifactory/commands/ruby/native_ruby.go
agrasth: The logic is correct as written: `authorizedForSource` gates every credential path on the target host matching the configured Artifactory host (or an explicit `--server-id`), so in practice this never authenticates against an arbitrary registry — only the discovered source or the server config host, whichever resolved. Happy to reword the comment further if the wording itself is still unclear — let me know what specifically read as ambiguous.

artifactory/commands/ruby/native_ruby.go
agrasth: Not a duplicate — `rubySaveBuildInfo` is a thin per-command wrapper around the shared `bld.SaveBuildInfo(bi)`, the same pattern the UV command uses (see `native_uv.go`), not a reimplementation of it.

artifactory/commands/ruby/native_ruby.go
agrasth: Removed `rubyEmbedCredsInHostArg` — it was dead code (zero callers) and nearly identical to `rubyEmbedCredsInSourceArg`. If this is the duplication you meant, it's gone now. `collectGemBuildDependencies` itself isn't a duplicate of anything — it delegates to the shared `collectLockfileDependencies`, the same helper the bundle-install path uses.

artifactory/commands/ruby/native_ruby_test.go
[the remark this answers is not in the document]
agrasth: Fixed — all three tests now also set `USERPROFILE` (in addition to `HOME`), since `os.UserHomeDir()` reads `USERPROFILE` on Windows.

artifactory/commands/ruby/native_ruby.go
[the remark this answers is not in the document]
agrasth: Fixed — `opensRubyBlock` now also matches `... do |x|` and bare `if`/`unless`/`case`/`begin`/`while`/`until` openers, not just lines ending in the literal `do`.

artifactory/commands/ruby/native_ruby.go
reshmifrog: okay understood but if we are assuming that setup cmd never ran then while doing the push how it will know which registry/repo to push to the command is anyway going to fail right?
Please elaborate how create /.gem file while pushing is valid.

artifactory/commands/ruby/native_ruby.go
reshmifrog: so the comment is not correct then right? because it says that it will fallback to artifactory.

reshmifrog (changes requested):

artifactory/commands/ruby/native_ruby.go
agrasth: Fixed in 73c2eec — the primary mechanism no longer parses stdout at all. It now snapshots `Gem::Specification` (RubyGems' own API, via `ruby -e`) before and after the command runs and diffs the two, the same approach build-info-go's RubygemsFlexPack now uses for Bundler's dependency graph. This also closes a real gap the old mechanism had: stdout parsing produced nothing under `--quiet`.

artifactory/commands/ruby/native_ruby.go
agrasth: Reworded in 73c2eec. The comment was only describing which hostname string gets used to build the credential env-var candidates below — it was never the authorization decision, which is `authorizedForSource` a few lines down. That function does restrict to Artifactory's own host by default; an explicit `--server-id` is the one documented case where a different host is allowed. Made that explicit in the comment so it doesn't read as a blanket "always Artifactory" claim.

artifactory/commands/ruby/native_ruby.go
agrasth: Done a full pass on this in 73c2eec. Beyond the gem install/fetch stdout parsing (previous comment), two more text-parsing paths are removed: `bundleInstalledPackages`/`parseBundleListLine` (parsed `bundle list` stdout) and `parseGemfileGroups` and its helpers (a hand-rolled Gemfile parser with regex block-tracking for if/unless/case/begin/while/until — its own doc comment still said ungrouped gems get "production", the same mislabel already fixed on the build-info-go side).

artifactory/commands/ruby/native_ruby.go
agrasth: Checked this properly — the two are unrelated, so push does not depend on setup having run. `jf setup ruby` only ever writes `~/.gemrc` and `~/.bundle/config` (confirmed in setup.go) — it never touches `~/.gem/credentials`. And `jf ruby gem push` gets its target repo from the `--repo` flag plus jfrog-cli's own server config (via rubyResolveServerDetails), not from anything setup writes. So `jf ruby gem push --repo my-repo my.gem` resolves its target the same way whether or not setup ever ran. The `~/.gem/credentials` write exists purely because RubyGems' own push_command.rb hardcodes reading that file for Basic-auth credentials keyed by the resolved --host — that's a RubyGems requirement, not something tied to our setup command at all.

reshmifrog (dismissed):

fluxxBot (approved):

udaykb2 (approved):

PR #500, opened 2026-07-03
RTECO-1560 - Fix agentplugins install issue

udaykb2:
-----

### PR Summary: Agent Plugins Installation Guide & Global-Only Scope Enforcement
**Overview**
Added comprehensive documentation and enforced installation scope constraints for JFrog CLI's jf agent plugins install command across three AI agents (Claude, Cursor, Codex).

**Key Changes**
1. Global-Only Installation Scope
All three agents now support global scope only - project-scoped installations are no longer supported
Added validation to reject --project-dir attempts with agent-specific error messages explaining the architectural constraints
Each agent provides clear guidance: Claude and Codex config are user-scoped only; Cursor only auto-discovers from global directory

2. Standardized Installation Paths
Claude: ~/.claude/plugins/local/jfrog/<repoKey>/<slug>/
Cursor: ~/.cursor/plugins/local/<repoKey>/<slug>/.  (because cursor only identifies the plugins installed under /local)
Codex: ~/.agents/plugins/local/jfrog/<repoKey>/<slug>/

3. Repo Key Isolation
Each Artifactory repository gets isolated subdirectories and marketplace manifests
Prevents plugin name collisions when installing from multiple repos
Allows multiple repos with same plugin names via @repoKey syntax

4. Comprehensive Installation Guide (INSTALL_GUIDE.md)
Documents complete installation flow: download → extract → register → metadata
Agent-specific marketplace structures and native CLI interactions
Scope constraints with rationale for each agent
Collision prevention strategies
Troubleshooting guide with real examples

**Files Modified**
agent/plugins/commands/install/install.go - Added project scope validation
agent/plugins/common/agents.go - Updated paths, marked as global-only
agent/plugins/common/hooks_claude.go - Updated for new paths
agent/plugins/common/hooks_codex.go - Updated for new paths
agent/plugins/INSTALL_GUIDE.md - NEW - Complete installation documentation

**Testing**
All agent/plugins/... unit tests pass
Agent plugin configurations properly validated
Marketplace structures correctly generated for each agent

### DIFF
--- a/agent/common/install_summary.go
+++ b/agent/common/install_summary.go
@@ -9,6 +9,27 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/log"
 )
 
+// WarningError wraps an error that should be reported as a warning, not a fatal failure.
+type WarningError struct {
+	msg string
+}
+
+// NewWarningError creates a warning error that is not fatal but should be reported to the user.
+func NewWarningError(msg string) *WarningError {
+	return &WarningError{msg: msg}
+}
+
+// Error implements the error interface.
+func (we *WarningError) Error() string {
+	return we.msg
+}
+
+// IsWarning returns true if an error is a WarningError.
+func IsWarning(err error) bool {
+	_, ok := err.(*WarningError)
+	return ok
+}
+
 // Summary row statuses for install/update tables and JSON output.
 const (
 	SummaryStatusOK        = "ok"
@@ -41,6 +62,18 @@ func InstallFailureRow(agentName, scope, destinationDir string, err error) Summa
 		Path:   destinationDir,
 		Status: SummaryStatusFailed,
 		Detail: err.Error(),
+	}
+}
+
+// InstallWarningRow builds a warning install/update summary row for one target.
+// Warning rows indicate partial success (e.g., files copied but native registration skipped).
+func InstallWarningRow(agentName, scope, destinationDir, detail string) SummaryRow {
+	return SummaryRow{
+		Agent:  agentName,
+		Scope:  scope,
+		Path:   destinationDir,
+		Status: SummaryStatusWarning,
+		Detail: detail,
 	}
 }
 
--- a/agent/common/install_summary.go
+++ b/agent/common/install_summary.go
@@ -14,6 +14,7 @@ const (
 	SummaryStatusOK        = "ok"
 	SummaryStatusFailed    = "failed"
 	SummaryStatusSkipped   = "skipped"
+	SummaryStatusWarning   = "warning"
 	SummaryDetailOKInstall = "Executed successfully with no issues."
 )
 
--- a/agent/plugins/commands/install/install.go
+++ b/agent/plugins/commands/install/install.go
@@ -234,6 +234,11 @@ func (ic *InstallCommand) CopyExtractedToTargets(unzipDir string, installTargets
 			results = append(results, agentcommon.InstallFailureRow(target.Agent.Name, string(target.Scope), target.DestinationDir, err))
 			continue
 		}
+		if hookErr := plugincommon.RunPostInstallHook(target.Agent.Name, ic.slug, ic.version, target.DestinationDir, ic.repoKey); hookErr != nil {
+			log.Warn(fmt.Sprintf("post-install hook for agent %q: %s", target.Agent.Name, hookErr))
+		} else {
+			log.Info(fmt.Sprintf("post-install hook completed for agent %q", target.Agent.Name))
+		}
 		results = append(results, agentcommon.SummaryRow{
 			Agent:  target.Agent.Name,
 			Scope:  string(target.Scope),
@@ -273,10 +278,25 @@ func (ic *InstallCommand) resolveAgentTargetDirectories() ([]plugincommon.AgentT
 	if ic.scope == agentcommon.InstallScopeProject && ic.projectDir == "" {
 		return nil, fmt.Errorf("project directory is required for project-scoped install")
 	}
+	if ic.scope == agentcommon.InstallScopeProject {
+		for _, agent := range ic.agents {
+			if strings.ToLower(agent.Name) == "cursor" {
+				return nil, fmt.Errorf(
+					"cursor does not support project-scoped plugin installs: " +
+						"Cursor only auto-discovers full plugins from ~/.cursor/plugins/local/. " +
+						"Use --global to install there instead.",
+				)
+			}
+		}
+	}
 	isGlobal := ic.scope == agentcommon.InstallScopeGlobal
 	// Path is "" because harness mode uses project or global scope
 	// e.g., jf agent plugins install web --harness claude --global
-	return agentcommon.ResolveAgentTargets(ic.slug, "", ic.agents, ic.projectDir, isGlobal)
+	targets, err := agentcommon.ResolveAgentTargets(ic.slug, "", ic.agents, ic.projectDir, isGlobal)
+	if err != nil {
+		return nil, err
+	}
+	return plugincommon.InjectRepoKey(targets, ic.repoKey), nil
 }
 
 func (ic *InstallCommand) writePluginInfoManifest(target plugincommon.AgentTarget) error {
--- a/agent/plugins/commands/install/install.go
+++ b/agent/plugins/commands/install/install.go
@@ -234,11 +234,11 @@ func (ic *InstallCommand) CopyExtractedToTargets(unzipDir string, installTargets
 			results = append(results, agentcommon.InstallFailureRow(target.Agent.Name, string(target.Scope), target.DestinationDir, err))
 			continue
 		}
-		if hookErr := plugincommon.RunPostInstallHook(target.Agent.Name, ic.slug, ic.version, target.DestinationDir, ic.repoKey); hookErr != nil {
-			log.Warn(fmt.Sprintf("post-install hook for agent %q: %s", target.Agent.Name, hookErr))
-		} else {
-			log.Info(fmt.Sprintf("post-install hook completed for agent %q", target.Agent.Name))
-		}
+	if hookErr := plugincommon.RunPostInstallHook(target.Agent.Name, ic.slug, ic.version, target.DestinationDir, ic.repoKey); hookErr != nil {
+		log.Warn(fmt.Sprintf("post-install hook for agent %q: %s", target.Agent.Name, hookErr))
+		results = append(results, agentcommon.InstallFailureRow(target.Agent.Name, string(target.Scope), target.DestinationDir, hookErr))
+	} else {
+		log.Info(fmt.Sprintf("post-install hook completed for agent %q", target.Agent.Name))
 		results = append(results, agentcommon.SummaryRow{
 			Agent:  target.Agent.Name,
 			Scope:  string(target.Scope),
@@ -246,6 +246,7 @@ func (ic *InstallCommand) CopyExtractedToTargets(unzipDir string, installTargets
 			Status: agentcommon.SummaryStatusOK,
 			Detail: agentcommon.SummaryDetailOKInstall,
 		})
+	}
 	}
 	return results
 }
@@ -283,22 +284,22 @@ func (ic *InstallCommand) resolveAgentTargetDirectories() ([]plugincommon.AgentT
 			agentLower := strings.ToLower(agent.Name)
 			if agentLower == "claude" {
 				return nil, fmt.Errorf(
-					"claude does not support project-scoped plugin installs: "+
-						"Claude plugin configuration is user-scoped only (~/.claude/settings.json). "+
+					"claude does not support project-scoped plugin installs: " +
+						"Claude plugin configuration is user-scoped only (~/.claude/settings.json). " +
 						"Use --global to install there instead",
 				)
 			}
 			if agentLower == "cursor" {
 				return nil, fmt.Errorf(
-					"cursor does not support project-scoped plugin installs: "+
-						"Cursor only auto-discovers full plugins from ~/.cursor/plugins/local/. "+
+					"cursor does not support project-scoped plugin installs: " +
+						"Cursor only auto-discovers full plugins from ~/.cursor/plugins/local/. " +
 						"Use --global to install there instead",
 				)
 			}
 			if agentLower == "codex" {
 				return nil, fmt.Errorf(
-					"codex does not support project-scoped plugin installs: "+
-						"Codex plugin configuration is user-scoped only (~/.codex/config.toml). "+
+					"codex does not support project-scoped plugin installs: " +
+						"Codex plugin configuration is user-scoped only (~/.codex/config.toml). " +
 						"Use --global to install there instead",
 				)
 			}
--- a/agent/plugins/commands/install/install.go
+++ b/agent/plugins/commands/install/install.go
@@ -151,7 +151,7 @@ func (ic *InstallCommand) Run() error {
 	}
 
 	for _, result := range results {
-		if result.Status != agentcommon.SummaryStatusOK {
+		if result.Status == agentcommon.SummaryStatusFailed {
 			return fmt.Errorf("installation failed for one or more agents (see summary above)")
 		}
 	}
@@ -235,8 +235,14 @@ func (ic *InstallCommand) CopyExtractedToTargets(unzipDir string, installTargets
 			continue
 		}
 	if hookErr := plugincommon.RunPostInstallHook(target.Agent.Name, ic.slug, ic.version, target.DestinationDir, ic.repoKey); hookErr != nil {
-		log.Warn(fmt.Sprintf("post-install hook for agent %q: %s", target.Agent.Name, hookErr))
-		results = append(results, agentcommon.InstallFailureRow(target.Agent.Name, string(target.Scope), target.DestinationDir, hookErr))
+		if agentcommon.IsWarning(hookErr) {
+			log.Warn(fmt.Sprintf("post-install hook for agent %q: %s", target.Agent.Name, hookErr))
+			results = append(results, agentcommon.InstallWarningRow(target.Agent.Name, string(target.Scope), target.DestinationDir,
+				fmt.Sprintf("Plugin files installed successfully but native registration incomplete: %s", hookErr.Error())))
+		} else {
+			log.Warn(fmt.Sprintf("post-install hook for agent %q: %s", target.Agent.Name, hookErr))
+			results = append(results, agentcommon.InstallFailureRow(target.Agent.Name, string(target.Scope), target.DestinationDir, hookErr))
+		}
 	} else {
 		log.Info(fmt.Sprintf("post-install hook completed for agent %q", target.Agent.Name))
 		results = append(results, agentcommon.SummaryRow{
--- a/agent/plugins/commands/install/install.go
+++ b/agent/plugins/commands/install/install.go
@@ -35,7 +35,7 @@ type InstallCommand struct {
 }
 
 func NewInstallCommand() *InstallCommand {
-	return &InstallCommand{scope: agentcommon.InstallScopeProject}
+	return &InstallCommand{scope: agentcommon.InstallScopeGlobal}
 }
 
 func (ic *InstallCommand) SetServerDetails(details *config.ServerDetails) *InstallCommand {
--- a/agent/plugins/commands/install/install.go
+++ b/agent/plugins/commands/install/install.go
@@ -283,23 +283,23 @@ func (ic *InstallCommand) resolveAgentTargetDirectories() ([]plugincommon.AgentT
 			agentLower := strings.ToLower(agent.Name)
 			if agentLower == "claude" {
 				return nil, fmt.Errorf(
-					"claude does not support project-scoped plugin installs: " +
-						"Claude plugin configuration is user-scoped only (~/.claude/settings.json). " +
-						"Use --global to install there instead.",
+					"claude does not support project-scoped plugin installs: "+
+						"Claude plugin configuration is user-scoped only (~/.claude/settings.json). "+
+						"Use --global to install there instead",
 				)
 			}
 			if agentLower == "cursor" {
 				return nil, fmt.Errorf(
-					"cursor does not support project-scoped plugin installs: " +
-						"Cursor only auto-discovers full plugins from ~/.cursor/plugins/local/. " +
-						"Use --global to install there instead.",
+					"cursor does not support project-scoped plugin installs: "+
+						"Cursor only auto-discovers full plugins from ~/.cursor/plugins/local/. "+
+						"Use --global to install there instead",
 				)
 			}
 			if agentLower == "codex" {
 				return nil, fmt.Errorf(
-					"codex does not support project-scoped plugin installs: " +
-						"Codex plugin configuration is user-scoped only (~/.codex/config.toml). " +
-						"Use --global to install there instead.",
+					"codex does not support project-scoped plugin installs: "+
+						"Codex plugin configuration is user-scoped only (~/.codex/config.toml). "+
+						"Use --global to install there instead",
 				)
 			}
 		}
--- a/agent/plugins/commands/install/install.go
+++ b/agent/plugins/commands/install/install.go
@@ -280,10 +280,25 @@ func (ic *InstallCommand) resolveAgentTargetDirectories() ([]plugincommon.AgentT
 	}
 	if ic.scope == agentcommon.InstallScopeProject {
 		for _, agent := range ic.agents {
-			if strings.ToLower(agent.Name) == "cursor" {
+			agentLower := strings.ToLower(agent.Name)
+			if agentLower == "claude" {
+				return nil, fmt.Errorf(
+					"claude does not support project-scoped plugin installs: " +
+						"Claude plugin configuration is user-scoped only (~/.claude/settings.json). " +
+						"Use --global to install there instead.",
+				)
+			}
+			if agentLower == "cursor" {
 				return nil, fmt.Errorf(
 					"cursor does not support project-scoped plugin installs: " +
 						"Cursor only auto-discovers full plugins from ~/.cursor/plugins/local/. " +
+						"Use --global to install there instead.",
+				)
+			}
+			if agentLower == "codex" {
+				return nil, fmt.Errorf(
+					"codex does not support project-scoped plugin installs: " +
+						"Codex plugin configuration is user-scoped only (~/.codex/config.toml). " +
 						"Use --global to install there instead.",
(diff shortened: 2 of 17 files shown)


agent/plugins/common/install_targets.go
```diff
@@ -13,3 +18,30 @@ const (
...
+// InjectRepoKey rewrites the DestinationDir of Claude and Codex targets to
+// include the Artifactory repo key as a subdirectory. This gives each repo its
+// own isolated marketplace directory so installs from different repos never
+// overwrite each other's marketplace registration.
+//
+//   Claude: <GlobalDir>/<slug>  →  <GlobalDir>/<repoKey>/<slug>
+//   Codex:  <GlobalDir>/<slug>  →  <GlobalDir>/<repoKey>/<slug>
+//
+// Cursor and --path targets are returned unchanged.
+func InjectRepoKey(targets []AgentTarget, repoKey string) []AgentTarget {
```
agrasth, line 31: Major: `InjectRepoKey`'s `switch` only handles `"claude"` and `"codex"` — Cursor falls through untouched. That directly contradicts the doc comment in `agents.go` (`Cursor: <GlobalDir>/<repoKey>/<slug>`), which promises the same per-repo isolation for Cursor. As written, two different Artifactory repos publishing a plugin with the same slug will both resolve to `~/.cursor/plugins/local/<slug>` and overwrite each other's install. Add a `case "cursor":` identical to the claude/codex branches (or, if Cursor truly shouldn't be isolated, fix the misleading comment in agents.go instead).

agent/plugins/commands/install/install.go
```diff
@@ -234,6 +234,11 @@ func (ic *InstallCommand) CopyExtractedToTargets(unzipDir string, installTargets
 			results = append(results, agentcommon.InstallFailureRow(target.Agent.Name, string(target.Scope), target.DestinationDir, err))
 			continue
 		}
+		if hookErr := plugincommon.RunPostInstallHook(target.Agent.Name, ic.slug, ic.version, target.DestinationDir, ic.repoKey); hookErr != nil {
```
agrasth: Major: `hookErr` from `RunPostInstallHook` is only logged via `log.Warn` — the `SummaryRow` appended a few lines below is unconditionally `Status: SummaryStatusOK, Detail: SummaryDetailOKInstall` ("Executed successfully with no issues."). Since `PrintInstallSummary` renders that Status/Detail to the user, the exact ticket scenario (Claude CLI not found / not working) will still print full success and hide that native marketplace registration failed. Consider threading hookErr into a distinct non-fatal status (e.g. a "partial"/"warning" row) so the CLI output actually reflects it.

agent/plugins/common/hooks_codex.go
```diff
@@ -0,0 +1,211 @@
...
+//	~/.agents/plugins/local/<repoKey>/          ← marketplace root
+//	  .agents/plugins/
+//	    marketplace.json                       ← written here
+//	  <slug>/                                  ← installDir (plugin files copied by jf)
+func codexPostInstall(slug, version, installDir, repoKey string) error {
+	manifestPath := codexMarketplaceManifestPath(installDir)
+	log.Info(fmt.Sprintf("[codex] writing marketplace entry for '%s' → %s", slug, manifestPath))
+	if err := upsertCodexMarketplaceEntry(manifestPath, slug, repoKey); err != nil {
+		log.Warn("codex post-install: failed to update marketplace manifest:", err)
+		return nil
```
agrasth: Major: on a real failure to write the Codex marketplace manifest, this returns `nil` after only logging a warning — silently treating a disk/permission error as success. Compare with `claudePostInstall` (hooks_claude.go), which correctly does `return err` for the equivalent `upsertLocalMarketplaceEntry` failure. `codexPostDelete` has the same swallow-on-failure pattern for `removeCodexMarketplaceEntry`. Return the error here to match Claude's behavior.

agent/plugins/common/hooks_claude.go
```diff
@@ -0,0 +1,93 @@
...
+func claudePostInstall(slug, version, installDir, repoKey string) error {
+	marketplacePath := claudeMarketplacePath(installDir)
+	log.Info(fmt.Sprintf("[claude] writing marketplace entry for '%s' → %s", slug, marketplacePath))
+	if err := upsertLocalMarketplaceEntry(marketplacePath, slug, version, repoKey); err != nil {
+		return err
+	}
+	if _, err := lookPathClaude(); err != nil {
+		log.Warn("[claude] claude CLI not found on PATH; skipping native marketplace registration. " +
+			"Install the Claude CLI to complete native plugin registration.")
+		return err
```
agrasth: Minor: this returns `err` (the `LookPath` failure) when the Claude CLI is missing, but `agent_hooks.go`'s `PostInstallFn` doc comment says the function "should log a warning and return nil" in that exact case, since "the file installation already succeeded." Because it returns non-nil, `install.go` logs a second, redundant warning on top of the detailed one already logged two lines above. Same mismatch exists in `claudePostDelete` and `codexPostInstall`/`codexPostDelete`. Either return `nil` here to match the documented contract, or update the doc comment to describe the actual (propagate-the-error) behavior.

agent/plugins/common/agent_hooks.go
```diff
@@ -0,0 +1,46 @@
...
+	fn, ok := agentPostInstallHooks[strings.ToLower(agentName)]
+	if !ok {
+		return nil
+	}
+	return fn(slug, version, installDir, repoKey)
+}
+
+// RunPostDeleteHook executes the registered post-delete hook for agentName.
+// Returns nil if no hook is registered.
+func RunPostDeleteHook(agentName, slug, installDir, repoKey string) error {
```
agrasth: Minor: `RunPostDeleteHook` (and the `agentPostDeleteHooks` map, `claudePostDelete`, `codexPostDelete` behind it) has zero call sites anywhere in the repo — there's no local "uninstall" command yet (the only `delete` command removes a version from Artifactory, not a local install). This is ~60 lines of unreachable production code shipped ahead of the feature that would call it. Either wire it into a real uninstall path in this PR, or drop it until that path exists.

agent/plugins/common/install_targets.go
```diff
@@ -13,3 +18,30 @@ const (
...
+			slug := filepath.Base(t.DestinationDir)
+			t.DestinationDir = filepath.Join(base, repoKey, slug)
+		case "codex":
+			base := filepath.Dir(t.DestinationDir)
+			slug := filepath.Base(t.DestinationDir)
+			t.DestinationDir = filepath.Join(base, repoKey, slug)
+		}
+		result[i] = t
+	}
+	return result
```
agrasth, line 46: Minor (Testing): `InjectRepoKey` is exported and brand new but has no direct unit test — existing tests only exercise the downstream path-builder helpers with manually pre-built paths, assuming `InjectRepoKey` already produced them correctly. A table test over claude/codex/cursor inputs would have caught the missing Cursor case (see the Major finding on line 31) before merge.

agent/plugins/commands/install/install.go
```diff
@@ -273,10 +278,40 @@ func (ic *InstallCommand) resolveAgentTargetDirectories() ([]plugincommon.AgentT
 	if ic.scope == agentcommon.InstallScopeProject && ic.projectDir == "" {
 		return nil, fmt.Errorf("project directory is required for project-scoped install")
 	}
+	if ic.scope == agentcommon.InstallScopeProject {
+		for _, agent := range ic.agents {
+			agentLower := strings.ToLower(agent.Name)
+			if agentLower == "claude" {
+				return nil, fmt.Errorf(
+					"claude does not support project-scoped plugin installs: "+
```
agrasth: Nit: `gofmt -l` flags this file (missing space before the `+` string-concat operators in these `fmt.Errorf` calls). `hooks_claude.go`, `hooks_codex.go`, and `install_targets.go` are flagged too (comment indentation / struct-tag alignment). Run `gofmt -w` — this also matches the unchecked "I used gofmt" box in the PR description.

agrasth (changes requested):
## Ticket Alignment
- **Implements**: both reported "wrong install directory" bugs are fixed — Claude now installs under `~/.claude/plugins/local` and Cursor under `~/.cursor/plugins/local` (agents.go), and Claude's "it does not work" complaint is addressed by wiring `claude plugin marketplace add` / `claude plugin install` into a new post-install hook.
- **Missing from the ticket**: Cursor's "it does not make cursor import it" is still unaddressed — no Cursor CLI/import call exists anywhere in this PR, and `InjectRepoKey` only covers claude/codex (see Finding 1), so Cursor plugins are still not auto-imported and are not even given the per-repo isolation the code comments claim they get.
- **Out of scope vs. the ticket**: a full Codex integration (native `codex` CLI hooks, marketplace manifest, per-repo isolation) was added even though the ticket only reports bugs for Claude and Cursor; likewise project-scoped installs were removed entirely for all three agents, which is a broader architectural change than the two reported bugs.

## Findings

### Major
1. **`InjectRepoKey` silently skips Cursor** (agent/plugins/common/install_targets.go) — contradicts the doc comment in `agents.go` that promises `Cursor: <GlobalDir>/<repoKey>/<slug>` isolation. Cursor installs from two different Artifactory repos with the same plugin slug land in the same directory (`~/.cursor/plugins/local/<slug>`) and overwrite each other. This is the exact class of bug (Cursor install path) the ticket is about, and it has zero test coverage (see Finding 6).
2. **Post-install hook failures never surface in the CLI's output** (agent/plugins/commands/install/install.go) — `hookErr` is only passed to `log.Warn`; the `SummaryRow` appended right after is unconditionally `Status: SummaryStatusOK, Detail: "Executed successfully with no issues."`. So running the exact scenario from the ticket (Claude CLI missing/misconfigured) still prints full success to the user, hiding that native marketplace registration never happened.
3. **Inconsistent swallowing of marketplace-write errors between Claude and Codex** (agent/plugins/common/hooks_codex.go) — `codexPostInstall` logs and returns `nil` when `upsertCodexMarketplaceEntry` fails, treating a real disk/permission error as success, while `claudePostInstall` correctly propagates the equivalent error via `return err`. `codexPostDelete` has the same swallow-on-failure pattern.

### Minor
4. **Hook implementations violate their own documented contract** (agent/plugins/common/agent_hooks.go doc-comment vs. hooks_claude.go / hooks_codex.go) — `PostInstallFn`'s doc says "If the native agent CLI is absent the function should log a warning and return nil," but `claudePostInstall`, `claudePostDelete`, `codexPostInstall`, and `codexPostDelete` all `return err` in that branch instead. Net effect is a duplicated/confusing warning: the hook itself already logs a detailed message, then install.go logs a second generic one for the same condition.
5. **`RunPostDeleteHook` and everything behind it is dead code** — `agentPostDeleteHooks`, `claudePostDelete`, and `codexPostDelete` have zero call sites anywhere in the repository (verified with a full-repo grep); there is no local "uninstall" command yet (only a remote `delete` that removes a version from Artifactory). ~60 lines of unreachable production code shipped speculatively, only indirectly unit-tested.
6. **No test exercises `InjectRepoKey` directly** (agent/plugins/common/install_targets.go) — existing tests only check the downstream path-builder helpers (`claudeMarketplacePath`, `codexMarketplaceManifestPath`) with pre-built paths; nothing calls `InjectRepoKey` itself for claude, codex, or cursor. A simple table test would have caught Finding 1 immediately.

### Nit
7. **`gofmt` is not clean** on 4 of the changed files (`hooks_claude.go`, `hooks_codex.go`, `install_targets.go`, `install.go`) — confirmed by running `gofmt -l`. Mostly comment-indentation and struct-tag alignment. Matches the unchecked "I used gofmt" box in the PR description.

agent/plugins/common/install_targets.go
udaykb2: If I add repoKey/slug folder, it is not recognising /slug as plugin, instead it is showing /repoKey as plugin. Which is wrong. So this implementation is correct

agent/plugins/common/install_targets.go
udaykb2: added tests

agent/plugins/commands/install/install.go
```diff
@@ -234,6 +234,11 @@ func (ic *InstallCommand) CopyExtractedToTargets(unzipDir string, installTargets
 			results = append(results, agentcommon.InstallFailureRow(target.Agent.Name, string(target.Scope), target.DestinationDir, err))
 			continue
 		}
+	if hookErr := plugincommon.RunPostInstallHook(target.Agent.Name, ic.slug, ic.version, target.DestinationDir, ic.repoKey); hookErr != nil {
```
agrasth, line 237: ### Finding 1 - Major 🟠

**Issue:** A `RunPostInstallHook` error is reported as a full `InstallFailureRow` even though the plugin files were already copied successfully, and a missing native CLI (the common case) is reported as full `SummaryStatusOK` success even though no native registration happened.

**Impact:** Users get a hard "installation failed" for a plugin that's actually on disk, or a false "success" when Claude/Codex never actually registered the plugin — the exact silent-failure mode reported in the ticket. The unused `SummaryStatusWarning` constant suggests a warning-status row was intended here but never wired in.

**Suggested fix:** Only fail the row when the hook error indicates the file-based marketplace write itself failed; when the native CLI is simply absent, return `SummaryStatusWarning` with a detail explaining registration was skipped, instead of `SummaryStatusOK`.

agent/plugins/common/hooks_claude.go
```diff
@@ -0,0 +1,78 @@
...
+		ClaudeExec("plugin", "marketplace", "add", marketplaceDir)
+		log.Info(fmt.Sprintf("[claude] installing plugin: claude plugin install %s@%s", slug, repoKey))
+		// Include the @<repoKey> qualifier so Claude resolves the correct marketplace source.
+		ClaudeExec("plugin", "install", slug+"@"+repoKey)
+	} else {
+		// CLI not found, log warning but continue (not a fatal error)
+		log.Warn("[claude] claude CLI not found on PATH; skipping native marketplace registration. " +
+			"Install the Claude CLI to complete native plugin registration.")
+	}
+	return nil
```
agrasth, line 54: ### Finding 2 - Major 🟠

**Issue:** `claudePostInstall` always `return nil` once the `claude` binary is found on `PATH`, regardless of whether `ClaudeExec("plugin", "marketplace", "add", ...)` or `ClaudeExec("plugin", "install", ...)` actually succeeded — `ClaudeExec` has no return value and only logs a warning on failure.

**Impact:** A real native registration failure (bad marketplace.json shape, version mismatch, plugin rejected by Claude) is reported identically to success, so the CLI's own success/failure signal for this feature can't be trusted — the same failure mode the ticket describes as "it does not work."

**Suggested fix:** Have `ClaudeExec`/`CodexExec` return an error (or a success bool) and propagate it out of `claudePostInstall`/`codexPostInstall` so a failed native registration surfaces as a warning/failure row instead of silent success.

agent/plugins/common/hooks_codex.go
```diff
@@ -0,0 +1,198 @@
...
+		CodexExec("plugin", "marketplace", "add", root)
+		log.Info(fmt.Sprintf("[codex] installing plugin: codex plugin add %s@%s", slug, repoKey))
+		// Include the @<repoKey> qualifier so Codex resolves the correct marketplace source.
+		CodexExec("plugin", "add", slug+"@"+repoKey)
+	} else {
+		// CLI not found, log warning but continue (not a fatal error)
+		log.Warn("[codex] codex CLI not found on PATH; skipping native marketplace registration. " +
+			"Run: codex plugin marketplace add " + codexMarketplaceRoot(installDir))
+	}
+	return nil
```
agrasth, line 98: ### Finding 2b (same as Finding 2) - Major 🟠

**Issue:** Same unconditional `return nil` pattern as `claudePostInstall` — `CodexExec` failures for `plugin marketplace add` / `plugin add` are only logged, never surfaced, so `codexPostInstall` always reports success once the `codex` binary is found.

**Suggested fix:** Same as Finding 2 — propagate exec failures instead of swallowing them.

agent/common/install_summary.go
```diff
@@ -14,6 +14,7 @@ const (
 	SummaryStatusOK        = "ok"
 	SummaryStatusFailed    = "failed"
 	SummaryStatusSkipped   = "skipped"
+	SummaryStatusWarning   = "warning"
```
agrasth, line 38: ### Finding 4 - Minor 🔵

**Issue:** `SummaryStatusWarning` is added but never referenced anywhere else in the diff.

**Impact:** Dead code, and a signal that the intended "installed but native CLI registration skipped" warning state (see Finding 1) was never actually wired up.

**Suggested fix:** Use this status for the missing-CLI case in `CopyExtractedToTargets`, or remove it if it's truly unused for now.

agent/plugins/common/agents.go
```diff
@@ -11,9 +11,23 @@ type AgentSpec = agentcommon.AgentSpec
...
-	"codex":  {GlobalDir: "~/.codex/plugins", ProjectDir: ".codex/plugins"},
+	// GlobalDir behavior varies by agent:
+	// - Claude & Codex: The Artifactory repo key is injected as a subdirectory at install time
+	//   so each repo gets its own isolated marketplace directory:
+	//     Claude: <GlobalDir>/<repoKey>/<slug>
+	//     Codex:  <GlobalDir>/<repoKey>/<slug>
+	// - Cursor: No repo key injection; paths are used as-is:
+	//     Cursor: <GlobalDir>/<slug>
+	//
+	// Paths ending in /jfrog are rooted under ~/.jfrog for JFrog-specific plugin configurations.
```
agrasth, line 22: ### Finding 5 - Minor 🔵

**Issue:** "Paths ending in /jfrog are rooted under ~/.jfrog..." is incorrect — `~/.claude/plugins/local/jfrog` and `~/.agents/plugins/local/jfrog` are rooted under `~/.claude` and `~/.agents`, not `~/.jfrog`. The identical comment was also copy-pasted into `agent/skills/common/agents.go:14`, where no path even ends in `/jfrog`.

**Suggested fix:** Reword to describe what the `/jfrog` suffix actually does (namespaces the JFrog-managed marketplace within the agent's own plugin directory), and drop the comment from `agent/skills/common/agents.go` where it doesn't apply.

agent/plugins/common/install_targets.go
```diff
@@ -13,3 +18,30 @@ const (
...
+// overwrite each other's marketplace registration.
+//
+//	Claude: <GlobalDir>/<slug>  →  <GlobalDir>/<repoKey>/<slug>
+//	Codex:  <GlobalDir>/<slug>  →  <GlobalDir>/<repoKey>/<slug>
+//
+// Cursor and --path targets are returned unchanged.
+func InjectRepoKey(targets []AgentTarget, repoKey string) []AgentTarget {
+	result := make([]AgentTarget, len(targets))
+	for i, t := range targets {
+		switch strings.ToLower(t.Agent.Name) {
```
agrasth, line 34: ### Finding 6 - Nit 🟢

**Issue:** The `"claude"` and `"codex"` cases in this `switch` have identical bodies and could be merged into `case "claude", "codex":`. Separately, `repoKey` flows into `filepath.Join` with no character validation (e.g. `..`), which is new since this is the first place `repoKey` drives a local filesystem path rather than an HTTP repo name.

**Suggested fix:** Collapse the duplicate cases; consider validating `repoKey` doesn't contain path separators or `..` before using it in a destination path.

agrasth (changes requested):
## Ticket Alignment

- **Implements the ask:** Yes for the core install-path bugs — Claude now installs under `~/.claude/plugins/local/jfrog/<repoKey>/<slug>` (was `~/.claude/plugins`), Cursor under `~/.cursor/plugins/local/<slug>` (was `~/.cursor/plugins`), both agents are restricted to global-only scope, and new post-install hooks call `claude plugin marketplace add` / `claude plugin install` (and the Codex equivalents) so the native CLI registers the plugin.
- **Appears missing / at risk:** The ticket's "it does not work" complaint for Claude is not reliably closed — `ClaudeExec`/`CodexExec` swallow all native-CLI errors (log-only, no error returned), so a failed `claude plugin install` still reports the install as a full success (see finding below). There's also no test exercising the "native CLI missing from PATH" or "hook failed" branches, which is exactly the scenario QA hit.
- **Out of scope / side effects:** `jf agent plugins update` and `update --all` were not updated for the new `<repoKey>` directory nesting used by Claude/Codex, so updating a plugin installed by this PR's `install` command will silently report "plugin not installed" for those two agents. The linked Jira comment says update/list changes are tracked separately (likely RTECO-1567), but merging this PR alone leaves `update` non-functional for Claude/Codex in the interim — worth flagging to the ticket reporter given they are explicitly testing both install and (implicitly) the follow-up experience. A stray, unrelated comment was also added to `agent/skills/common/agents.go` that doesn't apply to that file's paths.

## Findings by Severity

**Major**
1. `agent/plugins/commands/install/install.go` (`CopyExtractedToTargets`): a post-install hook failure is reported via `InstallFailureRow`, telling the user the *whole install failed* and causing `RunInstall` to return a hard error — even though `EnsureDestinationDir`/`CopyDir`/`writePluginInfoManifest` already succeeded and the plugin files are genuinely on disk. Conversely, when the native CLI isn't on `PATH`, `claudePostInstall`/`codexPostInstall` return `nil` and this same code path reports `SummaryStatusOK` / `SummaryDetailOKInstall` ("Executed successfully with no issues.") even though native registration never ran. The newly added `SummaryStatusWarning` constant (`agent/common/install_summary.go`) is never referenced anywhere, which strongly suggests this degraded-but-installed case was meant to surface as a warning row and got left unwired.
2. `agent/plugins/common/hooks_claude.go` / `hooks_codex.go`: `ClaudeExec`/`CodexExec` execute the native CLI and only `log.Warn` on failure — they have no return value, so `claudePostInstall`/`codexPostInstall` unconditionally `return nil` once the CLI binary is found, regardless of whether `plugin marketplace add` or `plugin install` actually succeeded. A real registration failure (bad marketplace format, version mismatch, etc.) is indistinguishable from success in the CLI's own output, which is precisely the class of bug ("it does not work") this PR is meant to close.
3. `agent/plugins/commands/update/update.go`: `runUpdateOnSlug` (via `agentcommon.ResolveAgentTargets`) and `discoverInstalledPluginTargets` resolve the destination as `<GlobalDir>/<slug>`, but `agent/plugins/common/install_targets.go`'s new `InjectRepoKey` (only called from `install.go`) nests Claude/Codex installs one level deeper at `<GlobalDir>/<repoKey>/<slug>`. After installing with this PR, `jf agent plugins update --slug <slug> --harness claude` will look in the wrong directory, hit `ReadInstalledPluginVersion`'s `os.ErrNotExist` path, and report `"plugin not installed at ...; run 'jf agent plugins install' first"` even though it is installed. `update --all`'s `DiscoverInstalledPluginSlugs` has the same one-level-too-shallow problem, so it will silently find zero plugins for Claude/Codex. This is disclosed in the Jira comment as follow-up work, but nothing in this PR guards against or messages the broken interim state.

**Minor**
4. `agent/plugins/commands/install/install.go` lines 237-247 are not `gofmt`-clean (indentation is one level off after the `if hookErr` block was added), which contradicts the PR checklist's "I used gofmt for formatting the code before submitting" — confirmed locally via `gofmt -l`.
5. No test covers the "native CLI not found on PATH" branch or the `RunPostInstallHook` error branch in `CopyExtractedToTargets` — the exact conditions from the bug report. All hook tests stub `lookPathClaude`/`lookPathCodex` to "found" (`hooks_marketplace_test.go` inits them to return `nil` error).
6. Misleading comment: `agent/plugins/common/agents.go:22` and `agent/skills/common/agents.go:14` both say "Paths ending in /jfrog are rooted under ~/.jfrog for JFrog-specific ... configurations," but the actual paths shown (`~/.claude/plugins/local/jfrog`, `~/.agents/plugins/local/jfrog`) are rooted under `~/.claude` / `~/.agents`, not `~/.jfrog`. In `agent/skills/common/agents.go` none of the listed paths even end in `/jfrog`, so the comment doesn't apply to anything in that file.

**Nit**
7. `agent/plugins/common/install_targets.go`'s `InjectRepoKey` has identical bodies for the `"claude"` and `"codex"` cases (`filepath.Dir`/`filepath.Base`/`filepath.Join`) — could be collapsed to `case "claude", "codex":`. Also, `repoKey` (user-supplied via `--repo`, resolved by `ResolveRepo`) is joined into the filesystem path with no character validation; low risk since it's local-user-controlled, but worth a sanity check against `..`/path separators given it now drives real filesystem writes rather than just an HTTP repo name.


---

**Additional notes** (these referenced lines outside the PR diff, so they couldn't be attached inline):

- **`agent/plugins/commands/update/update.go:143`** — ### Finding 3 - Major 🟠

**Issue:** `agentcommon.ResolveAgentTargets` resolves the destination as `<GlobalDir>/<slug>`, but `install.go`'s new `InjectRepoKey` (called only from install, not from update) nests Claude/Codex installs one level deeper at `<GlobalDir>/<repoKey>/<slug>`. `discoverInstalledPluginTargets` (used by `update --all`) has the same mismatch via `DiscoverInstalledPluginSlugs`.

**Impact:** After installing a plugin with this PR's `jf agent plugins install --harness claude`, running `jf agent plugins update --slug <slug> --harness claude` reports "plugin not installed at ...; run 'jf agent plugins install' first" even though it is installed, and `update --all` silently discovers zero plugins for Claude/Codex.

**Suggested fix:** Either apply `plugincommon.InjectRepoKey` in `update.go`'s target resolution too (both `runUpdateOnSlug` and `discoverInstalledPluginTargets`), or explicitly document/guard that `update` for Claude/Codex requires the follow-up ticket before merging this install-path change, so users don't hit a confusing "not installed" error for plugins that are, in fact, installed.

agent/plugins/commands/install/install.go
udaykb2: done

agent/plugins/common/hooks_codex.go
udaykb2: done

agent/plugins/common/hooks_claude.go
udaykb2: done

agent/plugins/common/agents.go
udaykb2: not needed

agent/plugins/common/install_targets.go
udaykb2: not needed

agent/common/install_summary.go
udaykb2: not needed

agrasth (dismissed):

naveenku-jfrog (approved):

fluxxBot (approved):

PR #503, opened 2026-07-09
RTECO-1564 - APT auth command command support

fluxxBot:
-----

## Summary

- New `artifactory/commands/apt/` package with two auth flows:
  - **On-the-fly** (`AptCommand`): writes a temp sources.list with embedded credentials, runs `apt-get` scoped exclusively to Artifactory via `Dir::Etc::sourcelist` + `Dir::Etc::sourceparts=-`, then removes the file
  - **Persistent** (`AptSetupCommand`): writes `/etc/apt/sources.list.d/jfrog-<repo>-<dist>.list` + `/etc/apt/preferences.d/jfrog-<repo>-<dist>.pref` (Pin-Priority: 1001)
- `--trusted` flag injects `[trusted=yes]` (testing without GPG)
- `--import-key` flag auto-detects `primaryKeyPairRef` via `GetRepository`, falls back to `/api/gpg/key/public`, writes `[signed-by=<path>]`
- `--remove` flag removes all managed apt files, optionally filtered by `--dist`
- Wired `project.Apt` into `SetupCommand` (`setup/setup.go`): adds `"debian"` to `packageManagerToRepositoryPackageType` for interactive repo selection, dispatches to `configureApt()` for interactive dist/component/GPG prompts via `jf setup`

## Implementation notes

- Credentials embedded in sources.list URL (not `apt_auth.conf`): simpler, no extra file
- `FetchAndInstallPublicKey` uses `CreateServiceManager` + `sm.GetRepository` (no custom HTTP) and `sm.Client().SendGet` for the key download — same HTTP client as all other commands (proxy/TLS/retry-aware)
- `keyringsDir`, `sourcesListDir`, `preferencesDir` declared as `var` for test patching
- Permission errors on write surface a sudo hint via `wrapPermErr`
- `writeSourcesListIdempotent` uses `strings.Contains` to avoid duplicate entries on re-run

## Summary by CodeRabbit

* **New Features**
  * Added authenticated APT repository setup and package operations for Debian/Ubuntu.
  * Supports temporary or persistent configuration, GPG key management, trusted mode, repository pinning, and cleanup/removal.
  * Added interactive APT configuration during project setup.
  * Improved validation, secure credential handling, and guidance for elevated permissions.
* **Tests**
  * Added comprehensive coverage for APT commands, configuration, key handling, cleanup, validation, and permissions.


artifactory/commands/apt/auth_test.go
```diff
@@ -0,0 +1,242 @@
...
+		case "/artifactory/api/security/keypair/mykey/public":
+			_, _ = w.Write([]byte(fakeKey))
+		default:
+			http.NotFound(w, r)
+		}
+	}))
+	defer srv.Close()
+
+	tmpDir := t.TempDir()
+	origKeyringsDir := keyringsDir
```
naveenku-jfrog, line 154: Bug: Default component to main in the on-the-fly Run path; otherwise a missing --component silently disables auth injection.

artifactory/commands/apt/auth.go
```diff
@@ -0,0 +1,180 @@
...
+		return "", fmt.Errorf("chmod temp sources list: %w", err)
+	}
+	if _, err = f.WriteString(content); err != nil {
+		_ = f.Close()
+		_ = os.Remove(f.Name())
+		return "", fmt.Errorf("write temp sources list: %w", err)
+	}
+	if err = f.Close(); err != nil {
+		_ = os.Remove(f.Name())
+		return "", fmt.Errorf("close temp sources list: %w", err)
```
naveenku-jfrog, line 54: suggestion: Consider /etc/apt/auth.conf.d/*.conf (600) instead of embedding creds in sources.list (skill §1 "no secret leak to config files"); at minimum document the accepted risk.

artifactory/commands/apt/command.go
```diff
@@ -0,0 +1,193 @@
...
+				log.Warn("Failed to create temporary sources.list — proceeding without auth injection: " + err.Error())
+			} else {
+				defer func() { _ = os.Remove(tmpPath) }()
+				// Dir::Etc::sourcelist replaces the main sources.list; Dir::Etc::sourceparts=-
+				// disables sources.list.d/ so ONLY the temp Artifactory entry is live for this
+				// command — packages cannot resolve to any other configured repository.
+				sourceOpts := []string{
+					"-o", "Dir::Etc::sourcelist=" + tmpPath,
+					"-o", "Dir::Etc::sourceparts=-",
+				}
```
naveenku-jfrog, line 190: suggestion:
sourceparts=- makes install resolve only via Artifactory — add a test/doc for the partial-local-repo dependency-resolution failure mode (skill §17).

artifactory/commands/apt/setup.go
```diff
@@ -0,0 +1,321 @@
...
+// writePinningFile writes an apt preferences file that gives Artifactory
+// packages priority 1001 — higher than any default (990) or pinned (1000)
+// source. apt will prefer Artifactory when both it and another repo have
+// the package; if only another repo has it, apt still installs from there.
+// This is not a hard block but prevents silent downgrades to non-Artifactory
+// versions when both sources carry the same package.
+func writePinningFile(path, artHost string) error {
+	content := fmt.Sprintf("Package: *\nPin: origin %s\nPin-Priority: 1001\n", artHost)
+	return os.WriteFile(path, []byte(content), 0644)
+}
```
naveenku-jfrog, line 301: nit:
Pin-Priority: 1001 forces downgrades; comment says "not a hard block." Reconcile or use 990 unless hard preference is intended; document.

naveenku-jfrog (dismissed):

artifactory/commands/apt/setup.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	removed := 0
+	for _, dir := range []struct{ path, ext string }{
+		{sourcesListDir, ".list"},
+		{preferencesDir, ".pref"},
+		{keyringsDir, ".asc"},
+	} {
+		// Glob only the fixed jfrog-* prefix, avoiding glob metacharacters in dist.
+		matches, err := filepath.Glob(filepath.Join(dir.path, "jfrog-*"))
```
agrasth, line 219: `--remove` never filters by `c.repoName` — it globs every `jfrog-*` file in `sourcesListDir`/`preferencesDir`/`keyringsDir` and only narrows by `--dist`. 

`--repo` is a registered flag for `AptSetup` (see `commandsflags.go`'s `AptSetup` flag list, which includes `setupRepo`), so a user running `jf setup apt --repo=my-repo --remove` will reasonably expect only `my-repo`'s config to be deleted — instead this wipes **every** repo's jfrog-managed apt config on the host (and `--remove --dist=noble` wipes `noble` for every repo, not just the target one). `TestRunRemove_RemovesAllDists` in `setup_test.go` documents this as intended, but it reads like a real footgun for multi-repo hosts (e.g. CI runners shared across teams). Suggest scoping the glob/suffix match to `c.repoName` when it's set, same as the existing `--dist` filtering.

artifactory/commands/apt/setup.go
```diff
@@ -0,0 +1,321 @@
...
+		// config (e.g. "... noble main") as already present when the file holds a
+		// broader line ("... noble main contrib"), silently keeping stale config.
+		for _, line := range strings.Split(strings.TrimRight(string(existing), "\n"), "\n") {
+			if line == sourceLine {
+				return false, nil
+			}
+		}
+		log.Info("Updating existing apt source configuration.")
+	}
+	return true, os.WriteFile(targetFile, []byte(sourceLine+"\n"), 0600)
```
agrasth: `os.WriteFile`'s mode argument is only applied when the file is *created*; on an existing file it's a no-op (Go stdlib semantics — `OpenFile` with `O_TRUNC`, no `O_CREATE` path re-chmod). 

Since `targetFile` can already exist here (this is the idempotent-update path), if it ever ends up with looser permissions than 0600 — e.g. written by an older binary, or manually edited — a subsequent `jf setup apt` run will silently keep those looser perms instead of tightening them, even though this file embeds credentials in the URL.

artifactory/commands/apt/auth.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	artURL := strings.TrimSuffix(serverDetails.GetArtifactoryUrl(), "/")
+	repoPath := artURL + "/" + repoName
+
+	parsed, err := url.Parse(repoPath)
+	if err != nil {
+		return "", fmt.Errorf("parse artifactory URL: %w", err)
+	}
+	parsed.User = url.UserPassword(user, password)
```
agrasth, line 150: Embedding credentials in the source URL (`parsed.User = url.UserPassword(...)`) means they end up wherever apt prints or logs that URL — progress lines (`Get:1 https://user:pass@host/...`), 404/error messages, and potentially `/var/log/apt/term.log`. 

Modern apt masks userinfo in most output, but the CI matrix in `aptTests.yml` spans apt versions from Ubuntu 20.04/Debian 11 (older apt) through Debian 13/Ubuntu 24.04 — has credential redaction been verified across that whole range, particularly on error paths (e.g. a 404 for a missing package, or a signature failure) where apt's message formatting differs from normal `Get:` progress lines? If any of those distros print the raw URL on stdout/stderr, that's a credential leak straight into CI console logs given `command.go` wires apt's stdout/stderr directly to the process's own.

artifactory/commands/apt/setup.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+// writePinningFile writes an apt preferences file that gives Artifactory
+// packages priority 1001 — higher than any default (990) or pinned (1000)
+// source. apt will prefer Artifactory when both it and another repo have
+// the package; if only another repo has it, apt still installs from there.
+// This is not a hard block but prevents silent downgrades to non-Artifactory
+// versions when both sources carry the same package.
+func writePinningFile(path, artHost string) error {
+	content := fmt.Sprintf("Package: *\nPin: origin %s\nPin-Priority: 1001\n", artHost)
```
agrasth, line 299: apt's `Pin: origin <host>` only matches by hostname, not by path/repo — so if two different Artifactory Debian repos are served from the same Artifactory host (very common: one instance, many repos), running `jf setup apt` for repo A also raises Pin-Priority to 1001 for *any* package apt resolves from that host, including ones served via a separate, unrelated repo B's sources entry (e.g. one a teammate configured manually, or a different `jf setup apt --repo=B` on the same host).

This is an apt limitation (no path-based pinning), so may not be fixable here, but it's worth calling out explicitly in the `--help`/design doc as a known limitation rather than leaving it implicit.

artifactory/commands/apt/command.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	if nativeTool == "apt-get" && !c.skipLogin {
+		switch {
+		case c.serverDetails != nil && c.repoName != "" && c.dist != "":
+			tmpPath, err := WriteTempSourcesList(c.serverDetails, c.repoName, c.dist, c.component, c.trusted)
+			if err != nil {
+				log.Warn("Failed to create temporary sources.list — proceeding without auth injection: " + err.Error())
+			} else {
+				defer func() { _ = os.Remove(tmpPath) }()
```
agrasth, line 183: **MINOR -** 
This cleanup relies on `defer`, so if the process is killed (OOM, SIGKILL, CI job cancellation) between `WriteTempSourcesList` and the deferred `os.Remove`, the temp file — which embeds the user's credentials in plain text — leaks into the OS temp dir until something else cleans it up. 
Given the file is already given a distinctive `jfrog-apt-sources-*` prefix, it might be worth a best-effort sweep of stale leftovers (e.g. by mtime) at the start of a run, similar to how `cleanAptTest` does for the test suite. 

Not blocking, but worth a conscious decision either way given secrets are involved.

agrasth (dismissed):

artifactory/commands/apt/auth.go
fluxxBot: this is what set me up suggests, so keeping the current implementation

artifactory/commands/apt/command.go
fluxxBot: whole idea is to resolve via artifactory

artifactory/commands/apt/setup.go
fluxxBot: why do you think document is required, whole idea to use artifactory without touching user's config

agrasth (dismissed):

bhanurp (dismissed):

bhanurp (approved):

udaykb2 (approved):

PR #508, opened 2026-07-14
Rteco 1567 - agent update command improvement

udaykb2:
-----

### DIFF
--- a/agent/common/resolve_repo.go
+++ b/agent/common/resolve_repo.go
@@ -59,7 +59,7 @@ func ResolveRepo(serverDetails *config.ServerDetails, flagValue string, quiet bo
 	}
 	selected := ioutils.AskFromListWithMismatchConfirmation(
 		fmt.Sprintf("Select a %s repository:", opts.Label),
-		"'%s' is not in the list of discovered repos.",
+		"Not in the list of discovered repos. Continue anyway?",
 		options,
 	)
 	return selected, nil
--- a/agent/plugins/commands/publish/publish.go
+++ b/agent/plugins/commands/publish/publish.go
@@ -23,6 +23,9 @@ import (
 // packageVersionExists checks whether a version folder exists in Artifactory. Tests may replace it.
 var packageVersionExists = common.PackageVersionExists
 
+// listPluginVersionsFunc lists all versions of a plugin. Tests may replace it.
+var listPluginVersionsFunc = common.ListPluginVersions
+
 type PublishCommand struct {
 	serverDetails      *config.ServerDetails
 	repoKey            string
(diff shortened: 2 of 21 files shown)


agent/plugins/common/hooks_claude.go
```diff
@@ -54,6 +47,32 @@ func claudePostInstall(slug, version, installDir, repoKey string) error {
...
+	marketplacePath := claudeMarketplacePath(installDir)
+	log.Info(fmt.Sprintf("[claude] refreshing marketplace entry for '%s' -> %s (v%s)", slug, marketplacePath, version))
+	if err := upsertLocalMarketplaceEntry(marketplacePath, slug, version, repoKey); err != nil {
+		return err
+	}
+	_, err := LookPathClaude()
+	if err == nil {
+		// CLI found, proceed with marketplace refresh + plugin update.
+		log.Info(fmt.Sprintf("[claude] refreshing marketplace: claude plugin marketplace update %s", repoKey))
+		ClaudeExec("plugin", "marketplace", "update", repoKey)
```
bhanurp, line 64: Major 🟠 — Error Handling: `ClaudeExec(...)` return value is discarded here (and again on line 67), and the function unconditionally `return nil`s afterward. If the native `claude plugin marketplace update`/`claude plugin update` command fails, the update is still reported as successful. Compare with `claudePostInstall` above, which wraps and returns `execErr`. Suggested fix: capture the error from both `ClaudeExec` calls and return a wrapped error (or at least log.Warn + propagate) instead of silently continuing.

agent/plugins/common/hooks_codex.go
```diff
@@ -98,6 +91,31 @@ func codexPostInstall(slug, version, installDir, repoKey string) error {
...
+	manifestPath := codexMarketplaceManifestPath(installDir)
+	log.Info(fmt.Sprintf("[codex] refreshing marketplace entry for '%s' -> %s (v%s)", slug, manifestPath, version))
+	if err := upsertCodexMarketplaceEntry(manifestPath, slug, repoKey); err != nil {
+		return err
+	}
+	_, err := LookPathCodex()
+	if err == nil {
+		// CLI found, proceed with marketplace refresh + plugin resync.
+		log.Info(fmt.Sprintf("[codex] refreshing marketplace: codex plugin marketplace upgrade %s", repoKey))
+		CodexExec("plugin", "marketplace", "upgrade", repoKey)
```
bhanurp, line 108: Major 🟠 — Error Handling: same pattern as `claudePostUpdate` — `CodexExec(...)` return value discarded here and on line 110, then unconditional `return nil`. A failing `codex plugin marketplace upgrade`/`codex plugin add` will silently be reported as a successful update. Suggested fix: capture and return/wrap the error, matching how `codexPostInstall`'s `CodexExec("plugin", "add", ...)` call on line 81 already does (`if execErr := CodexExec(...); execErr != nil { return ... }`).

agent/common/resolve_repo.go
```diff
@@ -59,7 +59,7 @@ func ResolveRepo(serverDetails *config.ServerDetails, flagValue string, quiet bo
 	}
 	selected := ioutils.AskFromListWithMismatchConfirmation(
 		fmt.Sprintf("Select a %s repository:", opts.Label),
-		"'%s' is not in the list of discovered repos.",
+		"Not in the list of discovered repos. Continue anyway?",
```
bhanurp, line 62: Minor 🔵 — Functionality: `ioutils.AskFromListWithMismatchConfirmation` in jfrog-cli-core appends its own `" continue anyway?"` suffix to this message unconditionally, so the final prompt reads "Not in the list of discovered repos. Continue anyway? continue anyway?" — duplicated question. Suggest ending this string with a period instead of a question, e.g. `"Not in the list of discovered repos."`.

agent/plugins/commands/publish/publish.go
```diff
@@ -154,6 +174,51 @@ func (pc *PublishCommand) Run() error {
...
+			}
+			return result, nil
+		},
+	})
+}
+
+// promptForVersion asks the user to enter a version when --version was not provided and prompts
+// are safe (interactive terminal, not --quiet/CI). manifestVersion, if any, is shown for reference
+// only; the user must type a value.
+func (pc *PublishCommand) promptForVersion(manifestVersion string) (string, error) {
```
bhanurp, line 203: Minor 🔵 — Maintainability: `promptForVersion` is no longer called from `Run()`/`resolveMissingVersion` (which now uses `common.ResolveMissingVersion`/`common.PromptLine` instead). Its only remaining callers are its own unit tests in publish_test.go. Recommend deleting this method and its dedicated tests to avoid confusing it with the live version-prompt path.

PR #509, opened 2026-07-16
RTECO-1644 - Add pnpm support: update workflow add version tracking, and improve build info metadata

fluxxBot:
-----

### DIFF
--- a/.github/workflows/cli-integration-tests.yml
+++ b/.github/workflows/cli-integration-tests.yml
@@ -5,12 +5,6 @@ on:
   push:
     branches:
       - main
-      # TEMP: pull_request_target below always runs the workflow file from the
-      # base branch (main), so edits to this file on a PR branch are invisible
-      # until merged. Adding this branch to `push` lets pushing here run CI
-      # using THIS branch's own copy of the workflow. Remove once the
-      # feature/RTECO-1644-pnpm11 PR's CI has passed.
-      - RTECO-1644-pnpm11
 
   pull_request_target:
     types: [labeled]
@@ -298,7 +292,7 @@ jobs:
         run: go test -v github.com/jfrog/jfrog-cli --timeout 0 --test.npm
 
   pnpm-Tests:
-    name: pnpm ${{ matrix.pnpm-version }} ${{ matrix.os.name }}
+    name: pnpm ${{ matrix.pnpm.version }} ${{ matrix.os.name }}
     needs: Detect-Deps
     if: github.event_name == 'workflow_dispatch' || github.event_name == 'push' || contains(github.event.pull_request.labels.*.name, 'safe to test')
     strategy:
@@ -309,7 +303,12 @@ jobs:
             version: 24.04
           - name: windows
             version: 2022
-        pnpm-version: [10, 11]
+        pnpm:
+          - version: 10
+            node: "20"
+          - version: 11
+            # pnpm 11 requires Node >=22.13 (pnpm 10 only requires >=18.12).
+            node: "22.13"
     runs-on: ${{ matrix.os.name }}-${{ matrix.os.version }}
     env:
       JFROG_CLI_LOG_LEVEL: DEBUG
@@ -333,13 +332,12 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v5
         with:
-          # pnpm 11 requires Node >=22.13 (pnpm 10 only requires >=18.12, so 22 covers both legs).
-          node-version: "22.13"
+          node-version: ${{ matrix.pnpm.node }}
 
       - name: Install pnpm
         uses: pnpm/action-setup@v4
         with:
-          version: ${{ matrix.pnpm-version }}
+          version: ${{ matrix.pnpm.version }}
 
       - name: Setup Go with cache
         uses: jfrog/.github/actions/install-go-with-cache@main
--- a/.github/workflows/cli-integration-tests.yml
+++ b/.github/workflows/cli-integration-tests.yml
@@ -5,6 +5,12 @@ on:
   push:
     branches:
       - main
+      # TEMP: pull_request_target below always runs the workflow file from the
+      # base branch (main), so edits to this file on a PR branch are invisible
+      # until merged. Adding this branch to `push` lets pushing here run CI
+      # using THIS branch's own copy of the workflow. Remove once the
+      # feature/RTECO-1644-pnpm11 PR's CI has passed.
+      - RTECO-1644-pnpm11
 
   pull_request_target:
     types: [labeled]
@@ -303,7 +309,7 @@ jobs:
             version: 24.04
           - name: windows
             version: 2022
-        pnpm-version: [10]
+        pnpm-version: [10, 11]
     runs-on: ${{ matrix.os.name }}-${{ matrix.os.version }}
     env:
       JFROG_CLI_LOG_LEVEL: DEBUG
@@ -327,7 +333,8 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v5
         with:
-          node-version: "20"
+          # pnpm 11 requires Node >=22.13 (pnpm 10 only requires >=18.12, so 22 covers both legs).
+          node-version: "22.13"
 
       - name: Install pnpm
         uses: pnpm/action-setup@v4
--- a/artifactory/commands/pnpm/utils.go
+++ b/artifactory/commands/pnpm/utils.go
@@ -25,7 +25,7 @@ import (
 func recordPnpmCommandMetadata(pnpmBuild *build.Build, pnpmVersion *version.Version, commandArgs []string) {
 	env := buildCommandMetadataEnv(pnpmVersion, commandArgs)
 	if err := pnpmBuild.SavePartialBuildInfo(&entities.Partial{Env: env}); err != nil {
-		log.Debug("Failed to save pnpm command metadata:", err.Error())
+		log.Warn("Failed to save pnpm command metadata:", err.Error())
 	}
 }
 
--- a/artifactory/commands/pnpm/utils.go
+++ b/artifactory/commands/pnpm/utils.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/jfrog/build-info-go/build"
 	"github.com/jfrog/build-info-go/entities"
 	"github.com/jfrog/gofrog/version"
 	buildUtils "github.com/jfrog/jfrog-cli-core/v2/common/build"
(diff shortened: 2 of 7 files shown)


artifactory/commands/pnpm/utils.go
```diff
@@ -15,54 +16,72 @@ import (
...
-func validatePnpmPrerequisites() error {
+// validatePnpmPrerequisites checks that pnpm and Node.js meet the version requirements,
+// returning the resolved pnpm version so callers don't need to re-query it.
+func validatePnpmPrerequisites() (*version.Version, error) {
 	pnpmVer, err := getPnpmVersion()
 	if err != nil {
-		return err
+		return nil, err
 	}
 	if pnpmVer.Compare(minSupportedPnpmVersion) > 0 {
```
bhanurp, line 73: **Major 🟠 — Functionality/API Contract:** The upper-bound pnpm version check (`firstUnsupportedPnpmVersion`) was removed entirely, not just raised to allow 11.x. Any future pnpm major version (12.x, 13.x…) now silently passes this check and will only fail later, and confusingly, if it changes the JSON output shapes this package parses (`pnpm ls/publish/pack --json`).

**Suggested fix:** keep a deliberate upper bound (e.g. `<12.0.0`) and raise it only when a new major version has actually been validated, instead of removing the ceiling.

artifactory/commands/pnpm/utils.go
```diff
@@ -15,54 +16,72 @@ import (
...
 
+// recordPnpmCommandMetadata saves the pnpm version and the full executed pnpm
+// command as build-info properties, mirroring the docker.build.command pattern
+// used by the docker build command (docker_build.go's getBiProperties). Takes
+// the already-resolved pnpm version and build handle so it doesn't re-spawn
+// `pnpm --version` or re-run build prerequisite prep, both already done by the
+// caller's command validation and build-info setup.
+func recordPnpmCommandMetadata(pnpmBuild *build.Build, pnpmVersion *version.Version, commandArgs []string) {
+	env := buildCommandMetadataEnv(pnpmVersion, commandArgs)
+	if err := pnpmBuild.SavePartialBuildInfo(&entities.Partial{Env: env}); err != nil {
```
bhanurp, line 27: **Minor 🔵 — Error Handling:** `SavePartialBuildInfo` failure for the new `pnpm.command`/`pnpm.version` metadata is only logged at Debug level, so at default log levels the metadata this PR adds can silently go missing with no user-visible signal. Mirrors the existing docker pattern, but this is a new feature—consider `log.Warn` instead.

artifactory/commands/pnpm/utils.go
```diff
@@ -15,54 +16,72 @@ import (
...
 func NewCommand(cmdName string, args []string, buildConfig *buildUtils.BuildConfiguration, serverDetails *config.ServerDetails) (commands.Command, error) {
-	if err := validatePnpmPrerequisites(); err != nil {
+	pnpmVer, err := validatePnpmPrerequisites()
+	if err != nil {
 		return nil, err
 	}
 	switch cmdName {
 	case "install", "i":
-		return NewPnpmInstallCommand().SetArgs(args).SetBuildConfiguration(buildConfig).SetServerDetails(serverDetails), nil
+		return NewPnpmInstallCommand().SetArgs(args).SetBuildConfiguration(buildConfig).SetServerDetails(serverDetails).SetPnpmVersion(pnpmVer), nil
```
bhanurp, line 58: **Minor 🔵 — Testing:** No test asserts that `NewCommand` actually wires `.SetPnpmVersion(pnpmVer)` onto the returned command, and there's no end-to-end test that a successful install/publish persists `pnpm.command`/`pnpm.version` into the saved build info. Only the pure `buildCommandMetadataEnv` helper is unit-tested.

.github/workflows/cli-integration-tests.yml
```diff
@@ -327,7 +333,8 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v5
         with:
-          node-version: "20"
+          # pnpm 11 requires Node >=22.13 (pnpm 10 only requires >=18.12, so 22 covers both legs).
+          node-version: "22.13"
```
bhanurp: **Minor 🔵 — Testing:** Bumping `node-version` to `22.13` applies to both matrix legs, so the pnpm 10 leg loses its previous Node 20 coverage even though Node 20 is still a supported combination (`minRequiredNodeVersion = 18.12.0`). Consider a per-leg node version so pnpm 10 keeps an older-Node combination covered.

.github/workflows/cli-integration-tests.yml
```diff
@@ -5,6 +5,12 @@ on:
   push:
     branches:
       - main
+      # TEMP: pull_request_target below always runs the workflow file from the
+      # base branch (main), so edits to this file on a PR branch are invisible
+      # until merged. Adding this branch to `push` lets pushing here run CI
+      # using THIS branch's own copy of the workflow. Remove once the
+      # feature/RTECO-1644-pnpm11 PR's CI has passed.
+      - RTECO-1644-pnpm11
```
bhanurp: **Minor 🔵 — Maintainability:** This `push` trigger is explicitly marked `TEMP`/to-be-removed, but nothing enforces that cleanup once this PR merges and the branch is deleted—it'll become permanent dead configuration in `main`'s workflow file unless manually removed. Consider stripping it as part of this merge or filing a tracked follow-up.

.github/workflows/cli-integration-tests.yml
```diff
@@ -5,6 +5,12 @@ on:
   push:
     branches:
       - main
+      # TEMP: pull_request_target below always runs the workflow file from the
+      # base branch (main), so edits to this file on a PR branch are invisible
+      # until merged. Adding this branch to `push` lets pushing here run CI
+      # using THIS branch's own copy of the workflow. Remove once the
+      # feature/RTECO-1644-pnpm11 PR's CI has passed.
```
bhanurp: **Nit 🟢 — Styling:** Comment says "feature/RTECO-1644-pnpm11" but the actual branch referenced below (and in the PR) is `RTECO-1644-pnpm11`, no `feature/` prefix. Could confuse the later cleanup.

bhanurp (changes requested):
.

artifactory/commands/pnpm/utils.go
agrasth: This is intentional. 
We've decided to keep `minSupportedPnpmVersion` as an open floor with no upper ceiling going forward, rather than re-gating on every new pnpm major.

artifactory/commands/pnpm/utils.go
agrasth: Good catch — bumped to `log.Warn` in 8efb4c6 so this isn't silently swallowed at default log levels.

artifactory/commands/pnpm/utils.go
agrasth: Added an assertion in `TestNewCommandUnsupported` (8efb4c6) that type-asserts the returned command and checks `pnpmVersion` is wired for both install and publish. Didn't add a full end-to-end build-info-persistence test since `TestPnpmInstallAndPublishNormalProject`-style tests already exercise the real save path against a live Artifactory and would need extending to assert on the specific `pnpm.command`/`pnpm.version` env keys — happy to follow up with that separately if you'd like it in this PR instead.

.github/workflows/cli-integration-tests.yml
agrasth: Fixed in 8efb4c6 — restructured the matrix to pair pnpm version with Node version (`matrix.pnpm: [{version: 10, node: 20}, {version: 11, node: 22.13}]`) instead of bumping Node for both legs, so the pnpm-10 leg keeps its original Node 20 coverage.

.github/workflows/cli-integration-tests.yml
agrasth: Removed the temp push trigger entirely in 8efb4c6 now that CI has passed on this branch — no more manual cleanup needed.

.github/workflows/cli-integration-tests.yml
agrasth: Moot now — removed the whole TEMP block in 8efb4c6 (see reply above), so this comment/branch reference is gone.

agrasth (dismissed):

artifactory/commands/pnpm/utils.go
```diff
@@ -15,54 +16,72 @@ import (
...
+	}
+	return env
+}
+
 const (
-	minSupportedPnpmVersion     = "10.0.0"
-	firstUnsupportedPnpmVersion = "11.0.0" // exclusive upper bound: 10.x only
-	minRequiredNodeVersion      = "18.12.0"
+	minSupportedPnpmVersion = "10.0.0"
+	minRequiredNodeVersion  = "18.12.0"
```
bhanurp: isn't minimum required node version for pnpm11 Node22? when on pnpm 11 and node 18 are users getting right error messages?

artifactory/commands/pnpm/utils.go
```diff
@@ -15,54 +16,72 @@ import (
...
 )
 
 // NewCommand creates a pnpm command by subcommand name with common fields set.
 func NewCommand(cmdName string, args []string, buildConfig *buildUtils.BuildConfiguration, serverDetails *config.ServerDetails) (commands.Command, error) {
-	if err := validatePnpmPrerequisites(); err != nil {
+	pnpmVer, err := validatePnpmPrerequisites()
```
bhanurp, line 52: Nit: try with unsupported sub command or typo in sub command the error says failed to determine pnpm version ensure pnpm is installed

artifactory/commands/pnpm/utils.go
```diff
@@ -15,54 +16,72 @@ import (
...
+	if err != nil {
 		return nil, err
 	}
 	switch cmdName {
 	case "install", "i":
-		return NewPnpmInstallCommand().SetArgs(args).SetBuildConfiguration(buildConfig).SetServerDetails(serverDetails), nil
+		return NewPnpmInstallCommand().SetArgs(args).SetBuildConfiguration(buildConfig).SetServerDetails(serverDetails).SetPnpmVersion(pnpmVer), nil
 	case "publish":
-		return NewPnpmPublishCommand().SetArgs(args).SetBuildConfiguration(buildConfig).SetServerDetails(serverDetails), nil
+		return NewPnpmPublishCommand().SetArgs(args).SetBuildConfiguration(buildConfig).SetServerDetails(serverDetails).SetPnpmVersion(pnpmVer), nil
```
bhanurp, line 60: aren't both cases same ?
`case "install", "i", "publish":` good enough?

bhanurp (dismissed):

artifactory/commands/pnpm/utils.go
agrasth: Traced this — for an unrecognized subcommand, pnpmCmd's switch/default passes straight through to native pnpm (`runNativePackageManagerCmd`), bypassing `NewCommand/getPnpmVersion` entirely, so that message can't fire here. 
Reproduced the same error with bare pnpm publissh (no jf) — it's pnpm's own auto-deps-check behavior on an unknown command, not something we control.

artifactory/commands/pnpm/utils.go
agrasth: can't merge — install/i build a `PnpmInstallCommand`, publish builds a `PnpmPublishCommand` (different structs, different Run() behavior).

artifactory/commands/pnpm/utils.go
agrasth: Fixed in a4c4e73 — `validatePnpmPrerequisites` now branches the required Node.js version on the resolved pnpm major: pnpm 10.x keeps the existing >=18.12.0 floor, pnpm >=11 requires >=22.13.0 (matches pnpm 11's own dropped support for Node 18-21).

bhanurp (approved):

agrasth (approved):

PR #512, opened 2026-07-19
RTECO-0000 - Validate --repo inside SetupCommand.Run so usage reporting is never skipped

bhanurp:
## Summary
- `jf setup <package-manager> --repo <name>` validated the repo existed in the `jfrog-cli` CLI layer, before ever invoking `commands.ExecWithPackageManager` - the only call site that triggers a usage report for this command. Any invalid `--repo` therefore returned early and silently skipped the usage report for that invocation. Companion fix: jfrog/jfrog-cli#3609.
- This moves the repo-existence check into `SetupCommand.Run()` itself, after `IsSupportedPackageManager` and before the interactive-repo-selection / package-manager-configuration logic. Since `Run()` is always reached via `commands.ExecWithPackageManager`, every return path - including this validation - is now covered by that single usage-report call site. No CLI-side workaround is needed.
- `validateRepoExistsFn` is a package-level function var (same seam pattern as other overridable hooks in this codebase) so existing unit tests, which run `Run()` against a fake `acme.jfrog.io` server, aren't broken by the added network call. `setup_test.go`'s new `TestMain` stubs it to a no-op for all tests; `TestSetupCommand_RepoValidationFailure` restores the real seam to verify `Run()` surfaces a validation error immediately.

## Test plan
- [x] `go build ./artifactory/commands/setup/...`
- [x] `go vet ./artifactory/commands/setup/...`
- [x] `go test ./artifactory/commands/setup/...` - all pass except `TestSetupCommand_UV`, which fails identically on `main` due to an outdated local `uv` binary (unrelated to this change).

### DIFF
--- a/artifactory/commands/setup/setup.go
+++ b/artifactory/commands/setup/setup.go
@@ -143,10 +143,34 @@ func (sc *SetupCommand) SetProjectKey(projectKey string) *SetupCommand {
 	return sc
 }
 
+// validateRepoExistsFn is overridable so tests can stub out the network call.
+var validateRepoExistsFn = validateRepoExists
+
+// validateRepoExists checks that the repository explicitly requested via --repo exists in Artifactory,
+// so that a bad repo name fails fast with a clear error instead of surfacing later as a confusing
+// package-manager-specific failure.
+func validateRepoExists(sc *SetupCommand) error {
+	serviceDetails, err := sc.serverDetails.CreateArtAuthConfig()
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


agrasth (approved):
Clean, well-scoped fix: moving --repo existence validation into SetupCommand.Run() correctly ensures it's covered by the single usage-report call site in commands.Exec, and the test seam/TestMain stub is consistent with existing patterns in this codebase (e.g. createDownloadServiceManagerForPackageZip) with no parallel-test race risk since no tests in this package use t.Parallel(). No significant issues found; approving as-is.

naveenku-jfrog (approved):

PR #514, opened 2026-07-23
RTECO-1647 - Add support for Maven Wrapper in native mode

fluxxBot:
-----

### DIFF
--- a/artifactory/commands/mvn/utils.go
+++ b/artifactory/commands/mvn/utils.go
@@ -34,6 +34,7 @@ type MvnUtils struct {
 	insecureTls               bool
 	disableDeploy             bool
 	outputWriter              io.Writer
+	preferWrapper             bool
 }
 
 func NewMvnUtils() *MvnUtils {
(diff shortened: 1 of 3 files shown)


naveenku-jfrog (approved):

artifactory/commands/mvn/utils.go
```diff
@@ -85,16 +86,54 @@ func (mu *MvnUtils) SetOutputWriter(writer io.Writer) *MvnUtils {
...
+	if err != nil {
+		return "", errorutils.CheckError(err)
+	}
+	if exists {
+		wrapperName := "mvnw"
+		if coreutils.IsWindows() {
+			wrapperName = "mvnw.cmd"
+		}
+		wrapperPath := filepath.Join(projectRoot, wrapperName)
+		if _, statErr := os.Stat(wrapperPath); statErr == nil {
```
agrasth, line 117: `os.Stat` only confirms the wrapper file exists, not that it's executable. If `mvnw` lost its execute bit (e.g. a plain zip/tarball checkout that doesn't preserve permissions, or a file copy tool that resets modes) this returns the path successfully, and the failure only surfaces later as a raw "permission denied" from `cmd.Run()` instead of this package's clearer `errorutils.CheckErrorf` messaging. 

Worth checking the executable bit here (on non-Windows) and folding that into the same "no usable Maven Wrapper found" error for a consistent message? 
Relatedly, `utils_test.go`'s `createWrapperFixture` always creates `.mvn/` and the wrapper script together, so there's no test for ".mvn present but wrapper script missing/non-executable" (partial or corrupted wrapper install) hitting this fallthrough.

agrasth (approved):

PR #518, opened 2026-07-28
RTECO-1648 - Implement jf agent apm command

udaykb2:
Adds jf agent apm install/publish/passthrough, wrapping Microsoft's APM CLI with per-run Artifactory registry authentication (env-var credentials, never written to disk) and build-info collection from apm.lock.yaml/apm.yml. Also fixes jf setup agent-apm to search local (not virtual) repositories for the agentpackages package type, since Artifactory has no remote/virtual support for it.

-----
Implements the core business logic for the `jf agent apm` command in jfrog-cli-artifactory, covering registry authentication injection, registry discovery, native `apm` execution, and build-info collection.

## What is included in this PR

- **ApmCommand dispatcher** (`agent/apm/cli`): routes to `install` / `publish`/ catch-all passthrough, each with full native `apm` stdio passthrough via `SkipFlagParsing` + manual flag extraction (`ExtractApmSubcommandOptions`).
- **Authentication injection (non-destructive, per-process env vars only):**
  - `APM_REGISTRY_TOKEN_<NAME>` (or `USER_`/`PASS_`) injected for every registry name discovered — never written to `~/.apm/config.json` or any file.
  - Respects existing native credentials — skips injection if the caller already exported the var.
  - No `--server-id`/`--repo` on any subcommand, including passthrough — a registry must already be declared, same requirement every other package-manager integration in this CLI already has.
- **Registry discovery** (matched by host, combining `apm.yml`'s `registries:` block with existing `~/.apm/config.json` entries) — errors before `apm` runs at all if nothing is found for the host.
- **`jf setup agent-apm`**: persistent-auth path (`project.AgentApm` case in the shared `jf setup <tool>` command) — writes directly to `~/.apm/config.json`, idempotent
- **Build-info collection:**
  - Dependencies: reads `apm.lock.yaml` for `install` (identical reader, identical resolved-deps shape); filters to `source: registry` only.
  - Scope/requestedBy: one `apm deps why <repo_url> --json` call per dependency — models direct vs. transitive and multi-parent chains;
  - Artifacts: `publish` records the uploaded package (owner parsed from `--package owner/name`, path `{repo}/{owner}/{name}/{name}-{version}.zip`).
- **Checksum resolution, three tiers:** previous build cache → single HTTP HEAD per dependency against `resolved_url` (up to 15 concurrent, replacing an older batched AQL query) → lockfile's own `resolved_hash` as last resort.
- Unit tests covering: registry discovery/env-injection, flag extraction, manifest parsing (including the `default:`-key regression), version parsing, checksum resolution, and the full install/publish dispatch.

## Summary by CodeRabbit

* **New Features**
  * Added `jf agent apm install` and `jf agent apm publish` commands with authentication, registry selection, native APM options, and build-info support.
  * Added authenticated passthrough for other APM commands.
  * Added APM setup support for AgentPackages, Ruby, and Cargo workflows.

* **Improvements**
  * Improved dependency and checksum resolution, credential handling, error reporting, and dry-run behavior.
  * Removed the APM `update` command.

* **Documentation**
  * Added comprehensive command help and AI-oriented descriptions.


agent/apm/commands/publish/publish.go
```diff
@@ -0,0 +1,144 @@
...
+			return args
+		}
+	}
+	for i, arg := range args {
+		if !strings.HasPrefix(arg, "-") {
+			rest := make([]string, 0, len(args)-1)
+			rest = append(rest, args[:i]...)
+			rest = append(rest, args[i+1:]...)
+			return append([]string{"--package", arg}, rest...)
+		}
```
naveenku-jfrog: Changes-Request:
withPackageFlag can mis-promote the value of an apm-native flag (e.g. --zip foo.zip acme/pkg) into --package. Require explicit --package or skip known value-flags; add a test.

agent/apm/common/apmenv.go
```diff
@@ -0,0 +1,379 @@
...
+// exception to "only jf setup agent-apm writes to the real home": apm has no env-var
+// equivalent for this flag (confirmed against apm's own docs), and unlike a registry
+// URL it can't collide across projects — it's a single global switch, not project-scoped
+// state, so enabling it as a side effect of ordinary usage carries none of the
+// cross-project collision risk a persisted registry entry would.
+func ensureExperimentalFlagEnabled(realHome string, existing *apmConfigJSON) error {
+	if existing.Experimental.Registries {
+		return nil
+	}
+	existing.Experimental.Registries = true
```
naveenku-jfrog, line 330: nit: ensureExperimentalFlagEnabled writes real ~/.apm on run commands; reconcile with "only setup writes real home" comments elsewhere.

agent/apm/common/checksums.go
```diff
@@ -0,0 +1,104 @@
...
+	log.Info(fmt.Sprintf("Checksum resolution: %d cached, resolving %d from Artifactory.", len(deps)-len(uncached), len(uncached)))
+
+	if len(uncached) == 0 {
+		return checksumMap, nil
+	}
+
+	headResults := resolveChecksumsByHead(uncached, servicesManager)
+	for _, dep := range uncached {
+		if checksum, ok := headResults[dep.ID]; ok {
+			checksumMap[dep.ID] = checksum
```
naveenku-jfrog: Suggestion: Extract + unit-test the three-tier checksum tier-selection (cache→HEAD→lockfile).

artifactory/commands/repository/template.go
```diff
@@ -131,38 +131,39 @@ const (
...
+	Chef          = "chef"
+	Puppet        = "puppet"
+	Vcs           = "vcs"
+	Alpine        = "alpine"
+	Conda         = "conda"
+	P2            = "p2"
+	Swift         = "swift"
+	Terraform     = "terraform"
+	Cargo         = "cargo"
+	AgentPackages = "agentpackages"
```
naveenku-jfrog: Nit: Add only the AgentPackages line; avoid re-aligning the whole const block (blame churn).

naveenku-jfrog (changes requested):

agent/apm/cli/cli.go
```diff
@@ -0,0 +1,78 @@
...
+		{
+			Name:            "publish",
+			Flags:           flagkit.GetCommandFlags(flagkit.AgentApm),
+			SkipFlagParsing: true,
+			Description:     "Publish an APM package to JFrog Artifactory.",
+			AIDescription:   publish.GetAIDescription(),
+			Action:          publish.RunPublish,
+		},
+		{
+			Name:            "update",
```
reshmifrog: do we really need to support update command for build info collection? can you please give an example here why ?

agent/apm/commands/install/install.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return err
+	}
+	serverDetails, err := agentcommon.GetServerDetails(c)
+	if err != nil {
+		return err
+	}
+
+	cmd := NewApmInstallCommand().
+		SetArgs(opts.RemainingArgs).
```
reshmifrog: what do you mean by remaining args? it seems vague can you please change the name.

agent/apm/commands/update/update.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (c *ApmUpdateCommand) CommandName() string {
+	return apmcommon.CommandNamePrefix + apmSubcommand
+}
+
+func (c *ApmUpdateCommand) ServerDetails() (*config.ServerDetails, error) {
+	return c.serverDetails, nil
+}
+
+// Run wraps "apm update", which re-resolves dependencies to their latest matching refs and, on
```
reshmifrog: so if we have a project and we have already installed deps and then again updated the deps then excuted bp so are we only going to collect the latest latest deps added?
can you please share a build info url after testing this scenerio?

agent/apm/common/apmenv.go
```diff
@@ -0,0 +1,631 @@
...
+
+// generateAccessTokenTimeout bounds the token-generation HTTP call, so a hung Artifactory
+// response can't block the whole install/publish/update command forever.
+const generateAccessTokenTimeout = 30 * time.Second
+
+// ApmBinaryName is the apm executable RunApmCommand always shells out to.
+const ApmBinaryName = "apm"
+
+// HelpFlag is the help flag this package constructs when forwarding to apm.
+const HelpFlag = "--help"
```
reshmifrog: Nit pick: we can combine it right?
const (
generateAccessTokenTimeout = 30 * time.Second
ApmBinaryName = "apm"
)

agent/apm/common/apmenv.go
```diff
@@ -0,0 +1,631 @@
...
+	}
+
+	// Priority 2: Generate token from username/password (secure - no plaintext in config)
+	if serverDetails.User != "" && serverDetails.Password != "" {
+		generatedToken := generateAccessToken(serverDetails)
+		if generatedToken != "" {
+			return base, generatedToken
+		}
+		// Fallback: if token generation fails, fall through to no-token case
+		// (APM CLI may handle auth differently or skip this registry)
```
reshmifrog: i do not see any fallback here?

agent/apm/common/apmenv.go
```diff
@@ -0,0 +1,631 @@
...
+
+	// No token available - return URL only
+	return base, ""
+}
+
+// generateAccessToken calls Artifactory's deprecated token generation API
+// (POST /artifactory/api/security/token, form-urlencoded - the JSON, plural
+// "/tokens" endpoint returns 405) to create an access token from username/
+// password. Returns empty string if generation fails.
+func generateAccessToken(serverDetails *config.ServerDetails) string {
```
reshmifrog: check this i think we already have a functions that created the access token , can you please check.

agent/apm/common/build_info.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+
+	moduleID := buildConfig.GetModule()
+	if moduleID == "" {
+		moduleID = moduleName + ":" + version
+	}
+
+	fileName := packageName + "-" + version + "." + apmPackageFileExtension
+	dirPath := packageName
+	artifactPath := fileName
```
reshmifrog: how are we sure that this is the artifact path ? how you sure this is how it is in artifactory?

agent/apm/common/build_info.go
```diff
@@ -0,0 +1,356 @@
...
+	if owner != "" {
+		dirPath = owner + "/" + packageName
+		artifactPath = dirPath + "/" + fileName
+	}
+
+	artifact := entities.Artifact{
```
reshmifrog, line 213: can you please let me know what all are we considering here as artifacts?
because i do not see any artifactory calls to get the artifacts so are we locally calculating and populating the build-info?

agent/apm/common/dependency_resolver.go
```diff
@@ -0,0 +1,173 @@
...
+
+// depsWhyTimeout bounds a single `apm deps why` subprocess, so a hang can't block forever.
+const depsWhyTimeout = 30 * time.Second
+
+// Dependency scope names. A dependency gets exactly one of these, chosen by finalScope's
+// priority ladder (prod > dev > transitive) - the same mutually-exclusive model pnpm's own
+// resolver in this repo uses (artifactory/commands/pnpm/dependency_resolver.go's addScope),
+// rather than combining them.
+const (
+	apmScopeProd       = "prod"
```
reshmifrog, line 41: why are we explicitly naming the scope of the deps , i think we need to use the native commands scope names

reshmifrog (changes requested):

cliutils/flagkit/flags.go
```diff
@@ -920,6 +926,9 @@ var commandFlags = map[string][]string{
 	SkillsList: {
 		url, user, password, accessToken, serverId, repo, harness, projectDir, agentGlobal, agentFormat, agentLimit, agentSortBy, agentSortOrder, agentCheckUpdates,
 	},
+	AgentApm: {
+		BuildName, BuildNumber, module, Project,
```
reshmifrog: check if we need to support --server-id

reshmifrog (changes requested):

agent/apm/common/apmenv.go
udaykb2: yes, we are using native apm commands to set experimental flags

cliutils/flagkit/flags.go
udaykb2: done

agent/apm/common/build_info.go
udaykb2: Everything that we publish are artifacts.

  ```
  "modules": [
        {
            "type": "apm",
            "id": "custom-publish-mod",
            "artifacts": [
                {
                    "type": "zip",
                    "sha1": "eeb403b4e73f6f2f8f1d0cbe4c0271d26faf2e41",
                    "sha256": "33f7e757bd390b9edb347c8731fafabf85cce58da8aca7dd48b36fe540b245a6",
                    "md5": "db5bde3e4b05c2787f42a47e65ee1cd0",
                    "name": "apm-feat-20260830232902-pub-module-1.0.101.zip",
                    "originalDeploymentRepo": "udaykb-apm-local",
                    "path": "udaykb/apm-feat-20260830232902-pub-module/apm-feat-20260830232902-pub-module-1.0.101.zip"
                }
            ]
        }
```

agent/apm/commands/install/install.go
udaykb2: changed

agent/apm/commands/publish/publish.go
udaykb2: Fixed — removed the positional-promotion logic entirely. --package is now required explicitly; a bare positional value or another flag's argument is never inferred as the package. Covered by TestApmPublishPackageFlagNotMisPromotedFromOtherFlagValues in jfrog-cli.

agent/apm/common/apmenv.go
udaykb2: Checked — no actual contradiction found. ConfigureApmRegistryPersistent's doc comment ("Called only by jf setup apm") is correctly scoped to that one function; it never claims ensureExperimentalFlagEnabled is setup-only, and ensureExperimentalFlagEnabled writing on every run (a safe, non-secret, global toggle) was always intentional.

agent/apm/common/checksums.go
udaykb2: Done — already split into selectCachedAndUncached, resolveChecksumsByHead, applyHeadResultsOrLockfileFallback, each with dedicated table-driven tests in checksums_test.go.

artifactory/commands/repository/template.go
udaykb2: Confirmed — the diff here is a single added line (AgentPackages = "agentpackages"), no re-alignment churn.

agent/apm/cli/cli.go
udaykb2: I have checked other package managers support.
None of them update command is supported. So, I have removed update command support in apm.

Thank you for pointing this issue

agent/apm/common/apmenv.go
udaykb2: Done, combined into a single const (...) block; the now-unused generateAccessTokenTimeout was removed too since the legacy fallback no longer uses a context timeout.

agent/apm/common/apmenv.go
udaykb2: Fixed, generateAccessToken now tries the Access API first (generateAccessTokenViaAccessAPI), falling back to the legacy Artifactory token endpoint only if that fails.

agent/apm/common/apmenv.go
udaykb2: Fixed, now reuses rtUtils.CreateAccessServiceManager + services.CreateTokenParams/CreateAccessToken, the same shared path used elsewhere in this repo, instead of a bespoke implementation.

agent/apm/common/build_info.go
udaykb2: verifyArtifactPathExists now issues a real HTTP HEAD against the assumed path to confirm it before recording build-info, and (per a later coderabbit finding) we now skip recording entirely rather than construct a malformed path when apm.yml has no version.

agent/apm/common/dependency_resolver.go
udaykb2: apm has no native scope vocabulary of its own to reuse (unlike npm, whose prod/dev finalScope borrows). So we used prod/dev/transitive

agent/apm/commands/update/update.go
udaykb2: Thank you for pointing out this scenario, removed update command support

naveenku-jfrog (dismissed):

reshmifrog (approved):

agrasth (approved):

PR #521, opened 2026-07-29
Say which configuration jf setup changed and how widely it applies

sverdlov93:
## Overview

`jf setup` writes **user-level** package manager configuration, so a run inside one project silently changes dependency resolution for every other project that user builds. The only output was:

```
Successfully configured maven to use JFrog repository 'libs-virtual'.
```

That says nothing about which file changed or how far the change reaches. When an unrelated project later fails to resolve, the error comes from the package manager itself and names only the missing artifact — there is nothing to connect it back to a `jf setup` run in a different directory. This came out of a report where configuring Maven against a demo repository broke a different project that resolved against another Artifactory, with no warning that the two were linked.

## Details

Adds a line after the success message naming the configuration that changed and stating that it applies beyond the current directory:

```
Successfully configured maven to use JFrog repository 'libs-virtual'.
This updated your user-level Maven settings (settings.xml), so it applies to every maven project for this user, not only the current directory.
```

Container logins are worded differently on purpose. `docker`, `podman` and `helm` authenticate to the platform rather than redirecting resolution — after `jf setup docker`, an unqualified `docker pull alpine` still goes to Docker Hub — so claiming their projects now resolve through Artifactory would be wrong:

```
Credentials were saved to your Docker credential store for your user account.
```

A package manager with no entry prints nothing rather than something vague.

## Notes

- Deliberately **not** included: reporting what the previous value was (“was pointing at X, now Y”). That is the strongest version and would have prevented the reported incident outright, but it needs per-package-manager read-before-write, so it belongs in its own change.
- Companion change in `jfrog-cli` adds the same scope statement to `jf setup --help`, so the behaviour is discoverable before running the command as well as after.
- Three tests: every supported package manager has a note; container logins never claim resolution while resolution-changing managers always state the scope; an unknown package manager stays silent.
- `TestSetupCommand_Pnpm` fails on this machine both with and without this change (pnpm 11.17.0) — pre-existing and unrelated; every other test in the package passes.

## Summary by CodeRabbit

* **New Features**
  * Setup now explains which user-level configuration area was updated for each supported package manager.
  * Container-based package managers clarify that credentials were saved through the container CLI.
  * Configuration messages distinguish settings that apply globally across projects from directory-specific changes.

### DIFF
--- a/artifactory/commands/setup/setup.go
+++ b/artifactory/commands/setup/setup.go
@@ -33,6 +33,54 @@ import (
 	"golang.org/x/exp/maps"
 )
 
+// packageManagerToConfigLocation names the configuration each package manager is
+// written to, so the command can say what it changed. Every entry is user-level:
+// the settings live in the user's home directory, not in the current project, so
+// they take effect for every project this user builds. Saying so matters because
+// nothing else in the output reveals that the change is not scoped to the
+// directory the command was run in.
+var packageManagerToConfigLocation = map[project.ProjectType]string{
+	project.Npm:    "your user-level npm configuration (.npmrc)",
+	project.Pnpm:   "your user-level pnpm configuration (.npmrc)",
+	project.Yarn:   "your user-level Yarn configuration (.yarnrc)",
+	project.Pip:    "your user-level pip configuration (pip.conf)",
+	project.Pipenv: "your user-level pip configuration (pip.conf)",
+	project.Poetry: "your user-level Poetry configuration",
+	project.Twine:  "your user-level Twine configuration (.pypirc)",
+	project.UV:     "your user-level uv configuration (uv.toml)",
+	project.Nuget:  "your user-level NuGet configuration (NuGet.Config)",
+	project.Dotnet: "your user-level NuGet configuration (NuGet.Config)",
+	project.Go:     "your Go environment (GOPROXY)",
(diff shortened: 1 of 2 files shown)


sverdlov93:
Companion help-text change: https://github.com/jfrog/jfrog-cli/pull/3627 — states the same scope in `jf setup --help`, so it is visible before running the command as well as in the output afterwards.

Worth noting `GetAIDescription()` already carried this for agents (*"Configuration is applied globally to the package manager's native config … affecting all projects on the machine, not just the current one"*), added when the command was un-hidden in jfrog/jfrog-cli#3626. These two changes bring the human-facing help and the command output in line with what agents were already told.

sverdlov93:
Folded into #520 so both `jf setup` changes go through a single two-approval cycle rather than two.

The commit is carried over there unchanged, plus a review fix on top: the scope note claimed user-level scope unconditionally, but `PIP_CONFIG_FILE` (pip, pipenv), `NPM_CONFIG_USERCONFIG` (npm, pnpm), `POETRY_CONFIG_DIR` and `UV_CONFIG_FILE` can move the configuration off its user-level default — possibly into the current project — which made the claim wrong in exactly those cases. Those runs now report the actual path and the variable responsible instead.

Nothing is lost by closing this; #520 is the single PR to review.

PR #27, opened 2024-11-14
Adding support in artifactory API to get package lead file

oshratZairi:

PR #523, opened 2026-07-29
Harden jf setup credential files to 0600 (pip, uv, Maven, Gradle, Go, Yarn, pnpm)

igalsi-t:
## Overview

Several `jf setup <pm>` commands write per-user config files that embed credentials (a token in a URL, an `_authToken` line, a `<password>` element) and previously left them world-readable (`0644`). This PR restricts every such file to owner-only (`0600`) after setup.

It started as the pip fix and now completes the credential-file hardening across the package managers that were world-readable.

## What changed

Two small helpers in `artifactory/utils/permissions` centralize this:

- **`WriteFileOwnerOnly`** — for files `jf setup` writes itself (`pip.conf`, `uv.toml`): creates them `0600`, and tightens a pre-existing `0644` file *before* writing so new credentials never land in a world-readable file. Failures are returned (our own write is not best-effort).
- **`RestrictExisting`** — best-effort hardening for files other modules write (`settings.xml`, the Gradle init script, the Go env file, `~/.yarnrc`, pnpm's config): it warns and continues if a file can't be located or tightened, so a hardening step never fails an already-configured setup.

Both are Unix-only; on Windows `os.Chmod` cannot express an owner-only DACL, so they are a documented no-op there.

| Package manager | File | Before | After |
|---|---|---|---|
| pip / pipenv | `pip.conf` / `pip.ini` | 0644 | **0600** |
| uv | `uv.toml` | 0644 | **0600** |
| Maven | `~/.m2/settings.xml` | 0644 | **0600** |
| Gradle | `init.d/jfrog.init.gradle` | 0644 | **0600** |
| Go | Go env file (`go env GOENV`) | 0644 | **0600** |
| Yarn | `~/.yarnrc` | 0644 | **0600** |
| pnpm | `auth.ini` (or the file pnpm reports as `globalconfig`) | 0644 | **0600** |

### Already safe (verified, left untouched)

- **npm** `~/.npmrc` — npm writes it `0600` itself.
- **NuGet/.NET** `NuGet.Config` — the dotnet client writes it `0600`.
- **Docker / Podman / Helm** — use the OS credential store, not a plaintext file we write (out of scope).

Each hardened manager gets a Unix-only `0600` test assertion (skipped on Windows, like pip/uv).

## Behavior change

`0600` intentionally removes group/other read. Single-user setups are unaffected — the owner's own package-manager tooling keeps working. However, a workflow where a **different uid** reads these files may now get *permission denied*: e.g. an image built as `root` and run as a non-root user, a shared CI cache, or a service account reading `settings.xml`. This is the intended effect of hardening a per-user credential file, not a regression — flagged here so a "permission denied" report isn't misread as one.

## #520 leftovers folded in

- Renamed the shadowing `packageManagerConfig` local to `cfg` in `configScopeNote`.
- Cross-referenced `FallbackOnlyIfNotFound` from `GoProxyUrlParams.Direct`'s doc comment.
- Shared a `UV_CONFIG_FILE` constant between `setup.go` and `python/uv.go`.

## Follow-up

The `permissions` helper is generic and could be promoted to `jfrog-cli-core`, next to the code that actually writes the Maven/npm/Yarn configs — making write+chmod atomic and benefiting every cli-core consumer. cli-core also has several inline `os.WriteFile(..., 0600)` calls that share the "mode only applies on create" gap fixed here. Tracked separately so it doesn't block this PR.

## Scope

This partially addresses the credential-exposure finding tracked in [jfrog-agent-hooks#27](https://github.jfrog.info/JFROG/jfrog-agent-hooks/issues/27) (DF-01: file permissions). It intentionally does **not** resolve that issue: **DF-02 — cleartext credential *contents*** in `index-url`, `.npmrc` `_authToken`, `settings.xml` `<password>`, etc. — is a separate problem (not a permissions fix) and remains tracked there.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Security**
  * Credential-bearing package manager configuration files are now restricted to owner-only access on supported systems.
  * Existing configuration files are also hardened when updated.
  * Windows permissions continue to be managed through its native ACL system.

* **Bug Fixes**
  * Improved detection of pip, Go, Maven, pnpm, and UV configuration locations across platforms.
  * Setup now handles missing or unresolved configuration files more safely.
  * Poetry configuration errors are reported more consistently.
  * Gradle initialization scripts are created more reliably.

* **Documentation**
  * Clarified Go proxy fallback behavior.

### DIFF
--- a/artifactory/commands/gradle/gradle.go
+++ b/artifactory/commands/gradle/gradle.go
@@ -383,6 +383,22 @@ func GenerateInitScript(config InitScriptAuthConfig) (string, error) {
 // More info on how Gradle invokes these init scripts can be found here:
 // https://docs.gradle.org/current/userguide/init_scripts.html#sec:using_an_init_script
 func WriteInitScript(initScript string) error {
+	jfrogInitScriptPath := GetInitScriptPath()
+	initScriptsDir := filepath.Dir(jfrogInitScriptPath)
+	if err := os.MkdirAll(initScriptsDir, 0755); err != nil { // #nosec G703 -- path sanitized with filepath.Clean
+		return fmt.Errorf("failed to create Gradle init.d directory: %w", err)
+	}
+	if err := os.WriteFile(jfrogInitScriptPath, []byte(initScript), 0644); err != nil { // #nosec G703 -- path sanitized with filepath.Clean
+		return fmt.Errorf("failed to write Gradle init script to %s: %w", jfrogInitScriptPath, err)
+	}
+	return nil
+}
+
+// GetInitScriptPath returns the path jf setup writes the Gradle init script to:
+// $GRADLE_USER_HOME/init.d/jfrog.init.gradle. When GRADLE_USER_HOME is unset it
+// falls back to Java's user.home (so it stays correct in containers where $HOME
+// differs from user.home) and finally to $HOME/.gradle.
+func GetInitScriptPath() string {
 	gradleHome := os.Getenv(UserHomeEnv)
 	if gradleHome == "" {
 		// Try Java's user.home first (fixes container issue where $HOME != user.home)
@@ -396,16 +412,7 @@ func WriteInitScript(initScript string) error {
 	}
 	// Sanitize the path to prevent directory traversal attacks
 	gradleHome = filepath.Clean(gradleHome)
-
-	initScriptsDir := filepath.Clean(filepath.Join(gradleHome, "init.d"))
-	if err := os.MkdirAll(initScriptsDir, 0755); err != nil { // #nosec G703 -- path sanitized with filepath.Clean
-		return fmt.Errorf("failed to create Gradle init.d directory: %w", err)
-	}
-	jfrogInitScriptPath := filepath.Clean(filepath.Join(initScriptsDir, InitScriptName))
-	if err := os.WriteFile(jfrogInitScriptPath, []byte(initScript), 0644); err != nil { // #nosec G703 -- path sanitized with filepath.Clean
-		return fmt.Errorf("failed to write Gradle init script to %s: %w", jfrogInitScriptPath, err)
-	}
-	return nil
+	return filepath.Clean(filepath.Join(gradleHome, "init.d", InitScriptName))
 }
 
 // GetJavaUserHome queries Java for its user.home system property.
--- a/artifactory/commands/gradle/gradle.go
+++ b/artifactory/commands/gradle/gradle.go
@@ -14,6 +14,7 @@ import (
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/generic"
 	artifactoryutils "github.com/jfrog/jfrog-cli-artifactory/artifactory/utils"
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/utils/civcs"
+	"github.com/jfrog/jfrog-cli-artifactory/artifactory/utils/permissions"
 	commandsutils "github.com/jfrog/jfrog-cli-core/v2/artifactory/commands/utils"
 	"github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
 	"github.com/jfrog/jfrog-cli-core/v2/common/build"
@@ -388,7 +389,10 @@ func WriteInitScript(initScript string) error {
 	if err := os.MkdirAll(initScriptsDir, 0755); err != nil { // #nosec G703 -- path sanitized with filepath.Clean
 		return fmt.Errorf("failed to create Gradle init.d directory: %w", err)
 	}
-	if err := os.WriteFile(jfrogInitScriptPath, []byte(initScript), 0644); err != nil { // #nosec G703 -- path sanitized with filepath.Clean
+	// The init script embeds the Artifactory access token, so write it owner-only
+	// up front rather than writing 0644 and hardening afterwards, which would leave
+	// the token briefly world-readable.
+	if err := permissions.WriteFileOwnerOnly(jfrogInitScriptPath, []byte(initScript)); err != nil {
 		return fmt.Errorf("failed to write Gradle init script to %s: %w", jfrogInitScriptPath, err)
 	}
 	return nil
--- a/artifactory/commands/python/pip.go
+++ b/artifactory/commands/python/pip.go
@@ -12,6 +12,8 @@ import (
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/python/dependencies"
 	"github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
+	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 )
 
 type PipCommand struct {
@@ -54,9 +56,47 @@ func CreatePipConfigManually(customPipConfigPath, repoWithCredsUrl string) error
 	if err := os.MkdirAll(filepath.Dir(cleanPath), os.ModePerm); err != nil {
 		return err
 	}
-	// Write the configuration to pip.conf.
+	// Write the configuration to pip.conf with owner-only permissions — the
+	// index-url may embed credentials (user:token@host).
 	configContent := fmt.Sprintf("[global]\nindex-url = %s\n", repoWithCredsUrl)
-	return os.WriteFile(cleanPath, []byte(configContent), 0644)
+	return os.WriteFile(cleanPath, []byte(configContent), 0600)
+}
+
+// ResolvePipConfigPath returns the pip config file path that `jf setup pip`
+// writes. Honors PIP_CONFIG_FILE; otherwise uses the platform default that
+// `pip config set` targets (Windows: %APPDATA%/pip/pip.ini, else ~/.config/pip/pip.conf).
+func ResolvePipConfigPath() (string, error) {
+	if custom := os.Getenv("PIP_CONFIG_FILE"); custom != "" {
+		return filepath.Clean(custom), nil
+	}
+	if coreutils.IsWindows() {
+		appData := filepath.Clean(os.Getenv("APPDATA"))
+		if appData == "" || appData == "." {
+			return "", errorutils.CheckErrorf("APPDATA environment variable not set")
+		}
+		return filepath.Join(appData, "pip", "pip.ini"), nil
+	}
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", errorutils.CheckErrorf("failed to determine home directory: %w", err)
+	}
+	return filepath.Join(homeDir, ".config", "pip", "pip.conf"), nil
+}
+
+// HardenPipConfigPermissions forces the resolved pip config file to 0600 so a
+// cleartext token in index-url is not left world-readable after `pip config set`.
+func HardenPipConfigPermissions() error {
+	confPath, err := ResolvePipConfigPath()
+	if err != nil {
+		return err
+	}
+	if _, err := os.Stat(confPath); err != nil {
+		if os.IsNotExist(err) {
+			return errorutils.CheckErrorf("pip config file missing after setup: %s", confPath)
+		}
+		return errorutils.CheckError(err)
+	}
+	return errorutils.CheckError(os.Chmod(confPath, 0600))
 }
 
 func (pc *PipCommand) CommandName() string {
--- a/artifactory/commands/python/pip.go
+++ b/artifactory/commands/python/pip.go
@@ -67,7 +67,7 @@ func CreatePipConfigManually(customPipConfigPath, repoWithCredsUrl string) error
 	}
 	// WriteFile applies the mode only when it creates the file, so a config left
 	// at 0644 by an earlier run would otherwise stay world-readable.
-	return errorutils.CheckError(os.Chmod(cleanPath, 0600))
+	return chmodOwnerOnly(cleanPath)
 }
 
 // pipWritingToPrefix is the prefix `pip config set` prints when it persists a
@@ -101,6 +101,12 @@ func PipConfigPathFromOutput(output string) string {
 //
 // os.UserConfigDir already implements the first three lines of that table
 // verbatim, so only pip's macOS "if the directory exists" twist is added here.
+//
+// pip's legacy ~/.pip/pip.conf is deliberately not derived: its user-config
+// list is [legacy, new] and `pip config set` edits the last entry, so it always
+// writes the new location. Verified on pip 26.1.2 - with a legacy file present
+// it still reports "Writing to ~/.config/pip/pip.conf". Preferring legacy here
+// would tighten a file that holds no token and miss the one that does.
 //
 // Prefer PipConfigPathFromOutput; this is the fallback for when pip's own
 // report is unavailable.
@@ -141,8 +147,9 @@ func ResolvePipConfigPath() (string, error) {
 // there would break `jf setup pip` on a machine it had just configured
 // correctly. A path pip itself reported must exist, so that stays fail-closed.
 //
-// On Windows os.Chmod only toggles the read-only attribute; the file is instead
-// protected by the per-user ACLs %APPDATA% already carries.
+// Unix only - see chmodOwnerOnly: on Windows the mode cannot be tightened, so
+// pip.ini keeps the ACLs of its parent directory and this function reports
+// success without having protected anything.
 func HardenPipConfigPermissions(reportedPath string) error {
 	confPath := reportedPath
 	derived := confPath == ""
@@ -151,6 +158,7 @@ func HardenPipConfigPermissions(reportedPath string) error {
 		if confPath, err = ResolvePipConfigPath(); err != nil {
 			return err
 		}
+		log.Debug("pip did not report the configuration file it wrote; falling back to the derived path", confPath)
 	}
 	if _, err := os.Stat(confPath); err != nil {
 		if os.IsNotExist(err) {
@@ -163,7 +171,7 @@ func HardenPipConfigPermissions(reportedPath string) error {
 		}
 		return errorutils.CheckError(err)
 	}
-	return errorutils.CheckError(os.Chmod(confPath, 0600))
+	return chmodOwnerOnly(confPath)
 }
 
 func (pc *PipCommand) CommandName() string {
--- a/artifactory/commands/python/pip.go
+++ b/artifactory/commands/python/pip.go
@@ -12,6 +12,7 @@ import (
 	"github.com/jfrog/build-info-go/entities"
 	"github.com/jfrog/build-info-go/utils/pythonutils"
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/python/dependencies"
+	"github.com/jfrog/jfrog-cli-artifactory/artifactory/utils/permissions"
 	"github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/coreutils"
@@ -67,7 +68,7 @@ func CreatePipConfigManually(customPipConfigPath, repoWithCredsUrl string) error
 	}
 	// WriteFile applies the mode only when it creates the file, so a config left
 	// at 0644 by an earlier run would otherwise stay world-readable.
-	return chmodOwnerOnly(cleanPath)
+	return permissions.ChmodOwnerOnly(cleanPath)
 }
 
 // pipWritingToPrefix is the prefix `pip config set` prints when it persists a
@@ -171,7 +172,7 @@ func HardenPipConfigPermissions(reportedPath string) error {
 		}
 		return errorutils.CheckError(err)
 	}
-	return chmodOwnerOnly(confPath)
+	return permissions.ChmodOwnerOnly(confPath)
 }
 
 func (pc *PipCommand) CommandName() string {
--- a/artifactory/commands/python/pip.go
+++ b/artifactory/commands/python/pip.go
@@ -7,7 +7,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"strings"
 
 	"github.com/jfrog/build-info-go/entities"
 	"github.com/jfrog/build-info-go/utils/pythonutils"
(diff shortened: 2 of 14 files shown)


sverdlov93:
🤖 *Review comment by Cursor AI*

Reviewed as the root-cause fix for [agent-hooks#27](https://github.jfrog.info/JFROG/jfrog-agent-hooks/issues/27). Right layer, right idea — the hook-side workaround ([agent-hooks#66](https://github.jfrog.info/JFROG/jfrog-agent-hooks/pull/66)) should be closed in favour of this. Findings below, all verified by execution rather than reading.

---

### 🔴 The Windows unit-test job is red, and the reason is that `chmod 0600` does nothing on Windows

The `windows` job of `Tests` fails with **8 assertion failures**, all at the lines this PR adds ([run 30444094609](https://github.com/jfrog/jfrog-cli-artifactory/actions/runs/30444094609/job/90550083297)):

```
pip_test.go:32    TestCreatePipConfigManually            expected: 0x180
pip_test.go:55    TestHardenPipConfigPermissions         expected: 0x180
setup_test.go:216 TestSetupCommand_Pip/*  (×6 subtests)  expected: 0x180
```

`0x180` is `0600`. On Windows, Go's `os.Chmod` only toggles `FILE_ATTRIBUTE_READONLY`, and `FileInfo.Mode().Perm()` returns `0666` (or `0444` if read-only) — never `0600`. So the assertion cannot pass there.

This is not just a test problem. It means **the security fix does not protect Windows users at all**: `HardenPipConfigPermissions` returns `nil` on Windows while `pip.ini` stays readable by every other principal on the box per its inherited NTFS ACLs. The `coreutils.IsWindows()` branch in `ResolvePipConfigPath` makes the code read as though Windows is covered.

Please make the intent explicit — either restrict both the chmod and the assertions to non-Windows (`runtime.GOOS != "windows"`), or set a real DACL on Windows. Silently returning success is the one option that shouldn't ship.

---

### 🟠 `ResolvePipConfigPath` re-derives a path that pip does not always use → `jf setup pip` hard-fails, *after* writing the token at 0644

`ResolvePipConfigPath` hard-codes `~/.config/pip/pip.conf` for every non-Windows platform. pip's actual user-config resolution is different in two reachable cases:

**1. macOS — verified empirically on this machine (pip 26.1.2):**

```
$ mkdir -p "$HOME/Library/Application Support/pip"
$ pip config set global.index-url https://example.invalid/simple
Writing to /…/Library/Application Support/pip/pip.conf      # ← not ~/.config
$ stat -f '%OLp' "…/Library/Application Support/pip/pip.conf"
644
```

Remove that directory and the same command writes `~/.config/pip/pip.conf`. This is deliberate in pip — `pip/_internal/utils/appdirs.py`:

```python
def _macos_user_config_dir(appname, roaming=True):
    # Use ~/Application Support/pip, if the directory exists.
    path = _appdirs.user_data_dir(appname, appauthor=False, roaming=roaming)
    if os.path.isdir(path):
        return path
    # Use a Linux-like ~/.config/pip, by default.
```

**2. Linux — `XDG_CONFIG_HOME`.** pip's vendored `platformdirs/unix.py` reads it directly, and [pip's docs](https://pip.pypa.io/en/stable/topics/configuration/) state the Unix path "respects the `XDG_CONFIG_HOME` environment variable". Neither is honoured here.

In both cases the sequence is: `pip config set` succeeds and writes the credential-bearing file at `0644` → `HardenPipConfigPermissions` stats a path that doesn't exist → `errorutils.CheckErrorf("pip config file missing after setup: %s")` → `configurePip` returns an error → **`jf setup pip` fails a command that worked before this PR, and the cleartext token is on disk at 0644 anyway.** Strictly worse than the bug being fixed.

Don't re-derive the path — pip already tells you. `pip config set` prints `Writing to <path>` on stdout, and this package already has the idiom (`poetry.go:348` uses `gofrogcmd.RunCmdWithOutputParser`). Suggested shape:

```go
// RunConfigCommand → capture stdout, parse the "Writing to <path>" line,
// fall back to ResolvePipConfigPath() only if parsing fails.
```

---

### 🟠 `CreatePipConfigManually`: `WriteFile(…, 0600)` is a no-op when the file already exists

`os.WriteFile` applies `perm` only on **creation** — for an existing file it opens `O_WRONLY|O_CREATE|O_TRUNC` and the mode argument is ignored. Verified:

```
before=0644  after WriteFile(...,0600)=0644
```

So for the `PIP_CONFIG_FILE` branch, every user who already ran `jf setup pip` on a pre-fix CLI keeps a world-readable file forever — and that branch `return nil`s before ever reaching `HardenPipConfigPermissions`. Since the hook re-runs `jf setup` each session and on hourly JWT renewal, "the file already exists" is the *normal* state, not the edge case.

This repo already has the correct pattern one directory over — `pypirc.go:48`:

```go
return os.Chmod(pypircPath, 0600)
```

Add the same `os.Chmod` after the write (or route this branch through `HardenPipConfigPermissions` too). The PR description says it "matches the existing Twine `.pypirc` pattern", but Twine chmods and this doesn't. The new assertion in `TestCreatePipConfigManually` passes only because `t.TempDir()` guarantees the file is absent, so the regression is untested.

---

### 🟡 Fail-closed on a *hardening* step converts a permissions concern into an availability outage

`configurePip` now returns an error if the chmod can't be applied. Combined with the path-derivation gap above, a machine that is otherwise perfectly configured cannot complete `jf setup pip` at all — and in the eager-setup hook path that means a `failed` receipt every session, so the agent is told PyPI is unconfigured while pip is in fact correctly pointed at Artifactory.

`log.Warn` + continue is the proportionate behaviour for "I configured you correctly but couldn't tighten the mode". If you want a hard failure, it should be gated on actually having found the file.

---

### 🔵 Dead condition

```go
appData := filepath.Clean(os.Getenv("APPDATA"))
if appData == "" || appData == "." {
```

`filepath.Clean("")` returns `"."`, so `appData == ""` is unreachable. Drop the first clause.

---

### Scope: this does not close agent-hooks#27

#27 covers two things: pip.conf permissions **and** DF-02 — plaintext tokens in `.npmrc` (`_authToken`), `settings.xml` (`<password>`), and embedded in the `index-url` itself. This PR fixes the first, on Unix only, and its own notes acknowledge it "does not remove cleartext credentials from the URL". Worth stating in the description that #27 stays open with reduced scope, so it isn't auto-closed on merge.

Adjacent, for a follow-up rather than this PR: `uv.go:321` writes `uv.toml` with `os.WriteFile(…, 0600)` — same "no-op on an existing file" shape as the finding above.

---

### `/is-all-tested`

| Behaviour introduced | Covered? |
|---|---|
| `CreatePipConfigManually` creates at 0600 | ✅ `TestCreatePipConfigManually` |
| `CreatePipConfigManually` **hardens an existing 0644 file** | ❌ untested — and currently broken |
| `ResolvePipConfigPath` honours `PIP_CONFIG_FILE` | ✅ |
| `ResolvePipConfigPath` Windows `%APPDATA%` branch | ❌ no test |
| `ResolvePipConfigPath` unset-`APPDATA` error | ❌ no test |
| `ResolvePipConfigPath` macOS `Library/Application Support` variant | ❌ no test — and unimplemented |
| `ResolvePipConfigPath` `XDG_CONFIG_HOME` | ❌ no test — and unimplemented |
| `HardenPipConfigPermissions` happy path / missing file | ✅ |
| `HardenPipConfigPermissions` stat error that isn't `IsNotExist` | ❌ no test |
| `configurePip` propagates a harden failure | ❌ no test asserting `jf setup pip` fails |
| End-to-end mode after `jf setup pip` | ✅ `setup_test.go:216` — but red on Windows |

Every `ResolvePipConfigPath` test sets `PIP_CONFIG_FILE`, which short-circuits the derivation on line 1 — so the platform-specific logic that carries the actual risk has **zero** coverage. That's the gap that let the macOS and XDG cases through.

---

### Cross-repo / cross-PR

- **agent-hooks#66** implements the same chmod in the hook's eager-setup worker, explicitly as a stopgap "until the CLI fix ships". This PR is that fix. Recommending #66 be closed — commented there separately.
- No other open PR in either repo touches pip config permissions; `git grep` finds no other caller of `CreatePipConfigManually` outside `setup.go`.

sverdlov93:
🤖 *Review comment by Cursor AI*

Opened **igalsi-t/jfrog-cli-artifactory#2** against this branch with fixes for the two findings above that `54ccf04` did not cover — so it stays your change and you can take or drop it.

- **Path derivation** — now taken from pip's own `Writing to <path>` output, with the derivation as fallback. Confirmed against pip 26.1.2: with `~/Library/Application Support/pip` present, `pip config set` writes there, not `~/.config`, which made the old `os.Stat` miss and fail `jf setup pip` *after* the token was already on disk at 0644.
- **`os.WriteFile(…, 0600)` no-op on an existing file** — `before=0644 after=0644`. Now chmods after writing, as `pypirc.go:48` already does.
- **Deduplication** — the derivation delegates to `os.UserConfigDir`, which implements pip's Windows and Unix rules exactly, leaving only pip's macOS conditional. That removes the hand-rolled `%APPDATA%`/XDG branches instead of adding a third copy next to `getUserUVConfigPath`. The unreachable `appData == ""` check goes away with it (`filepath.Clean("")` returns `"."`).
- **Third copy** — `globalGlobalPipConfigPath` in `setup_test.go` re-derived the same path and would back up the wrong file on those hosts.

On `54ccf04`: gating the assertions is right, but `HardenPipConfigPermissions` still returns `nil` on Windows, so the code reads as though it hardened something. I left the behaviour alone and added a comment noting `%APPDATA%` is already ACL-restricted to the owner — worth deciding whether that is the claim you want to make.

One deliberate softening flagged in the PR: a *derived*-path miss now warns instead of erroring, while a miss on a path pip itself named stays fail-closed. Reasoning in the PR description; easy to revert if you disagree.

artifactory/commands/python/pip.go
```diff
@@ -54,9 +59,111 @@ func CreatePipConfigManually(customPipConfigPath, repoWithCredsUrl string) error
...
+			if homeErr != nil {
+				return "", errorutils.CheckError(homeErr)
+			}
+			pipDir = filepath.Join(homeDir, ".config", "pip")
+		}
+	}
+	configName := "pip.conf"
+	if coreutils.IsWindows() {
+		configName = "pip.ini"
+	}
```
naveenku-jfrog, line 112: os.UserConfigDir + macOS “pip dir must exist” matches pip. Fallback still ignores legacy ~/.pip/pip.conf (pip’s USER list is [legacy, new]). Mitigated when stdout parsing works; if parsing fails and only legacy exists, you warn and leave 0644. Optional follow-up: if ~/.pip/pip.conf exists, prefer it in the derivation.

artifactory/commands/python/pip.go
```diff
@@ -54,9 +59,111 @@ func CreatePipConfigManually(customPipConfigPath, repoWithCredsUrl string) error
...
+// HardenPipConfigPermissions forces the pip config file to 0600 so a cleartext
+// token in index-url is not left world-readable after `pip config set`.
+//
+// reportedPath is the path pip printed; pass "" to fall back to
+// ResolvePipConfigPath. A *derived* path that turns out not to exist is only
+// warned about: pip's resolution can legitimately differ from ours, and failing
+// there would break `jf setup pip` on a machine it had just configured
+// correctly. A path pip itself reported must exist, so that stays fail-closed.
+//
+// On Windows os.Chmod only toggles the read-only attribute; the file is instead
```
naveenku-jfrog: On Windows, os.Chmod(0600) does not set owner-only ACLs; success is a no-op for the security goal. Comment claims %APPDATA% ACLs — true for typical profiles, not a guarantee on shared/custom dirs. Either keep as “Unix-only harden + documented Windows limitation,” or set a real DACL later. Don’t imply Windows is equally hardened.

artifactory/commands/setup/setup.go
```diff
@@ -234,9 +234,16 @@ func (sc *SetupCommand) configurePip() error {
...
-	if err := python.RunConfigCommand(project.Pip, []string{"set", "global.index-url", repoWithCredsUrl}); err != nil {
+	output, err := python.RunConfigCommand(project.Pip, []string{"set", "global.index-url", repoWithCredsUrl})
+	if err != nil {
 		return fmt.Errorf("failed to configure pip index-url: %w", err)
 	}
+	// pip config set creates the file with umask-derived permissions (often 0644);
+	// harden to 0600 because index-url embeds credentials. pip reports the file it
+	// wrote, which beats re-deriving a path that can disagree with it.
+	if err := python.HardenPipConfigPermissions(python.PipConfigPathFromOutput(output)); err != nil {
+		return fmt.Errorf("failed to harden pip config permissions: %w", err)
```
naveenku-jfrog: If pip ever stops printing Writing to (or locale/logging changes), you fall back to derivation + warn. Acceptable; a debug log when PipConfigPathFromOutput returns "" would help ops.

naveenku-jfrog (dismissed):
LGTM. Some nit-picks are available.

artifactory/commands/python/pip.go
sverdlov93: 🤖 *Review comment by Cursor AI*

Checked this empirically before implementing it, and it should **not** be implemented — pip never writes the legacy file.

pip 26.1.2, legacy file present:

```
$ printf "[global]\nindex-url = https://legacy.example/simple\n" > $HOME/.pip/pip.conf
$ pip config set global.index-url https://written.example/simple
Writing to /…/.config/pip/pip.conf          # ← new location, not legacy
```

`pip config debug` confirms both are in the USER list, but `Configuration._get_parser_to_modify` returns `parsers[-1]`, and `get_configuration_files()` orders USER as `[legacy_config_file, new_config_file]` — so `config set` always edits the new one. Preferring legacy in the derivation would chmod the file that holds **no** token and skip the one that does.

Documented that reasoning on `ResolvePipConfigPath` in igalsi-t/jfrog-cli-artifactory#3 rather than changing behaviour.

artifactory/commands/python/pip.go
sverdlov93: 🤖 *Review comment by Cursor AI*

Agreed, and this is the same point raised in the PR review. Took your first option — Unix-only harden with the limitation documented — in igalsi-t/jfrog-cli-artifactory#3:

- New shared `chmodOwnerOnly` returns early on Windows with a debug log, so the no-op is visible rather than reported as success.
- The doc comment no longer claims `%APPDATA%` ACLs are equivalent; it says the file keeps its inherited ACLs, which are per-user on a default profile but not on a redirected or shared one.
- Same helper now used by `CreatePipConfigManually` and the uv writer, so there is one place to add a real DACL later.

artifactory/commands/setup/setup.go
sverdlov93: 🤖 *Review comment by Cursor AI*

Added in igalsi-t/jfrog-cli-artifactory#3 — `HardenPipConfigPermissions` now emits a debug line naming the derived path whenever pip did not report one, so the fallback is visible in `--log-level DEBUG` before the warn-only path is reached.

sverdlov93:
🤖 *Review comment by Cursor AI*

Round 2 on `a388bd4`. Of my earlier findings, four are fixed and one is half-open; opened **igalsi-t/jfrog-cli-artifactory#3** against this branch for the remainder plus Naveen's three nits.

| Earlier finding | Status |
|---|---|
| 🟠 Re-derived path hard-fails after writing the token at 0644 | ✅ pip's own `Writing to` output is authoritative, derivation is fallback |
| 🟠 `WriteFile(…, 0600)` no-op on an existing file | ✅ chmod after write + `TestCreatePipConfigManually_HardensExistingFile` |
| 🟡 Fail-closed hardening → availability outage | ✅ hard failure now gated on pip naming the path; derived miss warns |
| 🔵 Dead `appData == ""` | ✅ branch replaced by `os.UserConfigDir` |
| 🔴 Windows `chmod` no-op | ⚠️ **half** — the Windows job is green now, but `HardenPipConfigPermissions` still chmods on Windows and returns `nil`, so the code reads as though it hardened something. Same point as Naveen's `pip.go:144`. Fixed in #3. |

**Naveen's nits**, all three answered in-thread:

- **Windows** — took option 1 (Unix-only harden, documented limitation) via a shared `chmodOwnerOnly`.
- **Debug log** — added when pip reports no path.
- **Legacy `~/.pip/pip.conf`** — deliberately not implemented. Verified on pip 26.1.2 that `config set` writes the *new* location even when the legacy file exists (`parsers[-1]` of the `[legacy, new]` USER list), so preferring legacy would tighten the file without the token and miss the one with it. Reasoning is now a comment on `ResolvePipConfigPath`.

**Also in #3:** `uv.go:321` had the same `WriteFile`-mode-only-on-create shape — `uv.toml` carries an index URL with embedded credentials, so a pre-fix run leaves it at 0644 indefinitely. Plus tests for the two branches that had none (non-`IsNotExist` stat error, uv hardening).

**Housekeeping:** igalsi-t/jfrog-cli-artifactory#2 is still open but its tree is byte-identical to `a388bd4` — you cherry-picked it, so it can be closed.

**Scope reminder:** [agent-hooks#27](https://github.jfrog.info/JFROG/jfrog-agent-hooks/issues/27) stays open with reduced scope after this — cleartext tokens remain in `index-url`, `.npmrc`, and `settings.xml`. Worth one line in the description so the issue is not read as closed on merge.

artifactory/commands/setup/setup.go
sverdlov93: 🤖 *Review comment by Cursor AI*

✅ **Fixed on this PR** (commits `Make the Windows permission gap explicit and harden uv.toml` / `b1c01fa`).

`HardenPipConfigPermissions` now emits a debug line naming the derived path whenever pip did not report one:

```go
log.Debug("pip did not report the configuration file it wrote; falling back to the derived path", confPath)
```

Visible at `--log-level DEBUG` before the warn-only path is reached.

artifactory/commands/python/pip.go
sverdlov93: 🤖 *Review comment by Cursor AI*

✅ **Fixed on this PR** (same commits on `fix/pip-conf-0600-permissions`).

Took option 1 — Unix-only harden with the limitation documented:

- New shared `chmodOwnerOnly` in `python.go` returns early on Windows with a debug log, so the no-op is visible rather than reported as success.
- Doc comments no longer claim `%APPDATA%` ACLs are equivalent; they state the file keeps inherited ACLs (per-user on a default profile, not on redirected/shared dirs).
- Same helper used by `CreatePipConfigManually`, `HardenPipConfigPermissions`, and `writeUVConfig`.

artifactory/commands/python/pip.go
sverdlov93: 🤖 *Review comment by Cursor AI*

✅ **Addressed on this PR** — deliberately not implemented; reasoning documented in `ResolvePipConfigPath` on this branch.

Verified on pip 26.1.2: with `~/.pip/pip.conf` present, `pip config set` still reports `Writing to …/.config/pip/pip.conf` — it edits `parsers[-1]`, not legacy. Preferring legacy in the derivation would chmod the file without the token and miss the one that has it.

sverdlov93:
🤖 *Review comment by Cursor AI*

**Correction:** three earlier thread replies (and the round-2 summary) incorrectly referenced `igalsi-t/jfrog-cli-artifactory#3`. That was a mistake — the fixes were pushed **directly onto this PR's branch** (`fix/pip-conf-0600-permissions`, head `b1c01fa`). Corrected replies are on each Naveen thread; those threads are now resolved.

**Status on this PR:**

| Source | Item | Status |
|---|---|---|
| Naveen | Windows `chmod` is a no-op | ✅ `chmodOwnerOnly` — Unix-only, debug log on Windows |
| Naveen | Debug log when pip omits `Writing to` | ✅ `log.Debug` in `HardenPipConfigPermissions` |
| Naveen | Prefer legacy `~/.pip/pip.conf` | ✅ Won't fix — documented why (pip writes new location); thread resolved |
| Review (round 1) | Re-derived path hard-fails after 0644 write | ✅ Fixed in `a388bd4` (pip's own output) |
| Review (round 1) | `WriteFile(0600)` no-op on existing file | ✅ chmod after write (pip + uv) |
| Review (round 1) | Windows tests red | ✅ `assertOwnerOnly` skips Windows |
| Review (round 1) | `uv.toml` same gap | ✅ `writeUVConfig` + test |
| Review (round 1) | agent-hooks#27 fully closed | ❌ Out of scope — cleartext tokens remain; #27 stays open with reduced scope |
| Review (round 1) | `configurePip` e2e harden-failure test | ❌ Not added (needs fake pip on PATH) |

CI on `b1c01fa`: Tests + Static Analysis/Go-Sec green.

sverdlov93 (dismissed):

artifactory/commands/gradle/gradle.go
```diff
@@ -383,6 +383,22 @@ func GenerateInitScript(config InitScriptAuthConfig) (string, error) {
 // More info on how Gradle invokes these init scripts can be found here:
 // https://docs.gradle.org/current/userguide/init_scripts.html#sec:using_an_init_script
 func WriteInitScript(initScript string) error {
+	jfrogInitScriptPath := GetInitScriptPath()
+	initScriptsDir := filepath.Dir(jfrogInitScriptPath)
+	if err := os.MkdirAll(initScriptsDir, 0755); err != nil { // #nosec G703 -- path sanitized with filepath.Clean
+		return fmt.Errorf("failed to create Gradle init.d directory: %w", err)
+	}
+	if err := os.WriteFile(jfrogInitScriptPath, []byte(initScript), 0644); err != nil { // #nosec G703 -- path sanitized with filepath.Clean
```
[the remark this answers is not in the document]
sverdlov93: Fixed in d404f71. \`WriteInitScript\` now writes the token-bearing init script owner-only up front via \`permissions.WriteFileOwnerOnly\` (0600, pre-tightening any stale file) instead of writing 0644 and hardening afterwards, so the TOCTOU window where the token was briefly world-readable is closed. Added unit coverage for the owner-only mode and for tightening a pre-existing 0644 script. Resolving.

artifactory/commands/setup/setup.go
```diff
@@ -407,7 +416,42 @@ func (sc *SetupCommand) configureNpmPnpm() error {
...
+	}
+	globalConfig := strings.TrimSpace(string(out))
+	if globalConfig == "" {
+		return nil
+	}
+	authIniPath := filepath.Join(filepath.Dir(globalConfig), "auth.ini")
+	if _, err := os.Stat(authIniPath); err != nil {
+		// No auth.ini (e.g. anonymous access) means there is no token to protect.
+		return nil
+	}
```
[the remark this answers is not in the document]
sverdlov93: Superseded by the pnpm refactor. The old `hardenPnpmAuthConfig` (which returned nil on any stat error) is gone; pnpm hardening now goes through `pnpmCredentialFiles` + `permissions.RestrictExisting`, which is intentionally best-effort: it warns and continues rather than failing `jf setup` if a credential file can not be inspected/tightened. So a permission/I/O error no longer silently reports success without a signal — the user gets a warning telling them to restrict the file manually. Resolving as no longer applicable.

naveenku-jfrog (approved):

sverdlov93 (approved):

PR #524, opened 2026-07-29
Publish the jf setup package-manager capability table

sverdlov93:
## Overview

Publishes the package-manager table `jf setup` already owns, so the four downstream copies of it can be deleted.

> **Stacked on #520** (base is that branch, not `main`, to avoid duplicating `packageManagerConfigs`). GitHub will retarget to `main` when #520 merges.

## The problem

Four consumers each hand-maintain their own approximation of this table, and none can be checked against this package:

| Consumer | Its copy | Drift guard |
|---|---|---|
| Fly Desktop (TS) | `SUPPORTED_PACKAGE_MANAGERS` — binaries, aliases, version commands, min versions | none |
| Fly client (Go) | `handlers/register.go` — names, aliases, binaries, repo types, config groups | none |
| JFrog agent hooks | `TYPE_TO_PMS` + `PM_BINARIES` + `PMS_SETUP_WITHOUT_CLIENT` | none |
| JFrog skills | the manifest table in `SKILL.md` | none |

Each drifts silently the day a package manager is added here. Two live examples:

- The agent hooks gated **maven and gradle on a PATH lookup**, even though `configureMaven` goes through `NewSettingsXmlManager` and `configureGradle` through `WriteInitScript` — neither needs a client. That broke wrapper-only projects (`./mvnw`, `./gradlew`) until it was patched there.
- Their pip probe tries **`pip3` before `pip`**, while `getExecutable` tries `pip` first — so the two can resolve different interpreters' pip.

Both are the same root cause: the knowledge lives here and is re-derived elsewhere.

## What this adds

`GetCapabilities()` / `GetCapability(pm)` / `MarshalCapabilities()`, built **from the existing maps** — no second table to keep in step.

Entries now carry, beyond `location` / `overrideEnv` / `credentialsOnly`:

| Field | Why |
|---|---|
| `clientBinaries` | Defaults to the package manager's own name. `maven` → `mvn`; `pip` → `pip`, `pip3`, matching `getExecutable`'s order. |
| `versionCmd` | **`nuget` and `dotnet` reject `--version`** and print it via `help`; `go` uses `version`; `helm` uses `version --short`. |
| `minVersion` | Only where a release introduced the mechanism setup depends on. `helm` **3.8.0** — `configureHelm` runs `helm registry login` against an OCI registry, and OCI reached GA in 3.8.0. |
| `requiresClient` (from `configOnly`) | False for maven and gradle, so callers stop gating them on a client that wrapper-only projects never install. |
| `role` (from `publishOnly`) | `twine` is upload configuration, not install-time routing — it is `publish`, and still `pypi`. |
| `configGroup` | Package managers writing the **same file**: `pip`+`pipenv` (`pip.conf`), `nuget`+`dotnet` (`NuGet.Config`). Callers must serialize those setups. npm/pnpm/Yarn each write their own file and are deliberately **not** grouped, as are Docker and Podman. |

`configGroup` is the field none of the copies had except the Fly client, and it is a correctness constraint, not metadata: two concurrent `jf setup` runs in one group race on the same file.

## Sample output

```json
{
  "packageManager": "pip",
  "repoPackageType": "pypi",
  "clientBinaries": ["pip", "pip3"],
  "versionCmd": ["--version"],
  "requiresClient": true,
  "role": "install",
  "credentialsOnly": false,
  "configGroup": "pip-conf",
  "configLocation": "your user-level pip configuration (pip.conf)",
  "overrideEnv": "PIP_CONFIG_FILE"
}
```

## A divergence found while writing the tests

**`gradle` maps to the `gradle` repository package type here, not `maven`.** The Fly client registers gradle under its maven repo name, and the agent hooks map the `maven` package type to `[maven, gradle]`. Since `packageManagerToRepositoryPackageType` drives `promptUserToSelectRepository`, `jf setup gradle` will **not** offer a maven-type repository — so both copies contradict this mapping, and a caller passing a maven repo key to `jf setup gradle` is relying on Artifactory's Maven layout rather than on the type model.

I pinned the current behaviour in a test rather than changing it, because which one is correct is your call. Flagging it because it is exactly the kind of disagreement this table exists to prevent, and it went unnoticed in three repos.

## Tests

`TestGetCapabilities_EveryEntryIsFullyPopulated` is now what actually keeps `packageManagerConfigs` in step with `packageManagerToRepositoryPackageType`. The existing comment claimed a test did that, but the only one (`TestPackageManagerConfigs_OverridesAreExactlyTheVerifiedSet`) covers the override-env subset — I corrected the comment.

Mutation-checked, not just green:

| Mutation | Caught by |
|---|---|
| Add `project.Cocoapods` to the repo-type map with no config entry | `EveryEntryIsFullyPopulated/cocoapods` → *"configLocation must be set"* |
| Flip maven's `configOnly` to false | `KnownSpecialCases/maven_and_gradle_need_no_client_on_PATH` |

Also covered: default application, every special case above, group membership and that no group has a single member, JSON round-trip, and that optional fields stay **absent** rather than emitting empty values a consumer would have to disambiguate.

Full `artifactory/commands/setup` suite passes locally (15s, includes the real per-PM setup tests).

## Not in this PR

- **The `jf setup --list` flag** lives in `jfrog-cli`; `MarshalCapabilities` is the whole implementation, so that side is a thin flag. Happy to open it as a companion once the shape here is agreed — the flag name and JSON keys become a versioned contract, so they are worth settling first.
- **`uv`'s `minVersion` is unset.** `configureUV` needs `uv auth login` and I could not confirm which release introduced it. Fly Desktop asserts `0.8.15`; worth confirming before publishing it as a CLI-level claim.
- Deleting the downstream copies — follow-ups in those repos.

### DIFF
--- a/artifactory/commands/setup/capabilities.go
+++ b/artifactory/commands/setup/capabilities.go
@@ -1,0 +1,145 @@
+package setup
+
+import (
+	"encoding/json"
+
+	"github.com/jfrog/jfrog-cli-core/v2/common/project"
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+)
+
+// Roles a package manager can be configured for.
+const (
+	// RoleInstall means setup redirects dependency resolution through Artifactory.
+	RoleInstall = "install"
+	// RolePublish means setup configures uploads, not resolution.
+	RolePublish = "publish"
+)
+
+// Configuration-file groups. Every package manager in one group writes the SAME
+// file, so a caller setting up more than one of them must serialize those runs.
+const (
+	ConfigGroupPipConf     = "pip-conf"
+	ConfigGroupNugetConfig = "nuget-config"
+)
+
+// defaultVersionArg is what most clients accept to print their version.
+var defaultVersionArg = []string{"--version"}
+
+// Capability is the machine-readable description of one package manager that
+// `jf setup` supports.
+//
+// It exists because every downstream consumer had grown its own copy of this
+// table — Fly Desktop (TypeScript), the Fly client (Go) and the JFrog agent
+// hooks each hand-maintain a package-manager list, binary names, shared-config
+// groupings and which package managers need a client on PATH. None of those
+// copies can be checked against this package, so each one drifts silently the
+// day a package manager is added here. Publishing the table makes those copies
+// deletable.
+type Capability struct {
+	// PackageManager is the token accepted by `jf setup <package manager>`.
+	PackageManager string `json:"packageManager"`
+	// RepoPackageType is the Artifactory repository package type it resolves
+	// against, e.g. "pypi" for pip, pipenv, poetry, twine and uv.
+	RepoPackageType string `json:"repoPackageType"`
+	// Aliases are other accepted names for the same package manager.
+	Aliases []string `json:"aliases,omitempty"`
+	// ClientBinaries are the executables to look for on PATH, in preference order.
+	ClientBinaries []string `json:"clientBinaries"`
+	// VersionCmd is the argument list that makes the client print its version.
+	// Not every client accepts --version.
+	VersionCmd []string `json:"versionCmd"`
+	// MinVersion, when set, is the oldest client release setup can configure.
+	MinVersion string `json:"minVersion,omitempty"`
+	// RequiresClient reports whether setup runs the client. When false the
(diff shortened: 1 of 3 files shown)


sverdlov93:
🤖 For the record: this closed as *merged* because #520's branch was fast-forwarded to include these commits — it was not merged to `main`. Per @michaelsv, this repo takes one PR per change set rather than a stack, so all three commits now ship as part of **#520**. No separate review needed here.

PR #526, opened 2026-08-03
RTECO-1646 - Enhance Maven Native command for maven multi-module workspaces and server-id

fluxxBot:
-----

- Added server details to the Maven command execution to improve build info collection.
- Introduced new utility functions for extracting resolution arguments and sanitizing packaging types.
- Implemented tests for various Maven functionalities including artifact collection and module ID splitting.
- Updated the build-info-go dependency to a newer version for enhanced features.

## Summary by CodeRabbit

- **New Features**
  - Improved Maven build-info collection using invoked Maven arguments and resolution settings.
  - Added support for multi-module projects and module-specific artifact collection.
  - Associated artifacts with effective deployment repositories and verified them using checksums.
  - Maven wrapper detection now works from the current directory or parent directories.
  - JAR-producing packaging types now use `.jar` artifact filenames while retaining their original type.

- **Bug Fixes**
  - Improved validation of Maven coordinates, packaging, repository URLs, and module locations.
  - Enhanced legacy Maven build-mode metadata handling.
  - Improved handling of incomplete resolution arguments and artifact search errors.


go.mod
```diff
@@ -203,3 +203,5 @@ require (
 // replace github.com/gfleury/go-bitbucket-v1 => github.com/gfleury/go-bitbucket-v1 v0.0.0-20230825095122-9bc1711434ab
 
 // replace github.com/ktrysmt/go-bitbucket => github.com/ktrysmt/go-bitbucket v0.9.80
+
+replace github.com/jfrog/build-info-go => github.com/jfrog/build-info-go v1.13.1-0.20260804091857-2e9f7c89c1d5
```
agrasth: **Blocking.** Active `replace` pointing at a specific untagged commit on build-info-go: `2e9f7c89` = the HEAD of the still-open build-info-go PR #408. Merging in this state ties `main` to an unmerged branch; if #408 is rebased or its HEAD changes, this pin breaks.

Must be deleted and the `require` at line 11 bumped to whatever tag/commit is merged into build-info-go `main`, in that order.

artifactory/commands/flexpack/maven.go
```diff
@@ -144,372 +112,199 @@ func wasDeployCommand() bool {
...
-	isSnapshot := isSnapshotVersion(version)
-	log.Debug(fmt.Sprintf("Project version: %s, isSnapshot: %v", version, isSnapshot))
-
-	// Priority 1: Check settings.xml (Maven standard precedence)
-	// settings.xml alt*DeploymentRepository overrides pom.xml in Maven
-	repoKey, err := getRepositoryFromSettings(isSnapshot)
-	if err == nil {
-		log.Debug("Found deploy repository from settings.xml (overriding pom.xml): " + repoKey)
+	// /artifactory/<REPO-KEY>, or just <REPO-KEY> when Artifactory is at the host root.
+	if repoKey := segments[len(segments)-1]; repoKey != "" {
```
agrasth, line 278: The single-segment fallback returns `"artifactory"` when the URL is exactly `https://host/artifactory/` (no repo key).

Example: user's `<distributionManagement>` says `<url>https://acme.jfrog.io/artifactory/</url>` (misconfigured but common). `segments == ["artifactory"]`, the 4-segment check fails, so we fall through and return `"artifactory"` as the repo key. Downstream code then tries to tag artifacts in a repo literally named `artifactory`.

We can add an explicit check: if `repoKey == "artifactory"` (case-insensitive), return the "unable to extract" error.

artifactory/commands/flexpack/maven.go
```diff
@@ -144,372 +112,199 @@ func wasDeployCommand() bool {
...
-func getSettingsXmlPath() string {
-	// Check if -s or --settings flag was used
-	args := os.Args
-	for i, arg := range args {
-		if (arg == "-s" || arg == "--settings") && i+1 < len(args) {
-			return args[i+1]
+	// One AQL + one (parallelized) SetProps per physical repo.
+	buildProps := mavenBuildProperties(buildName, buildNumber, buildArgs.GetProject(), workingDir)
+	total := 0
+	for physicalRepo, sha256s := range sha256sByRepo {
```
agrasth, line 199: If every located module happens to be a `pom` aggregator with zero artifacts (edge case: `mvn deploy` on a parent-only reactor), `sha256sByRepo[physicalRepo]` is populated but empty — `checksumAql(repo, sha256s)` then produces `{"repo":"x","$or":[]}` which Artifactory rejects.

Guard: `if len(sha256s) == 0 { continue }` before calling `tagArtifactsInRepo`.

artifactory/commands/flexpack/maven.go
```diff
@@ -144,372 +112,199 @@ func wasDeployCommand() bool {
...
-
-	if repoUrl != "" {
-		repoKey, err := extractRepoKeyFromUrl(repoUrl)
-		if err == nil {
-			log.Debug("Found deploy repository from pom.xml: " + repoKey)
-			return repoKey, nil
+// checksumAql builds an AQL find body matching any of the given sha256 checksums within repo. sha256 is
+// the collision-resistant content identifier; scoping to the (physical) deploy repo keeps identical
+// content stored elsewhere from being matched, without assuming any repository path layout.
+func checksumAql(repo string, sha256s []string) string {
```
agrasth, line 290: `checksumAql` hand-builds JSON string. If a repo name or checksum ever contained a `"` or `\` character the AQL becomes malformed JSON.

Repo keys and hex sha256s have restricted charsets so this is defensive only, but `json.Marshal` on a `map[string]interface{}` costs nothing and removes the class of bug.

artifactory/commands/flexpack/maven.go
```diff
@@ -87,11 +46,28 @@ func CollectMavenBuildInfoWithFlexPack(workingDir, buildName, buildNumber string
...
+		// each module to its own repo (reactors may deploy modules to different repos); overrideURL, when
+		// set (-DaltDeploymentRepository / settings), applies to all modules.
+		moduleDeployURLs, overrideURL, repoErr := mavenFlex.GetDeploymentRepositories()
+		if repoErr != nil {
+			log.Warn("Failed to resolve Maven deployment repository: " + repoErr.Error())
+		}
+		// Ensure the build's general details (start timestamp) exist before finalizing, so build.timestamp
+		// resolves to the build's real timestamp. finalize runs before the build info is saved, so the
+		// save's own GetOrCreateBuild has not recorded the timestamp yet.
+		if _, buildErr := build.NewBuildInfoService().GetOrCreateBuildWithProject(buildName, buildNumber, buildConfiguration.GetProject()); buildErr != nil {
```
agrasth, line 67: `GetOrCreateBuildWithProject` is called here to seed the timestamp, and again inside `saveMavenFlexPackBuildInfo` (line 92). If the first succeeds and the second fails on a transient error, users get a build-info with a timestamp but no artifacts persisted.

One call at the top of `finalizeDeployedArtifacts` and skip the second in `saveMavenFlexPackBuildInfo` is cleaner.

artifactory/commands/mvn/utils.go
```diff
@@ -226,9 +233,50 @@ func RunMvn(mu *MvnUtils) error {
...
+	updated, err := json.Marshal(raw)
+	if err != nil {
+		log.Debug("Skipping maven build-mode stamp, could not serialize build info: " + err.Error())
+		return
+	}
+	perm := os.FileMode(0644)
+	if info, statErr := os.Stat(buildInfoPath); statErr == nil {
+		perm = info.Mode().Perm()
+	}
+	if err = os.WriteFile(buildInfoPath, updated, perm); err != nil {
```
agrasth: `os.WriteFile` truncates then writes. A reader (another process, or a build-info-collect step) that reads during that window sees a truncated file.

Extractor callers don't concurrently read today, so this is defensive — but use temp-file + `os.Rename` for atomicity. Also note the JSON round-trip strips comments/formatting; if you meant "every field preserved verbatim" as the doc-comment suggests, that's semantically true but not byte-for-byte.

artifactory/commands/flexpack/maven.go
```diff
@@ -144,372 +112,199 @@ func wasDeployCommand() bool {
...
+// deployRepoURL is Maven's EFFECTIVE deployment URL (resolved from effective-pom/effective-settings, so
+// inheritance, interpolation and active profiles are already applied). Its repo key is resolved to the
+// physical repository - a virtual repo becomes its default-deployment local repo, via GetRepository -
+// and artifacts are then matched by sha256 scoped to that repo. That way identical content stored
+// elsewhere is never mis-tagged, and no repository path layout is assumed. build.timestamp comes from
+// the build's real timestamp via buildUtils.CreateBuildProperties.
+//
+// Must run BEFORE the build info is saved so OriginalDeploymentRepo is persisted for `rt bp`.
+func finalizeDeployedArtifacts(workingDir string, buildInfo *entities.BuildInfo, moduleDeployURLs map[string]string, overrideURL, buildName, buildNumber string, buildArgs *buildUtils.BuildConfiguration, serverDetails *config.ServerDetails) error {
+	if overrideURL == "" && len(moduleDeployURLs) == 0 {
```
agrasth, line 130: When `overrideURL == "" && len(moduleDeployURLs) == 0` this warns at `Warn` level. Common case when `mvn install` (not `deploy`) was run but the `wasDeployCommand` check somehow misfires.

Should we drop to `Debug` and return `nil` silently? warning here is noise for non-deploy runs.

artifactory/commands/flexpack/maven.go
```diff
@@ -550,96 +345,147 @@ func validateMavenCoordinate(value string) error {
...
+// they can be replayed on the internal `mvn dependency:tree` call. Forwarded flags:
+//   - -P / --activate-profiles     active profiles
+//   - -D / --define                system property overrides
+//   - -s / --settings              user settings file (value in next token or = form)
+//   - -f / --file                  alternate POM (changes what gets resolved)
+//   - -gs / --global-settings      global settings file
+//   - -o / --offline               offline mode
+//
+// Everything else (goals, deploy-only flags, etc.) is ignored.
+func extractResolutionArgs(userArgs []string) []string {
```
agrasth, line 390: `extractResolutionArgs` doesn't handle `-P` split form (`-P foo` with a space). Maven does accept it, though most users write `-Pprod`. Please confirm the test at `TestExtractResolutionArgs` covers both.

artifactory/commands/flexpack/maven.go
fluxxBot: Fixed. Changed `log.Warn` to `log.Debug` for the no-deploy-URL case. Normal for `mvn install` (no deploy goal), so a warning is noise.

artifactory/commands/flexpack/maven.go
fluxxBot: Fixed. Added `if len(sha256s) == 0 { continue }` before calling `tagArtifactsInRepo`. Pom-only aggregator modules (zero artifacts) no longer send an empty `$or` to Artifactory.

artifactory/commands/flexpack/maven.go
fluxxBot: Fixed. Added `strings.EqualFold(repoKey, "artifactory")` check: a URL that resolves to the Artifactory root now returns an explicit error instead of the string `"artifactory"` as a repo key.

artifactory/commands/flexpack/maven.go
fluxxBot: Fixed. `checksumAql` now uses `json.Marshal` on a typed struct with a `$or` tag, eliminating the hand-built JSON string. Repo keys and sha256 hashes have restricted charsets so this was defensive, but the JSON-safe version costs nothing extra.

artifactory/commands/flexpack/maven.go
fluxxBot: Fixed. `-P` alone (space-separated form) now goes through the separate-value branch — the next token is consumed as its value. Only `-P<value>` (len > 2, inline form) stays in the prefix branch. The existing test `TestExtractResolutionArgs` has been extended with a `-P prod` case.

artifactory/commands/mvn/utils.go
fluxxBot: Fixed. `stampMavenBuildMode` now writes to a temp file in the same directory then renames atomically over the target path. Any partial-write window is eliminated.

artifactory/commands/flexpack/maven.go
fluxxBot: The two calls serve different purposes: the one at line 66 seeds the build timestamp BEFORE `finalizeDeployedArtifacts` runs (so `build.timestamp` resolves correctly in the tagged properties), and the one inside `saveMavenFlexPackBuildInfo` ensures the build instance exists before saving. They are idempotent but not redundant — removing the first would cause finalization to tag artifacts with an unset timestamp. Keeping both as-is.

artifactory/commands/flexpack/maven.go
```diff
@@ -144,372 +113,201 @@ func wasDeployCommand() bool {
...
 	}
-
-	var settings SettingsXml
-	if err := xml.Unmarshal(data, &settings); err != nil {
-		return nil, err
-	}
-
-	return &settings, nil
+	return count, nil
 }
```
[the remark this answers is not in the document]
fluxxBot: Fixed. `reader.GetError()` is now called after `SetProps` consumes the reader and before returning, so streaming or parse failures are propagated instead of being silently swallowed.

artifactory/commands/flexpack/maven.go
```diff
@@ -550,96 +348,147 @@ func validateMavenCoordinate(value string) error {
...
+			extracted = append(extracted, arg)
+			// These flags take their value as the next token.
+			if i+1 < len(userArgs) {
+				i++
+				extracted = append(extracted, userArgs[i])
+			}
+		}
 	}
+	return extracted
+}
```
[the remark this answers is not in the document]
fluxxBot: Fixed. `-P` and `-D` exact tokens now go through the separate-value branch (consuming the next token as their value). Inline forms `-Pprofile` and `-Dkey=value` (len > 2) stay in the prefix branch. Tests added for both space-separated forms.

artifactory/commands/flexpack/maven.go
```diff
@@ -550,96 +348,147 @@ func validateMavenCoordinate(value string) error {
...
+			// within targetDir regardless of the (already validated) input. This collapses the value to a
+			// single path element and is the canonical sanitizer for stored path-traversal data flows.
+			mainArtifactName = filepath.Base(mainArtifactName)
+			mainArtifactPath := filepath.Join(targetDir, mainArtifactName)
+
+			if _, statErr := os.Stat(mainArtifactPath); statErr == nil {
+				artifacts = append(artifacts, createArtifactFromFile(mainArtifactPath, groupId, artifactId, version, packagingType))
+			}
+		}
 	}
```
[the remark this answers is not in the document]
fluxxBot: Fixed. Added `packagingFileExtension` which maps `maven-plugin`, `bundle`, `ejb`, `maven-archetype` to `jar` for the deployed filename, while the original `packagingType` is still passed as the artifact's Type field. The `mainArtifactName` now uses the extension from this helper.

artifactory/commands/mvn/utils.go
```diff
@@ -226,9 +233,50 @@ func RunMvn(mu *MvnUtils) error {
...
+		log.Debug("Skipping maven build-mode stamp, could not write build info: " + err.Error())
```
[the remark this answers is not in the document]
fluxxBot: Fixed (addressed in the same batch as agrasth's comment). `stampMavenBuildMode` now uses temp-file + `os.Rename` for atomic writes.

artifactory/commands/flexpack/maven.go
```diff
@@ -144,372 +112,199 @@ func wasDeployCommand() bool {
...
-				Aql: specutils.Aql{
-					ItemsFind: aqlPattern,
-				},
-			},
-		}
-
-		reader, err := servicesManager.SearchFiles(searchParams)
-		if err != nil {
-			log.Warn(fmt.Sprintf("Failed to search for artifact %s: %s", artifact.Name, err))
+		repoKey, keyErr := extractRepoKeyFromUrl(deployURL)
```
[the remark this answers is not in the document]
fluxxBot: Multi-server validation (checking if a module's `deployURL` targets a different Artifactory instance) is out of scope for this PR. Will track as a follow-up — today's flow assumes all modules deploy to the configured server.

artifactory/commands/mvn/utils.go
```diff
@@ -162,8 +168,9 @@ func RunMvn(mu *MvnUtils) error {
 					return errorutils.CheckError(err)
 				}
 
-				// Use FlexPack to collect Maven build info
-				err = flexpack.CollectMavenBuildInfoWithFlexPack(workingDir, buildName, buildNumber, mu.buildConf)
+				// Use FlexPack to collect Maven build info. The user's goals/flags are forwarded so the
+				// internal dependency resolution matches the profiles/settings the build ran with.
+				err = flexpack.CollectMavenBuildInfoWithFlexPack(workingDir, buildName, buildNumber, mu.buildConf, mu.goals, mu.serverDetails)
```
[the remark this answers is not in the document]
fluxxBot: `mu.serverDetails` is already set before `CollectMavenBuildInfoWithFlexPack` is called — see `cli.go:723-727` where `GetSpecificConfig` resolves the server (with `defaultOrEmpty=true` covering the no-flag case) and passes it into `MvnCommand.SetServerDetails`. No change needed here.

artifactory/commands/flexpack/maven.go
```diff
@@ -144,372 +113,205 @@ func wasDeployCommand() bool {
...
-			return repoKey, nil
-		}
-		log.Debug("Failed to extract repository from pom.xml URL: " + err.Error())
+	type aqlBody struct {
+		Repo string              `json:"repo"`
+		Or   []map[string]string `json:"$or"`
 	}
-
-	return "", fmt.Errorf("no deployment repository found in settings.xml or pom.xml")
+	b, _ := json.Marshal(aqlBody{Repo: repo, Or: orList})
```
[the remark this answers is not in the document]
fluxxBot: Fixed. `json.Marshal` error is now handled: on failure the function falls back to manual string building (same as the original implementation). The error path is unreachable in practice since repo keys and sha256 hashes have restricted charsets, but the linter is satisfied.

bhanurp (approved):

agrasth (approved):

PR #532, opened 2026-08-18
RTECO-1574 - Implementation of Nuget Support for client -

bhanurp:
## Summary

- **Push 403 fix**: `nuget.exe push` always sends `X-NuGet-ApiKey` regardless of credential source; Artifactory rejects access tokens via that header with 403. When targeting a known Artifactory repo (and `-Source`/`-ApiKey`/`-SymbolApiKey` are not explicitly passed), bypass `nuget.exe` and `PUT` directly to `/api/nuget/v2/<repo>/` using Basic Auth.
- **Restore credential injection**: `nuget.exe` (mono) re-embeds credentials into MSBuild `RestoreSources` regardless of source, causing NU1301 on V3 feeds. Now writes a temp `nuget.config` with V3 source URL and `<packageSourceCredentials>` using `ClearTextPassword` (cross-platform). Only `-ConfigFile` is appended for restore so MSBuild reads sources from the config file without re-embedding.
- **Proxy support**: Added `http.ProxyFromEnvironment` to push HTTP transport so `HTTPS_PROXY`/`NO_PROXY` env vars are respected.
- **Flag handling**: `-SkipDuplicate`, `-NoSymbols`, sibling `.snupkg` auto-push, `.slnx` guard for `nuget.exe`.
- **`RequiresServerDetails`**: scoped to commands that actually need credentials (push with `--repo`, restore with `--repo-resolve`).
- **`hasNativeAuthOverride`**: prevents bypass when user explicitly passes `-Source`, `-ApiKey`, or `-SymbolApiKey`.

## Test plan

- [ ] `jf nuget push <pkg.nupkg> --repo <local-repo> --server-id <id>` succeeds (no 403)
- [ ] `jf nuget push <pkg.nupkg> -ApiKey <key> --repo <local-repo>` falls through to nuget.exe (bypass skipped)
- [ ] `jf nuget restore --repo-resolve <virtual-repo> --server-id <id>` succeeds on all project types (packages.config, SDK-style .csproj, .sln)
- [ ] `HTTPS_PROXY` is respected during push
- [ ] `-SkipDuplicate` skips 409 conflict without error
- [ ] Sibling `.snupkg` is pushed alongside `.nupkg`

**Source:** https://jfrog-int.atlassian.net/browse/RTECO-1574


artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+	// .snupkg flat when pushed to the symbol-package endpoint. Using the wrong endpoint
+	// causes Artifactory to rename the file (e.g. snupkg → nupkg), breaking path-based
+	// stamping and downstream artifact lookup.
+	nupkgPushURL := rtURL + "/api/nuget/v2/" + c.repoDeploy + "/"
+	snupkgPushURL := rtURL + "/api/nuget/v2/" + c.repoDeploy + "/symbolpackage"
+	skipDuplicate := hasSkipDuplicate(c.args)
+	noSymbols := hasNoSymbols(c.args)
+
+	//nolint:gosec
+	tlsCfg := &tls.Config{InsecureSkipVerify: c.allowInsecureConnections}
```
agrasth: **Silent security regression.** `InsecureSkipVerify: c.allowInsecureConnections` conflates two very different settings. `allowInsecureConnections` is documented (in `injectCredentialsViaTempConfig` around line 273) as the **NuGet 6.8+ "allow HTTP feeds" flag** — it says nothing about TLS certificate verification. 
Using it here to also disable TLS validation means a user who passes `--allow-insecure-connections` for an `http://` Artifactory *silently stops verifying certs* on any HTTPS endpoint the push touches. 

Example: `jf nuget push --allow-insecure-connections` against a corporate HTTPS Artifactory with a self-signed cert previously failed loudly; now it silently succeeds. Split into two flags (e.g. `--insecure-tls`) or don't touch `InsecureSkipVerify` here.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+	switch sub {
+	case "restore", "install", "update", "build", "add":
+		return true
+	}
+	return false
+}
+
+// isPushCommand returns true for push subcommands.
+func isPushCommand(sub string) bool {
+	return sub == "push" || sub == "nuget push"
```
agrasth, line 895: `isPushCommand(sub)` checks `sub == "push" || sub == "nuget push"`, but `buildCmd` around line 233 does `strings.Fields(c.subCommand)` on the same value. 
Any inconsistency between how the jfrog-cli wrapper (PR #3664) populates `subCommand` (`"nuget push"` as one string vs. `"nuget"` + arg) silently mis-dispatches into the native tool instead of the intended bypass. 
We can add a test that uses the *exact* string PR #3664 will pass in.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+	var params services.VirtualRepositoryBaseParams
+	if err := servicesManager.GetRepository(repoKey, &params); err != nil {
+		log.Warn(fmt.Sprintf("Could not resolve repo type for %q; using as-is for OriginalDeploymentRepo: %v", repoKey, err))
+		return repoKey, nil
+	}
+	if params.Rclass != "virtual" {
+		return repoKey, nil
+	}
+	if params.DefaultDeploymentRepo == "" {
+		log.Warn(fmt.Sprintf("Virtual repo %q has no defaultDeploymentRepo configured; OriginalDeploymentRepo will be the virtual repo key.", repoKey))
```
agrasth: Fail-open on virtual-repo resolution: this logs a `Warn` and returns the original `repoKey`, so `OriginalDeploymentRepo: "nuget-virtual"` ends up in build-info. Downstream tools that look up by that key will 404. 
We can consider making this a hard error unless the user opts in (e.g. `--assume-local`), or at least require the caller to have already resolved to a local key. WDYT?

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+		}
+		ext := strings.ToLower(filepath.Ext(arg))
+		if ext != ".nupkg" && ext != ".snupkg" {
+			continue
+		}
+		pattern := arg
+		if !filepath.IsAbs(pattern) {
+			pattern = filepath.Join(workingDir, pattern)
+		}
+		matches, err := filepath.Glob(pattern)
```
agrasth, line 543: The extension check earlier in this function filters the *pattern*, not the *matches* from `filepath.Glob`. 
Safe today because only literal `.nupkg`/`.snupkg` args are accepted before glob expansion, but defensive re-filtering of `matches` by extension is a two-line change and prevents a future regression.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// isRestoreCommand returns true for commands that download packages (need dependency collection).
+func isRestoreCommand(sub string) bool {
+	switch sub {
+	case "restore", "install", "update", "build", "add":
```
agrasth, line 887: `isRestoreCommand` returns true for `"build"` and `"add"`. 
`dotnet add package` uses a different auth path than `restore`; `dotnet build` implicitly restores but doesn't accept all the same source flags. 
Can we confirm both are actually exercised end-to-end in the CLI PR — if not, we can drop them so we don't inject `--source` where it's invalid.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+}
+
+// isPackCommand returns true for the pack subcommand, which produces .nupkg/.snupkg files locally.
+func isPackCommand(sub string) bool {
+	return sub == "pack"
+}
+
+// needsConfig returns true when the subcommand needs a NuGet source (restore, push, etc.).
+// pack is a local-only operation and does not use a NuGet source.
+func needsConfig(sub string) bool {
```
agrasth: `needsConfig` is defined but never referenced anywhere in this file. Dead code — please delete.

artifactory/commands/nuget/command.go
bhanurp: Fixed `InsecureSkipVerify` is now hardcoded to `false`. The TLS verification path is handled by the Artifactory HTTP client which already honours `allowInsecureConnections`; setting it again on the NuGet-specific client was a regression.

artifactory/commands/nuget/command.go
bhanurp: Fixed `buildCmd` now normalises the subcommand before the comparison so `"nuget push"` and `"push"` are both handled consistently.

artifactory/commands/nuget/command.go
bhanurp: Fixed virtual-repo resolution failure is now a hard error instead of a warn-and-continue.

artifactory/commands/nuget/command.go
bhanurp: Fixed glob matches are now re-filtered by extension after `filepath.Glob` returns, guarding against OS-level glob expansion returning files that match the directory pattern but not the extension.

artifactory/commands/nuget/command.go
bhanurp: `dotnet add package` triggers a restore implicitly and respects `--source` the same way `dotnet restore` does, so credential injection is correct for it. The comment in `isRestoreCommand` was updated to document this.

artifactory/commands/nuget/command.go
bhanurp: Fixed `needsConfig` deleted.

artifactory/commands/nuget/auth.go
```diff
@@ -0,0 +1,43 @@
...
+	protocolVersion := "3"
+	if useV2 {
+		protocolVersion = "2"
+	}
+
+	content := fmt.Sprintf(dotnetutils.ConfigFileFormat,
+		sourceURL, protocolVersion, allowInsecure, user, password)
+
+	configPath = filepath.Join(tmpDir, "nuget.config")
+	if err := os.WriteFile(configPath, []byte(content), 0600); err != nil {
```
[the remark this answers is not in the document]
bhanurp: Fixed — XML attribute values are now escaped via `encoding/xml.EscapeText` before interpolation into the NuGet config template.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,433 @@
...
+	}()
+	length, _ := reader.Length()
+	if length == 0 {
+		return fmt.Errorf("no uploaded NuGet artifacts found at the expected paths for property stamping: %s", strings.Join(patterns, ", "))
+	}
+	if _, err := servicesManager.SetProps(services.PropsParams{Reader: reader, Props: props}); err != nil {
+		return fmt.Errorf("stamp build properties on uploaded NuGet artifacts: %w", err)
+	}
+	log.Info(fmt.Sprintf("Stamped build properties on %d NuGet artifact(s).", length))
+	return nil
```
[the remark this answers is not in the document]
bhanurp: Acknowledged — retry logic for post-push Artifactory search (eventual consistency) is tracked as a follow-up. The current implementation does a single search; a transient indexing delay would cause the artifact path to be missing from build-info. Will address in a follow-up PR.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,433 @@
...
+		"-r", "--runtime",
+		"-s", "--source", "-source",
+		"-solutiondirectory",
+		"--tl",
+		"-v", "--verbosity", "-verbosity":
+		return true
+	default:
+		return false
+	}
+}
```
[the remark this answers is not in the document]
bhanurp: Acknowledged — pass-through flags for the `build` subcommand (e.g. `--configuration`, `--framework`) are forwarded unchanged to the native CLI. No special handling is needed here.

go.mod
```diff
@@ -204,3 +204,5 @@ require (
 // replace github.com/gfleury/go-bitbucket-v1 => github.com/gfleury/go-bitbucket-v1 v0.0.0-20230825095122-9bc1711434ab
 
 // replace github.com/ktrysmt/go-bitbucket => github.com/ktrysmt/go-bitbucket v0.9.80
+
+replace github.com/jfrog/build-info-go => github.com/bhanurp/build-info-go v1.10.10-0.20260812084325-6352a2560238
```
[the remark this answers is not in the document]
bhanurp: Fixed — the personal fork `replace` directive has been removed from `go.mod`.

artifactory/commands/nuget/auth.go
```diff
@@ -0,0 +1,45 @@
...
+		return "", fmt.Errorf("get NuGet source details: %w", err)
+	}
+
+	u, err := url.Parse(sourceURL)
+	if err != nil {
+		return "", fmt.Errorf("parse NuGet source URL: %w", err)
+	}
+	u.User = url.UserPassword(user, password)
+	return u.String(), nil
+}
```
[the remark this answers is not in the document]
bhanurp: The credentials are injected as a `--source` URL with Basic Auth embedded (not as separate `--username`/`--password` flags), which is the standard NuGet V2/V3 auth pattern. The URL is not written to disk and is only visible in the process table for the duration of the nuget.exe/dotnet invocation — same as any other CLI tool that accepts credentials via URL.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+				// index.json from a URL with embedded JWT credentials.
+				// Note: dotnet CLI uses --source / -s (POSIX style); nuget.exe uses -Source.
+				authenticatedURL, err := SourceURLWithCredentials(c.serverDetails, repo, c.useNugetV2)
+				if err != nil {
+					return err
+				}
+				c.args = append(c.args, "--source", authenticatedURL)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
bhanurp: Fixed — credential injection is now gated to restore commands only (`isRestoreCommand`). Pack and passthrough commands do not receive `--source` injection.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+    <clear />
+    <add key=%q value=%q`+allowInsecure+` />
+  </packageSources>
+  <packageSourceCredentials>
+    <%s>
+      <add key="Username" value=%q />
+      <add key="ClearTextPassword" value=%q />
+    </%s>
+  </packageSourceCredentials>
+</configuration>`, sourceName, sourceURL, sourceName, user, password, sourceName)
```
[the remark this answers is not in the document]
bhanurp: Fixed — values interpolated into the NuGet config XML are now escaped via `encoding/xml.EscapeText` so special characters in usernames or tokens cannot break the XML structure.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,793 @@
...
+	snupkgPushURL := rtURL + "/api/nuget/v2/" + c.repoDeploy + "/symbolpackage"
+	skipDuplicate := hasSkipDuplicate(c.args)
+	noSymbols := hasNoSymbols(c.args)
+
+	//nolint:gosec
+	tlsCfg := &tls.Config{InsecureSkipVerify: c.allowInsecureConnections}
+	httpClient := &http.Client{Transport: &http.Transport{
+		TLSClientConfig: tlsCfg,
+		Proxy:           http.ProxyFromEnvironment,
+	}}
```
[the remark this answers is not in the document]
bhanurp: Fixed — `InsecureSkipVerify` is hardcoded to `false`. TLS verification is handled by the Artifactory HTTP client which already applies `allowInsecureConnections`.

artifactory/commands/nuget/command.go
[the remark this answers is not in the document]
bhanurp: Fixed — added exponential-backoff retry (up to 5 attempts, starting at 2 s, doubling each round) around `SearchItems` so a transiently-empty index does not cause a spurious error. Also propagated the `reader.Length()` error instead of silently ignoring it. Pushed in commit `511204e`.

artifactory/commands/nuget/command.go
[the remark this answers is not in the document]
bhanurp: Fixed in commit `b2af3f6` — added `-o`/`--output`, `-c`/`--configuration`, and `-f`/`--framework` to `restoreOptionTakesValue` so their values are not mistaken for the solution/project target path when `restoreTarget` parses `dotnet build` args.

artifactory/commands/nuget/command.go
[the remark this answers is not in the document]
bhanurp: Fixed in commit `b2af3f6` — added `Timeout: 5 * time.Minute` to the direct-push `http.Client` so a stalled `client.Do(req)` cannot wait indefinitely.

artifactory/commands/nuget/auth.go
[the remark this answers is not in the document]
bhanurp: Acknowledged. The `dotnet` restore path currently embeds credentials in `--source <url>`, which is visible via process inspection. Switching to `NuGetPackageSourceCredentials_*` env vars (the non-argv mechanism) for `dotnet` is the correct long-term fix and will be tracked as a follow-up. The `nuget.exe` path already uses a temp config file (`injectCredentialsViaTempConfig`) to avoid this.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,945 @@
...
+				// dotnet CLI restore: use a temp nuget.config (same approach as nuget.exe) to
+				// avoid embedding credentials in the --source process argument, which is visible
+				// to all local users via /proc/<pid>/cmdline or ps aux.
+				// dotnet restore supports --configfile (double-dash, POSIX style).
+				cleanup, err := c.injectCredentialsViaTempConfig(repo)
+				if err != nil {
+					return err
+				}
+				defer cleanup()
+			}
```
fluxxBot: Both the if c.toolchainType == dotnetutils.Nuget && isRestoreCommand(...) (line 141) and else if c.toolchainType != dotnetutils.Nuget && isRestoreCommand(...) (line 158) branches call c.injectCredentialsViaTempConfig(repo) identically — only the comments differ. Simplify to if isRestoreCommand(c.subCommand) with a single call. The per-toolchain distinction is documented inside injectCredentialsViaTempConfig itself

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,945 @@
...
+// into the process environment (via NuGetPackageSourceCredentials_) before this is called.
+func (c *NuGetFlexPackCommand) buildCmd() *exec.Cmd {
+	if c.toolchainType == dotnetutils.DotnetCore {
+		return exec.Command("dotnet", append(strings.Fields(c.subCommand), c.args...)...)
+	}
+	return exec.Command("nuget", append([]string{c.subCommand}, c.args...)...)
+}
+
+// injectCredentialsViaTempConfig handles credential injection for nuget.exe. It writes a
+// temporary nuget.config with a V2 source URL and <packageSourceCredentials>.
```
fluxxBot: Header says "a V2 source URL" but line 274 calls NuGetExeV3SourceDetails. The inner comment at line 272 correctly explains V3 is safe in the config-file path. The function header is stale — update "V2 source URL" to "V3 source URL (safe in config file; not re-embedded by nuget.exe into RestoreSources)".

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,945 @@
...
+		_ = os.Remove(tmpFile.Name())
+		return nil, fmt.Errorf("close temp nuget.config: %w", err)
+	}
+
+	// nuget.exe uses single-dash POSIX style; dotnet CLI uses double-dash POSIX style.
+	configFlag := "-ConfigFile"
+	if c.toolchainType != dotnetutils.Nuget {
+		configFlag = "--configfile"
+	}
+	c.args = append(c.args, configFlag, tmpFile.Name())
```
fluxxBot, line 307: c.args = append(c.args, configFlag, tmpFile.Name()) mutates the command's arg slice. The cleanup func removes the temp file but does not revert c.args. If Run() is ever called twice (in tests or retry logic), each call appends another --configfile entry. Snapshot and restore: save origArgs := c.args before appending, restore in cleanup.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,945 @@
...
+	case isPushCommand(c.subCommand):
+		return c.collectAndStampPushArtifacts(buildName, buildNumber)
+	case isPackCommand(c.subCommand):
+		return c.collectPackArtifacts(buildName, buildNumber, packSnapshot, packOutputDir)
+	}
+	return nil
+}
+
+// buildCmd builds the exec.Cmd for the native nuget.exe or dotnet CLI.
+// Credentials are already injected into c.args (via -Source <url> or -ConfigFile) and
```
fluxxBot: "into the process environment (via NuGetPackageSourceCredentials_)" — no environment variable injection exists in this codebase. Only --configfile is appended. Remove or correct this sentence.

artifactory/commands/nuget/command.go
```diff
@@ -0,0 +1,945 @@
...
+				delay *= 2
+			}
+		}
+	}
+	return 0, fmt.Errorf("no uploaded NuGet artifacts found at the expected paths after %d attempts: %s", maxAttempts, strings.Join(patterns, ", "))
+}
+
+// resolvePackagePaths returns all .nupkg and .snupkg file paths from args, expanding
+// glob patterns relative to workingDir. Flags (args starting with -) are skipped.
+func resolvePackagePaths(workingDir string, args []string) ([]string, error) {
```
fluxxBot, line 529: pushPackagesToArtifactory calls the local resolvePackagePaths to get file paths; collectAndStampPushArtifacts then calls nugetflex.CollectPushArtifacts on the same args, doing the same glob expansion a second time. The two callers share the same c.args source — extract paths once and pass them to both, or have pushPackagesToArtifactory accept pre-resolved paths.

fluxxBot (dismissed):

fluxxBot (approved):

agrasth (approved):

PR #541, opened 2026-08-27
Fix misleading security scan status message in jf skills publish

tal-qwak:
## Summary
- `jf skills publish` (without `--skip-scan`) prints `[SUCCESS] Skill "<slug>" v<version> passed security scan.` as soon as the Skills Xray gate (`GET .../api/skills/{repoKey}/xrayStatus`) returns `APPROVED`.
- In testing, this message printed ~2 seconds after upload, before Xray's own async scan pipeline for skill artifacts had produced a real verdict — a subsequent direct query to `xray/api/v1/summary/artifact` for the same artifact showed a Critical "Malicious skill" finding (`XRAY-00672`) that the gate never surfaced.
- The CLI's polling logic in `xray_gate.go` is correct as-is (it already polls on `SCAN_IN_PROGRESS` until a terminal status). The issue is purely the wording: `APPROVED` reflects Artifactory/Xray's current indexing state at the moment of the check, not a guaranteed completed content scan, so asserting the artifact "passed security scan" overclaims what the gate can currently promise.

## Change
Reworded both `APPROVED` log lines (the immediate case in `CheckXrayGate` and the post-poll case in `pollUntilDone`):

```
- [SUCCESS] Skill "<slug>" v<version> passed security scan.
+ [SCAN PENDING] Skill "<slug>" v<version>: no violations found yet.
```

Matches the file's existing bracket-tag convention (`[VIOLATION]` for the blocked case), and leads with the one word that matters (`PENDING`, not `SUCCESS`) so it's skimmable without reading the full line.

No behavior change — same statuses handled the same way, same control flow. Confirmed `go build ./agent/skills/...` and `go test ./agent/skills/...` pass unchanged.

## Note for maintainers
This does not close the underlying gap — it only makes the CLI honest about what it doesn't know. The actual cause (why `APPROVED` can be returned before a skill's async content scan completes) is server-side, in Xray's artifact-processed status computation, not in this CLI. `jf skills install` shares the same gate check and remains exposed to the same window. Happy to share full details on request.

## Test plan
- [x] `go build ./agent/skills/...`
- [x] `go test ./agent/skills/...`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated approved-scan messages to clearly indicate when no Xray violations are found, including both immediate and polling completion results.


tal-qwak:

agent/skills/common/xray_gate.go
```diff
@@ -63,7 +63,7 @@ func CheckXrayGate(params XrayGateParams) error {
 		log.Info(fmt.Sprintf("Xray scanning is disabled for repository '%s'. Skipping scan.", params.RepoKey))
 		return nil
 	case services.SkillXrayStatusApproved:
-		log.Info(fmt.Sprintf("[SUCCESS] Skill \"%s\" v%s passed security scan.", params.Slug, params.Version))
+		log.Info(fmt.Sprintf("[SCAN PENDING] Skill \"%s\" v%s: no violations found yet.", params.Slug, params.Version))
```
bhanurp, line 66: If jfrog-cli is doing checks to figure out what is the XRAY status via artifactory API, then changes has to be made on the API to return proper status instead of Approved. If the api is returning approved but still it means that scanning is not completed then API response should be improved.

tal-qwak: Agreed, and this PR is intentionally scoped as a mitigation, not the real fix. We traced this down to the actual root cause: Xray's `GetArtifactState()` (`xray/backend/backend/service/artifact_details/artifact_details_service.go` in `JFROG/xray`) sets `ArtifactProcessed = true` for `PendingScanValue` ("Pending Scan") and `INDEXING`, not just `SCANNED` - so the Skills gate can report `APPROVED` before the async content-scan (aiscanner) pipeline has actually produced a verdict. We drafted a fix scoped to `PkgType == skills` there, but couldn't push a branch to `JFROG/xray` (hit what looks like a repo-level policy blocking direct branch creation) - happy to hand off the diff or open it properly once we know the right contribution path for that repo. This PR just reduces the blast radius on the CLI side in the meantime.

bhanurp: There should be an improvement again when there is a proper fix on `JFROG/xray`.

bhanurp (approved):

fluxxBot (approved):

PR #542, opened 2026-08-30
RTECO-1362 - Add --fail-on-uncollected-deps flag for npm install/ci

udaykb2:
## Summary
Expose `--fail-on-uncollected-deps` on `jf npm install` / `ci` (and other npm commands that collect build-info). Fail the command when selected dependency types cannot be collected into build-info.
- Flag name: `--fail-on-uncollected-deps` (not “missing” — the package is usually installed; collection of integrity/checksum failed). Reusable later on other package managers; `peer` / `bundle` are npm *values*.
- Values: `all`, or a comma-separated combination of `regular`, `peer`, `optional`, `bundle`. `all` cannot be mixed with other tokens.
- Requires build-info collection: `--build-name` and `--build-number` (or `JFROG_CLI_BUILD_NAME` / `JFROG_CLI_BUILD_NUMBER` / project config). Rejected up front if the flag is set without them.
- Extraction stays in npm `Init()` with `--disable-cvs-check` and `ExtractNpmOptionsFromArgs`. `jf npm` uses `SkipFlagParsing`; jfrog-cli does not parse these flags today (only `--run-native` is stripped before `Init()`). Moving only this flag to jfrog-cli would be inconsistent with the rest of npm.
Depends on [build-info-go#418](https://github.com/jfrog/build-info-go/pull/418) @ `ea18305`.

-----

## Summary by CodeRabbit

* **New Features**
  * Renamed the npm option to `--fail-on-uncollected-deps`.
  * Builds can fail when dependency integrity or checksum data cannot be collected for build information.
  * Supports accepted dependency types as comma-separated values.
  * Accepts `all` only when used by itself; invalid combinations and values are rejected.
  * Requires a build name and number when this option is provided.
  * Updated npm command configuration and help text.
  * Added configuration support for the agent APM command.
  * Normalized unnamed npm commands to `rt_npm`.


cliutils/flagkit/flags.go
```diff
@@ -371,6 +371,7 @@ const (
 	runNative          = "run-native"
 	npmWorkspaces      = "workspaces"
 	disableCVSCheck    = "disable-cvs-check"
+	failOnMissingDeps  = "fail-on-missing-deps"
```
bhanurp: `Failonmissingdeps` name is not clear. Isn't this related to build-info? Lets have a common flag name so the same flag name can be reused across multiple packages.

artifactory/commands/npm/npmcommand.go
```diff
@@ -168,8 +181,19 @@ func (nc *NpmCommand) Init() error {
 	if err != nil {
 		return err
 	}
+	// Extract --fail-on-missing-deps flag. Accepts a granular string value: "all", "" (default, never fail),
+	// or a comma-separated combination of "regular", "peer", "optional", "bundle".
+	filteredNpmArgs, failOnMissingDeps, err := coreutils.ExtractStringOptionFromArgs(filteredNpmArgs, "fail-on-missing-deps")
+	if err != nil {
+		return err
+	}
```
fluxxBot: there are function in jfrog-cli and core which does this, IMO, flag extraction should happen in jfrog-cli, artifactory should only contain the flag values passed by jfrog-cli

artifactory/commands/npm/npmcommand.go
```diff
@@ -591,6 +616,33 @@ func (nc *NpmCommand) GetRepo() string {
...
+		if !validValues[trimmed] {
+			return errorutils.CheckErrorf(
+				"invalid --fail-on-missing-deps value: '%s'. "+
+					"Valid values are: all, peer, optional, regular, bundle, or comma-separated combinations (e.g., peer,optional,bundle)",
+				trimmed)
+		}
+	}
+	return nil
+}
+
```
fluxxBot: there are fucntion for validation, please check and try to use them

fluxxBot (changes requested):

cliutils/flagkit/flags.go
```diff
@@ -1157,6 +1158,7 @@ var flagsMap = map[string]components.Flag{
 	npmDetailedSummary:       components.NewBoolFlag(detailedSummary, "Set to true to include a list of the affected files in the command summary.", components.WithBoolDefaultValueFalse()),
 	nugetV2:                  components.NewBoolFlag(nugetV2, "Set to true if you'd like to use the NuGet V2 protocol when restoring packages from Artifactory.", components.WithBoolDefaultValueFalse()),
 	disableCVSCheck:          components.NewBoolFlag(disableCVSCheck, "Set to true to disable the CVS check that verifies if 404 errors are due to blocked packages.", components.WithBoolDefaultValueFalse()),
+	failOnMissingDeps:        components.NewStringFlag(failOnMissingDeps, "Fail the build if a dependency's tarball can't be resolved from the npm cache. Accepts 'all' (fail for every missing dependency type), or a comma-separated combination of 'regular', 'peer', 'optional', 'bundle' (e.g. 'peer,optional,bundle') to fail only for the specified types. Only applies when collecting build-info with --build-name and --build-number.", components.SetMandatoryFalse()),
```
bhanurp:
```suggestion
	failOnMissingDeps:        components.NewStringFlag(failOnMissingDeps, "Fail the build (non-zero exit code) if a dependency's tarball can't be resolved from the npm cache. Accepts 'all' (fail for every missing dependency type), or a comma-separated combination of 'regular', 'peer', 'optional', 'bundle' (e.g. 'peer,optional,bundle') to fail only for the specified types. Only applies when collecting build-info with --build-name and --build-number.", components.SetMandatoryFalse()),
```

artifactory/commands/npm/npmcommand.go
udaykb2: jfrog-cli doesn't parse any npm flags today. Every npm flag - disable-cvs-check, detailed-summary, xray-scan, format, and now this one, gets extracted here in jfrog-cli-artifactory

cliutils/flagkit/flags.go
udaykb2: done

cliutils/flagkit/flags.go
udaykb2: renamed it to fail-on-uncollected-deps, 

I feel this is the right name, because it is used along with --build-name and --build-number, so if we are unable to collect information for any dependency, it will fail

artifactory/commands/npm/npmcommand.go
udaykb2: Didn't find an existing generic "one of these allowed strings" helper elsewhere in the codebase to reuse directly (closest precedent is format.ParseOutputFormat in jfrog-cli-core, which is typed to a different enum) - let me know if you had a specific function in mind

cliutils/flagkit/flags.go
udaykb2: changed flag ame to failOnUncollectedDeps.

Because this flag will be used with --build-name and --build-number and if we are unable to collect those dependencies for build-info collection, we are throwing error (if user uses flag, else we continue with legacy behaviour of Warn + Debug)

fluxxBot (dismissed):

bhanurp (dismissed):

fluxxBot (approved):

naveenku-jfrog (approved):

PR #543, opened 2026-08-31
XRAY-145659 - Add generic Zero Touch Remediation service

attiasas:
## Summary

Adds a build-tool-neutral **Zero Touch Remediation (ZTR)** package under `artifactory/zerotouchremediation`. Callers opt in with `JFROG_CLI_ZTR_COMPONENTS_ENABLED` (`strconv.ParseBool`: `true`, `TRUE`, `1`, `t`, …). Disabled by default.

When enabled, `RunIfEnabled`:

1. Skips if the command is not in the tool’s `RelevantCommands`.
2. Checks JFrog Xray ≥ `3.154.0` (`ZeroTouchRemediationMinXrayVersion`) for the component-resolution API.
3. Resolves the project root, bootstraps missing lockfiles via the tool, and discovers lock artifacts.
4. Sends each lockfile to Xray and writes returned changes.
5. Returns a restore callback so the **caller** can revert lockfiles if the subsequent build command fails. On success the remediating lockfiles stay on disk.

Operational failures are best-effort: warn/debug and continue. Remediation must not fail the user’s build.

Lockfile writes stay under the project root: relative paths only, refuse `..`, and `Lstat` every path component so a symlink parent cannot redirect the write. Missing leaf files are allowed (bootstrap). Apply failures restore immediately.

This PR is the generic contract and implementation only. It does **not** wire any CLI command.

### Follow-up

- **#544** — opt-in ZTR for `jf npm install` / `jf npm ci` (the path that actually installs from the lockfile).
- **npm publish is out of scope.** `jf npm publish` packs and deploys; it does not resolve or install from the lockfile. `package-lock.json` is omitted from the npm tarball, so lockfile remediation would not change the published artifact. We are not taking that work forward.

### Files

- `artifactory/zerotouchremediation/buildtool.go` — `BuildTool` / `CommandRunner` contracts (per-tool; a manager without a lock artifact returns empty slices and the generic flow skips apply)
- `artifactory/zerotouchremediation/service.go` — gating, Xray call, apply, rollback, symlink checks
- `go.mod` / `go.sum` — `jfrog-client-go` bump for the component-resolution API

## Test plan

- [x] `go test ./artifactory/zerotouchremediation`
- [x] `go vet ./artifactory/zerotouchremediation`

### DIFF
--- a/artifactory/zerotouchremediation/buildtool.go
+++ b/artifactory/zerotouchremediation/buildtool.go
@@ -16,8 +16,10 @@ type BuildTool interface {
 	ProjectRoot(workingDir string) (string, error)
 	// EnsureLockfiles materializes expected lock artifacts when absent and the command allows it.
 	// Returns paths (relative to project root) that were bootstrapped and did not exist before.
+	// Tools without lockfiles (for example Maven) return a nil/empty slice.
 	EnsureLockfiles(ctx context.Context, projectRoot, command string, runner CommandRunner, bootstrapArgs ...string) (bootstrapped []string, err error)
-	// DiscoverLockfiles returns all lockfiles (paths relative to project root).
+	// DiscoverLockfiles returns lock artifacts relative to project root.
+	// Tools without lockfiles return a nil/empty slice; the generic flow then skips apply.
 	DiscoverLockfiles(workingDir string) ([]Lockfile, error)
 }
 
--- a/artifactory/zerotouchremediation/buildtool.go
+++ b/artifactory/zerotouchremediation/buildtool.go
@@ -1,0 +1,26 @@
+package zerotouchremediation
+
+import (
+	"context"
+	"slices"
+)
+
+// CommandRunner runs a build-tool subprocess (injected for tests).
+type CommandRunner func(ctx context.Context, projectRoot string, args ...string) error
+
+// BuildTool describes a package manager for the generic resolution flow.
+type BuildTool interface {
+	ToolName() string
+	RelevantCommands() []string
+	// ProjectRoot resolves monorepo/reactor/solution root from workingDir.
+	ProjectRoot(workingDir string) (string, error)
+	// EnsureLockfiles materializes expected lock artifacts when absent and the command allows it.
+	// Returns paths (relative to project root) that were bootstrapped and did not exist before.
+	EnsureLockfiles(ctx context.Context, projectRoot, command string, runner CommandRunner, bootstrapArgs ...string) (bootstrapped []string, err error)
+	// DiscoverLockfiles returns all lockfiles (paths relative to project root).
+	DiscoverLockfiles(workingDir string) ([]Lockfile, error)
+}
+
+func IsRelevantCommand(tool BuildTool, command string) bool {
+	return slices.Contains(tool.RelevantCommands(), command)
+}
--- a/artifactory/zerotouchremediation/service.go
+++ b/artifactory/zerotouchremediation/service.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	clientutils "github.com/jfrog/jfrog-client-go/utils"
@@ -13,11 +14,13 @@ import (
 	"github.com/jfrog/jfrog-client-go/xray/services"
 )
 
-// ZtrComponentsEnabledEnvVar enables Zero Touch Remediation when set to "true".
-// Feature is disabled by default.
+// ZtrComponentsEnabledEnvVar enables Zero Touch Remediation when set to a
+// truthy value per strconv.ParseBool ("true", "TRUE", "1", "t", …). Disabled by default.
 const ZtrComponentsEnabledEnvVar = "JFROG_CLI_ZTR_COMPONENTS_ENABLED"
 
-const ZeroTouchRemediationMinVersion = "3.154.0"
+// ZeroTouchRemediationMinXrayVersion is the minimum JFrog Xray version that
+// supports the component-resolution / Zero Touch Remediation API.
+const ZeroTouchRemediationMinXrayVersion = "3.154.0"
 
 var noopRestore = func() error { return nil }
 
@@ -34,10 +37,6 @@ func SkipRemediation(message string, cause error) (func() error, bool, error) {
 	return noopRestore, false, nil
 }
 
-func skipRemediationWarn(message string, cause error) (func() error, bool, error) {
-	return skipWithRestore(noopRestore, message, cause)
-}
-
 func skipWithRestore(restore func() error, message string, cause error) (func() error, bool, error) {
 	if cause != nil {
 		log.Warn(message + cause.Error())
@@ -51,7 +50,8 @@ func skipWithRestore(restore func() error, message string, cause error) (func() 
 }
 
 func IsComponentResolutionEnabled() bool {
-	return os.Getenv(ZtrComponentsEnabledEnvVar) == "true"
+	value, err := strconv.ParseBool(os.Getenv(ZtrComponentsEnabledEnvVar))
+	return err == nil && value
 }
 
 // Lockfile is a CLI-side lock artifact. Path is relative to project root (read/write only — not sent to Xray).
@@ -81,21 +81,21 @@ func RunIfEnabled(ctx context.Context, client ComponentResolutionClient, repo st
 	}
 	version, err := client.GetVersion()
 	if err != nil {
-		return skipRemediationWarn("Zero Touch Remediation skipped: could not get Xray version: ", err)
+		return SkipRemediation("Zero Touch Remediation skipped: could not get Xray version: ", err)
 	}
 	log.Debug("Xray version: ", version)
-	if versionErr := clientutils.ValidateMinimumVersion(clientutils.Xray, version, ZeroTouchRemediationMinVersion); versionErr != nil {
-		return skipRemediationWarn("Zero Touch Remediation is not supported on the current Xray version. ", versionErr)
+	if versionErr := clientutils.ValidateMinimumVersion(clientutils.Xray, version, ZeroTouchRemediationMinXrayVersion); versionErr != nil {
+		return SkipRemediation("Zero Touch Remediation is not supported on the current Xray version. ", versionErr)
 	}
 	log.Debug("Running Zero Touch Remediation at '"+repo+"' RT repository for tool:", tool.ToolName())
 	projectRoot, err := tool.ProjectRoot(workingDir)
 	if err != nil {
-		return skipRemediationWarn("Zero Touch Remediation skipped: could not resolve project root: ", err)
+		return skipWithRestore(noopRestore, "Zero Touch Remediation skipped: could not resolve project root: ", err)
 	}
 	log.Debug("Ensuring lockfiles in project root: ", projectRoot)
 	bootstrapped, err := tool.EnsureLockfiles(ctx, projectRoot, command, runner, bootstrapArgs...)
 	if err != nil {
-		return skipRemediationWarn("Zero Touch Remediation skipped: ", err)
+		return skipWithRestore(noopRestore, "Zero Touch Remediation skipped: ", err)
 	}
 	skipAfterBootstrap := func(message string, cause error) (func() error, bool, error) {
 		return skipWithRestore(restoreAbsentLockfiles(projectRoot, bootstrapped), message, cause)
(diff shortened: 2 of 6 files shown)


artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,244 @@
...
+			continue
+		}
+		toWrite = append(toWrite, Lockfile{Path: lf.Path, Content: []byte(resp.Lockfile)})
+		totalChanges += len(resp.Changes)
+		log.Debug("Remediated", lf.Path, "with", len(resp.Changes), "package change(s)")
+		for _, ch := range resp.Changes {
+			log.Debug("  ", ch.Package, ":", ch.BeforeIntegrity, "→", ch.AfterIntegrity)
+		}
+	}
+	if len(toWrite) == 0 {
```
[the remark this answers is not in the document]
attiasas: Verified: restore only tracked lockfiles in toWrite. A bootstrapped file with no Xray changes was left behind on command failure, including the empty-toWrite noopRestore path.

Restore now deletes every validated bootstrapped path, whether Xray changed it or not. Covered by TestRunIfEnabled_RestoresBootstrappedLockfileWhenNoChanges and TestRunIfEnabled_RestoresUnchangedBootstrappedLockfile.

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+		return nil, errorutils.CheckError(cause)
+	}
+
+	for _, lf := range lockfiles {
+		fullPath, pathErr := containedLockfilePath(projectRoot, lf.Path)
+		if pathErr != nil {
+			return fail(pathErr)
+		}
+		if linkErr := rejectSymlink(fullPath); linkErr != nil {
```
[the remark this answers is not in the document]
attiasas: Verified: containedLockfilePath does not resolve intermediate symlinks, so projectRoot/module pointing outside the tree passed the relative-path check and Lstat only on the lockfile accepted a regular target.

ApplyLockfiles now Lstats every component between projectRoot and the lockfile. It still does not walk ancestors of projectRoot: on macOS /var is a symlink to /private/var, and that would refuse writes under a typical TempDir.

Not using descriptor-relative openat/O_NOFOLLOW here — this CLI writes into the user's project, and that pattern is not used in this repo. Covered by TestApplyLockfiles_RejectsSymlinkParentDir.

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+		toWrite = append(toWrite, Lockfile{Path: lf.Path, Content: []byte(resp.Lockfile)})
+		totalChanges += len(resp.Changes)
+		log.Debug("Remediated", lf.Path, "with", len(resp.Changes), "package change(s)")
+		for _, ch := range resp.Changes {
+			log.Debug("  ", ch.Package, ":", ch.BeforeIntegrity, "→", ch.AfterIntegrity)
+		}
+	}
+	if len(toWrite) == 0 {
+		return restoreAbsentLockfiles(projectRoot, bootstrapped), false, nil
```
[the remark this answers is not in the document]
attiasas: Verified: after EnsureLockfiles, DiscoverLockfiles errors, Xray errors, and ApplyLockfiles failures still returned noopRestore via skipRemediationWarn, so the caller could not delete generated lockfiles if the later command failed.

Those paths now return restoreAbsentLockfiles. Covered by TestRunIfEnabled_DiscoverErrorRestoresBootstrapped, TestRunIfEnabled_RemediateErrorRestoresBootstrapped, and TestRunIfEnabled_ApplyErrorRestoresLeftoverBootstrapped.

artifactory/zerotouchremediation/service.go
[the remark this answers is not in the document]
attiasas: The intermediate-symlink case is covered. Not adding descriptor-relative openat/O_NOFOLLOW.

This is a user-local CLI writing lockfiles in the project it was invoked on, not a privileged service. A TOCTOU swap between Lstat and WriteFile requires a concurrent process already able to mutate that tree. This repo has no Openat/O_NOFOLLOW usage, and a portable no-follow walk is a large, platform-specific change for that threat model.

artifactory/zerotouchremediation/buildtool.go
```diff
@@ -0,0 +1,26 @@
...
+type BuildTool interface {
+	ToolName() string
+	RelevantCommands() []string
+	// ProjectRoot resolves monorepo/reactor/solution root from workingDir.
+	ProjectRoot(workingDir string) (string, error)
+	// EnsureLockfiles materializes expected lock artifacts when absent and the command allows it.
+	// Returns paths (relative to project root) that were bootstrapped and did not exist before.
+	EnsureLockfiles(ctx context.Context, projectRoot, command string, runner CommandRunner, bootstrapArgs ...string) (bootstrapped []string, err error)
+	// DiscoverLockfiles returns all lockfiles (paths relative to project root).
+	DiscoverLockfiles(workingDir string) ([]Lockfile, error)
```
fluxxBot, line 23: is it same for all packageManagers?
not all package managers have lockfile, example maven

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/jfrog/jfrog-client-go/utils/errorutils"
+	"github.com/jfrog/jfrog-client-go/utils/log"
+	"github.com/jfrog/jfrog-client-go/xray/services"
+)
+
+// ZtrComponentsEnabledEnvVar enables Zero Touch Remediation when set to "true".
+// Feature is disabled by default.
+const ZtrComponentsEnabledEnvVar = "JFROG_CLI_ZTR_COMPONENTS_ENABLED"
+
+const ZeroTouchRemediationMinVersion = "3.154.0"
```
fluxxBot: can we explicitly say which service's minimum version is this?

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,327 @@
...
+	} else {
+		log.Debug(message)
+	}
+	return noopRestore, false, nil
+}
+
+func skipRemediationWarn(message string, cause error) (func() error, bool, error) {
+	return skipWithRestore(noopRestore, message, cause)
+}
+
```
fluxxBot: is this wrapper really necessary here?

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,327 @@
...
+		log.Warn(message)
+	}
+	if restore == nil {
+		restore = noopRestore
+	}
+	return restore, false, nil
+}
+
+func IsComponentResolutionEnabled() bool {
+	return os.Getenv(ZtrComponentsEnabledEnvVar) == "true"
```
fluxxBot: let's not use string comparison, instead use - `value, err := strconv.ParseBool`
valid values are - true, TRUE, 1, etc.

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,327 @@
...
+// Operational failures are skipped (warn + continue) so remediation never fails the caller's build.
+// Returns a restore function to revert lockfile writes if the subsequent build-tool command fails, and remediated=true when at least one lockfile was updated.
+func RunIfEnabled(ctx context.Context, client ComponentResolutionClient, repo string, tool BuildTool, command, workingDir string, runner CommandRunner, bootstrapArgs ...string) (restore func() error, remediated bool, err error) {
+	if !IsComponentResolutionEnabled() || !IsRelevantCommand(tool, command) {
+		log.Debug("Zero Touch Remediation disabled or not relevant command: ", command)
+		return noopRestore, false, nil
+	}
+	version, err := client.GetVersion()
+	if err != nil {
+		return skipRemediationWarn("Zero Touch Remediation skipped: could not get Xray version: ", err)
```
fluxxBot: make sure this should be part of debug logs, otherwise unnecessary warn logs would be shown

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	log.Debug("Xray version: ", version)
+	if versionErr := clientutils.ValidateMinimumVersion(clientutils.Xray, version, ZeroTouchRemediationMinVersion); versionErr != nil {
+		return skipRemediationWarn("Zero Touch Remediation is not supported on the current Xray version. ", versionErr)
```
fluxxBot: make sure this should be part of debug logs, otherwise unnecessary warn logs would be shown

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,327 @@
...
+		return restoreAbsentLockfiles(projectRoot, bootstrapped), false, nil
+	}
+	log.Debug("Applying", len(toWrite), "remediated lockfile(s)...")
+	restore, err = ApplyLockfiles(projectRoot, toWrite, bootstrapped)
+	if err != nil {
+		return skipAfterBootstrap("Zero Touch Remediation skipped: failed to apply remediated lockfiles: ", err)
+	}
+	restore = composeRestore(restore, restoreAbsentLockfiles(projectRoot, bootstrappedNotWritten(bootstrapped, toWrite)))
+	log.Info("Zero Touch Remediation applied", totalChanges, "package change(s) across", len(toWrite), "lockfile(s)")
+	return restore, true, nil
```
fluxxBot, line 144: make sure we're restoring the backup in any means, wether it's a error thrown by us, crash or panic, we should be restoring the backup

artifactory/zerotouchremediation/service.go
```diff
@@ -0,0 +1,327 @@
...
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return errorutils.CheckError(err)
+	}
+	if info.Mode()&os.ModeSymlink != 0 {
+		return errorutils.CheckErrorf("refusing to write lockfile through symlink %s", path)
+	}
+	return nil
+}
```
fluxxBot, line 330: can you add a comment, explaining briefly about this functions, also mention any edge cases handled

fluxxBot (changes requested):

artifactory/zerotouchremediation/buildtool.go
attiasas: Not the same for every package manager. `BuildTool` is per-tool; Maven (and anything without a lock artifact) should return an empty slice from `EnsureLockfiles` / `DiscoverLockfiles`. The generic flow then skips apply. This slice does not add a Maven adapter.

Documented that on the interface.

artifactory/zerotouchremediation/service.go
attiasas: Renamed to `ZeroTouchRemediationMinXrayVersion` and noted in the comment that this is the minimum JFrog Xray version for the component-resolution API.

artifactory/zerotouchremediation/service.go
attiasas: Removed. Call sites now use `skipWithRestore(noopRestore, ...)` for warn-and-skip, or `SkipRemediation` for debug-and-skip.

artifactory/zerotouchremediation/service.go
attiasas: Switched to `strconv.ParseBool`. Unset, empty, and invalid values stay disabled (`err == nil && value`). Tests cover `TRUE`/`1`/`t` as well as `false`/`yes`.

artifactory/zerotouchremediation/service.go
attiasas: These compatibility gates now go through `SkipRemediation` (`log.Debug`). Warn stays on unexpected operational failures after the feature is actually running (project root, bootstrap, discover, Xray call, apply).

artifactory/zerotouchremediation/service.go
attiasas: Same as the GetVersion path: unsupported Xray version is now debug via `SkipRemediation`.

artifactory/zerotouchremediation/service.go
attiasas: `RunIfEnabled` returns a restore callback for the caller to invoke when the subsequent build command fails. That is intentional: on success the remediated lockfiles must stay on disk.

- Apply failures already restore immediately via `fail()` inside `ApplyLockfiles`.
- Process crash / SIGKILL cannot restore from user space.
- A panic after a successful apply is the integration layer's job (`defer` restore on failure). This PR has no command wiring; that belongs in the npm slice.

Not wrapping this package in `recover()` — it would either undo a successful remediation or only cover `log.Info` after apply.

artifactory/zerotouchremediation/service.go
attiasas: Added comments on `rejectSymlinksUnderRoot` and `rejectSymlink`: walk is Lstat-only, skips the project root (macOS `/var` -> `/private/var`), allows missing components so a lockfile can be created, and refuses symlink parents so writes cannot leave the tree.

fluxxBot: I do not see the comments on top of rejectSymlink, am I missing something here?
also are we creating this remediated lockfile in temp directory? can you explain me why this is required or maybe the high level flow? it will be easy for the reviewer to comprehend

attiasas: see latest commit
[CR changes](https://github.com/jfrog/jfrog-cli-artifactory/pull/543/commits/511f1f8c05c2ea15369ae073ed77b056b06bbab7) 

also you can checkout 
https://jfrog-int.atlassian.net/browse/XRAY-145659

fluxxBot (approved):

agrasth (approved):

agrasth (approved):

PR #544, opened 2026-08-31
XRAY-145659 - Add opt-in Zero Touch Remediation for npm install

attiasas:
## Summary

Wires opt-in **Zero Touch Remediation (ZTR)** into `jf npm install` and `jf npm ci`. Same gate as the generic service: `JFROG_CLI_ZTR_COMPONENTS_ENABLED` (`strconv.ParseBool`). Disabled by default.

Before the install runs, the command:

1. Discovers the npm project root (`package.json`, `package-lock.json`, or `npm-shrinkwrap.json`), including `--prefix` and workspaces.
2. Bootstraps a missing lockfile with `npm install --package-lock-only` on **install**. **ci** requires an existing lockfile.
3. Sends the lockfile through the generic ZTR service from **#543** (merged).
4. Writes remediating lockfiles, then runs the install.

If Xray remediates the lockfile, `jf npm install` switches to `npm ci` so npm honors lockfile integrity instead of re-resolving. On install failure, remediating lockfiles (and any lockfile created only for bootstrap) are restored.

Skipped when ZTR is off, the command is not install/ci, or the user is installing a named package (`jf npm install lodash`) rather than the project tree. Resolver repo comes from the project config’s resolver, else the npm registry URL.

Rebased onto `main` after #543. The Files tab is the npm integration only.

### Out of scope

**npm publish is not part of this work.** `jf npm publish` packs and deploys; it does not install from the lockfile. `package-lock.json` is omitted from the npm tarball, so lockfile remediation would not change the published artifact.

### Files

- `artifactory/commands/npm/ztr_discover.go` — project root and lockfile discovery
- `artifactory/commands/npm/ztr_buildtool.go` — npm `BuildTool` (relevant commands, bootstrap, discover)
- `artifactory/commands/npm/ztr_args_parse.go` — npm CLI flags that affect discovery/bootstrap (`--prefix`, `--omit`, `--registry`, …)
- `artifactory/commands/npm/zerotouchremediation.go` — Xray client, resolver repo, `RunIfEnabled` call
- `artifactory/commands/npm/npmcommand.go`, `npminstall.go`, `artifactoryinstall.go` — run ZTR, restore on failure, switch remediating install to `ci`
- `go.mod` / `go.sum` — `jfrog-cli-core` bump used by this integration

## Test plan

- [x] `go test ./artifactory/commands/npm`
- [x] `go vet ./artifactory/commands/npm`

## Summary by CodeRabbit

- **New Features**
  - Added opt-in zero-touch dependency remediation for npm projects.
  - Remediation can discover project and workspace lockfiles, generate missing lockfiles when needed, and apply updated resolutions.
  - Supports npm registries configured through command-line options and npm configuration.
  - Uses `npm ci` for dependency collection after lockfile remediation.

- **Bug Fixes**
  - Restores original lockfiles when installation fails.
  - Cleans up lockfiles created temporarily during remediation.
  - Skips unsupported symlinked lockfiles safely.

### DIFF
--- a/artifactory/commands/npm/common.go
+++ b/artifactory/commands/npm/common.go
@@ -14,6 +14,8 @@ type CommonArgs struct {
 	npmArgs            []string
 	serverDetails      *config.ServerDetails
 	useNative          bool
+	configFilePath     string
+	executablePath     string
 }
 
 func (ca *CommonArgs) SetServerDetails(serverDetails *config.ServerDetails) *CommonArgs {
@@ -42,6 +44,16 @@ func (ca *CommonArgs) UseNative() bool {
 
 func (ca *CommonArgs) SetUseNative(useNpmRc bool) *CommonArgs {
 	ca.useNative = useNpmRc
+	return ca
+}
+
+func (ca *CommonArgs) SetConfigFilePath(configFilePath string) *CommonArgs {
+	ca.configFilePath = configFilePath
+	return ca
+}
+
+func (ca *CommonArgs) SetExecutablePath(executablePath string) *CommonArgs {
+	ca.executablePath = executablePath
 	return ca
 }
 
--- a/artifactory/commands/npm/common.go
+++ b/artifactory/commands/npm/common.go
@@ -14,8 +14,6 @@ type CommonArgs struct {
 	npmArgs            []string
 	serverDetails      *config.ServerDetails
 	useNative          bool
-	configFilePath     string
-	executablePath     string
 }
 
 func (ca *CommonArgs) SetServerDetails(serverDetails *config.ServerDetails) *CommonArgs {
@@ -44,16 +42,6 @@ func (ca *CommonArgs) UseNative() bool {
 
 func (ca *CommonArgs) SetUseNative(useNpmRc bool) *CommonArgs {
 	ca.useNative = useNpmRc
-	return ca
-}
-
-func (ca *CommonArgs) SetConfigFilePath(configFilePath string) *CommonArgs {
-	ca.configFilePath = configFilePath
-	return ca
-}
-
-func (ca *CommonArgs) SetExecutablePath(executablePath string) *CommonArgs {
-	ca.executablePath = executablePath
 	return ca
 }
 
--- a/artifactory/commands/npm/npmcommand.go
+++ b/artifactory/commands/npm/npmcommand.go
@@ -26,6 +26,8 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	"github.com/jfrog/jfrog-client-go/utils/log"
 	"github.com/spf13/viper"
+
+	"github.com/jfrog/jfrog-cli-artifactory/artifactory/zerotouchremediation"
 )
 
 const (
@@ -355,14 +357,18 @@ func (nc *NpmCommand) Run() (err error) {
 	defer func() {
 		err = errors.Join(err, nc.installHandler.RestoreNpmrc())
 	}()
-	var installErr error
-	defer func() {
-		if installErr != nil && nc.restoreResolution != nil {
-			err = errors.Join(err, nc.restoreResolution())
-		}
-	}()
-	installErr = nc.installHandler.Install()
-	err = installErr
+	if !zerotouchremediation.IsComponentResolutionEnabled() {
+		err = nc.installHandler.Install()
+	} else {
+		var installErr error
+		defer func() {
+			if installErr != nil && nc.restoreResolution != nil {
+				err = errors.Join(err, nc.restoreResolution())
+			}
+		}()
+		installErr = nc.installHandler.Install()
+		err = installErr
+	}
 	if err != nil {
 		if !nc.disableCVSCheck && (nc.cmdName == "install" || nc.cmdName == "ci") {
 			if blockedErr := nc.handle404Errors(err); blockedErr != nil {
--- a/artifactory/commands/npm/npmcommand.go
+++ b/artifactory/commands/npm/npmcommand.go
@@ -26,8 +26,6 @@ import (
 	"github.com/jfrog/jfrog-client-go/utils/errorutils"
 	"github.com/jfrog/jfrog-client-go/utils/log"
 	"github.com/spf13/viper"
-
-	"github.com/jfrog/jfrog-cli-artifactory/artifactory/zerotouchremediation"
 )
 
 const (
@@ -357,23 +355,20 @@ func (nc *NpmCommand) Run() (err error) {
 	defer func() {
 		err = errors.Join(err, nc.installHandler.RestoreNpmrc())
 	}()
-	if !zerotouchremediation.IsComponentResolutionEnabled() {
-		err = nc.installHandler.Install()
-	} else {
-		var installErr error
-		defer func() {
-			if installErr != nil && nc.restoreResolution != nil {
-				err = errors.Join(err, nc.restoreResolution())
-			}
-		}()
-		installErr = nc.installHandler.Install()
-		err = installErr
-	}
-	if err != nil {
-		if !nc.disableCVSCheck && (nc.cmdName == "install" || nc.cmdName == "ci") {
-			if blockedErr := nc.handle404Errors(err); blockedErr != nil {
-				err = blockedErr
-			}
+	err = nc.installWithLockfileRestore()
+	return
+}
+
+func (nc *NpmCommand) installWithLockfileRestore() (err error) {
+	defer func() {
+		if err != nil && nc.restoreResolution != nil {
+			err = errors.Join(err, nc.restoreResolution())
+		}
+	}()
+	err = nc.installHandler.Install()
+	if err != nil && !nc.disableCVSCheck && (nc.cmdName == "install" || nc.cmdName == "ci") {
+		if blockedErr := nc.handle404Errors(err); blockedErr != nil {
+			err = blockedErr
 		}
 	}
 	return
@@ -525,19 +520,21 @@ func (nc *NpmCommand) prepareBuildInfoModule() error {
 	return nil
 }
 
-func (nc *NpmCommand) effectiveNpmCommand() string {
+func (nc *NpmCommand) dependencyCollectionArgs() []string {
+	npmArgs := nc.npmArgs
+	npmCommand := nc.cmdName
 	if nc.remediatedLockfile && nc.cmdName == "install" {
-		return "ci"
+		npmCommand = "ci"
+		npmArgs = stripNpmInstallOnlyArgs(npmArgs)
 	}
-	return nc.cmdName
+	return append([]string{npmCommand}, npmArgs...)
 }
 
 func (nc *NpmCommand) collectDependencies() error {
-	npmCommand := nc.effectiveNpmCommand()
-	if nc.remediatedLockfile {
+	if nc.remediatedLockfile && nc.cmdName == "install" {
 		log.Info("Using npm ci after Zero Touch Remediation to install from the remediated lockfile")
 	}
-	nc.buildInfoModule.SetNpmArgs(append([]string{npmCommand}, nc.npmArgs...))
+	nc.buildInfoModule.SetNpmArgs(nc.dependencyCollectionArgs())
 	return errorutils.CheckError(nc.buildInfoModule.Build())
 }
 
--- a/artifactory/commands/npm/npmcommand.go
+++ b/artifactory/commands/npm/npmcommand.go
@@ -2,7 +2,6 @@ package npm
 
 import (
 	"bufio"
-	"context"
 	"errors"
 	"fmt"
 	"net/http"
@@ -57,8 +56,9 @@ var (
 
 type NpmCommand struct {
 	CommonArgs
-	cmdName    string
-	jsonOutput bool
+	cmdName        string
+	jsonOutput     bool
+	executablePath string
 	// Function to be called to restore the user's old npmrc and delete the one we created.
 	restoreNpmrcFunc func() error
 	workingDirectory string
@@ -75,6 +75,8 @@ type NpmCommand struct {
 	installHandler      *NpmInstallStrategy
 	// When true, the subsequent install uses npm ci to honor remediated lockfile integrity.
 	remediatedLockfile bool
+	// Restores lockfiles written by Zero Touch Remediation if the install command fails.
+	restoreResolution func() error
 	// When true, skips the 404 error handling that checks if packages are blocked by curation
 	disableCVSCheck bool
 }
@@ -353,20 +355,10 @@ func (nc *NpmCommand) Run() (err error) {
 	defer func() {
 		err = errors.Join(err, nc.installHandler.RestoreNpmrc())
 	}()
-	if !nc.UseNative() {
-		if err = nc.CreateTempNpmrc(); err != nil {
-			return
-		}
-	}
-	var restoreResolution func() error
-	restoreResolution, nc.remediatedLockfile, err = nc.runZeroTouchRemediation(context.Background(), nc.cmdName, nc.workingDirectory, nc.npmArgs)
-	if err != nil {
-		return err
-	}
 	var installErr error
 	defer func() {
-		if installErr != nil && restoreResolution != nil {
-			err = errors.Join(err, restoreResolution())
+		if installErr != nil && nc.restoreResolution != nil {
+			err = errors.Join(err, nc.restoreResolution())
 		}
 	}()
 	installErr = nc.installHandler.Install()
@@ -536,7 +528,7 @@ func (nc *NpmCommand) effectiveNpmCommand() string {
 
 func (nc *NpmCommand) collectDependencies() error {
 	npmCommand := nc.effectiveNpmCommand()
-	if npmCommand != nc.cmdName {
+	if nc.remediatedLockfile {
 		log.Info("Using npm ci after Zero Touch Remediation to install from the remediated lockfile")
 	}
 	nc.buildInfoModule.SetNpmArgs(append([]string{npmCommand}, nc.npmArgs...))
--- a/artifactory/commands/npm/npmcommand.go
+++ b/artifactory/commands/npm/npmcommand.go
@@ -2,6 +2,7 @@ package npm
 
 import (
 	"bufio"
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
@@ -56,9 +57,8 @@ var (
 
 type NpmCommand struct {
 	CommonArgs
-	cmdName        string
-	jsonOutput     bool
-	executablePath string
+	cmdName    string
+	jsonOutput bool
 	// Function to be called to restore the user's old npmrc and delete the one we created.
 	restoreNpmrcFunc func() error
 	workingDirectory string
@@ -73,6 +73,8 @@ type NpmCommand struct {
 	collectBuildInfo    bool
 	buildInfoModule     *build.NpmModule
 	installHandler      *NpmInstallStrategy
+	// When true, the subsequent install uses npm ci to honor remediated lockfile integrity.
+	remediatedLockfile bool
 	// When true, skips the 404 error handling that checks if packages are blocked by curation
 	disableCVSCheck bool
 }
@@ -351,7 +353,24 @@ func (nc *NpmCommand) Run() (err error) {
 	defer func() {
 		err = errors.Join(err, nc.installHandler.RestoreNpmrc())
 	}()
-	err = nc.installHandler.Install()
+	if !nc.UseNative() {
+		if err = nc.CreateTempNpmrc(); err != nil {
+			return
+		}
+	}
+	var restoreResolution func() error
+	restoreResolution, nc.remediatedLockfile, err = nc.runZeroTouchRemediation(context.Background(), nc.cmdName, nc.workingDirectory, nc.npmArgs)
+	if err != nil {
+		return err
+	}
+	var installErr error
+	defer func() {
+		if installErr != nil && restoreResolution != nil {
+			err = errors.Join(err, restoreResolution())
+		}
+	}()
+	installErr = nc.installHandler.Install()
+	err = installErr
 	if err != nil {
 		if !nc.disableCVSCheck && (nc.cmdName == "install" || nc.cmdName == "ci") {
 			if blockedErr := nc.handle404Errors(err); blockedErr != nil {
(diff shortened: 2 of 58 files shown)


artifactory/commands/npm/common.go
```diff
@@ -14,6 +14,8 @@ type CommonArgs struct {
 	npmArgs            []string
 	serverDetails      *config.ServerDetails
 	useNative          bool
+	configFilePath     string
+	executablePath     string
 }
```
fluxxBot, line 17: this details are already part of other structs, please check if we can use that instead of defining it again here

artifactory/commands/npm/npmcommand.go
```diff
@@ -56,9 +57,8 @@ var (
 
 type NpmCommand struct {
 	CommonArgs
-	cmdName        string
-	jsonOutput     bool
-	executablePath string
```
fluxxBot, line 61: please be cautiuos here, make sure you're replacing all the caller as well who is using this executablePath, probably in build info go

artifactory/commands/npm/npmcommand.go
```diff
@@ -351,7 +353,24 @@ func (nc *NpmCommand) Run() (err error) {
 	defer func() {
 		err = errors.Join(err, nc.installHandler.RestoreNpmrc())
 	}()
-	err = nc.installHandler.Install()
+	if !nc.UseNative() {
+		if err = nc.CreateTempNpmrc(); err != nil {
+			return
+		}
+	}
```
fluxxBot: this should not be here, we already have strategy pattern to cater this requirement

artifactory/commands/npm/npmcommand.go
```diff
@@ -351,7 +353,24 @@ func (nc *NpmCommand) Run() (err error) {
...
+		return err
+	}
+	var installErr error
+	defer func() {
+		if installErr != nil && restoreResolution != nil {
+			err = errors.Join(err, restoreResolution())
+		}
+	}()
+	installErr = nc.installHandler.Install()
+	err = installErr
```
fluxxBot: I thought this would behind the checks, let's invoke this code only when the env is present and move it to a dedicated function

artifactory/commands/npm/npmcommand.go
```diff
@@ -508,8 +527,19 @@ func (nc *NpmCommand) prepareBuildInfoModule() error {
...
+		return "ci"
+	}
+	return nc.cmdName
+}
+
 func (nc *NpmCommand) collectDependencies() error {
-	nc.buildInfoModule.SetNpmArgs(append([]string{nc.cmdName}, nc.npmArgs...))
+	npmCommand := nc.effectiveNpmCommand()
+	if npmCommand != nc.cmdName {
+		log.Info("Using npm ci after Zero Touch Remediation to install from the remediated lockfile")
```
fluxxBot: let's use this bool here to check this condition - remediatedLockfile

artifactory/zerotouchremediation/buildtool.go
fluxxBot, line 1: please exclude files which are part of previous PRs, I think I saw this file in the earlier PRs

fluxxBot (changes requested):

artifactory/commands/npm/common.go
attiasas: Dropped configFilePath and executablePath from CommonArgs. Those already live on NpmCommand; ZTR now reads them from there instead of a second copy that SetConfigFilePath never populated.

artifactory/commands/npm/npmcommand.go
attiasas: Left executablePath on NpmCommand. It is unexported, so nothing outside this package (including build-info-go) reads it. The existing assignments in PreparePrerequisites / ConfigGet / CreateTempNpmrc are unchanged.

artifactory/commands/npm/npmcommand.go
attiasas: Removed the UseNative / CreateTempNpmrc branch from NpmCommand.Run. Temp .npmrc stays in npmRtInstall.Run; ZTR runs after that (and from npmInstall.Run for native).

artifactory/commands/npm/npmcommand.go
attiasas: applyZeroTouchRemediation returns immediately unless JFROG_CLI_ZTR_COMPONENTS_ENABLED is set, then the install strategies call it. NpmCommand.Run no longer enters the Xray path when the env is off.

artifactory/commands/npm/npmcommand.go
attiasas: The log now keys off remediatedLockfile instead of comparing the rewritten command name.

artifactory/zerotouchremediation/buildtool.go
attiasas: Rebased onto the updated #543 branch. artifactory/zerotouchremediation/* is no longer in this PR's diff.

artifactory/commands/npm/npmcommand.go
```diff
@@ -508,8 +519,19 @@ func (nc *NpmCommand) prepareBuildInfoModule() error {
 	return nil
 }
 
+func (nc *NpmCommand) effectiveNpmCommand() string {
```
fluxxBot: rename this function appropriately, can't make out why exactly this is required

artifactory/commands/npm/npmcommand.go
```diff
@@ -351,7 +355,14 @@ func (nc *NpmCommand) Run() (err error) {
 	defer func() {
 		err = errors.Join(err, nc.installHandler.RestoreNpmrc())
 	}()
-	err = nc.installHandler.Install()
+	var installErr error
+	defer func() {
+		if installErr != nil && nc.restoreResolution != nil {
+			err = errors.Join(err, nc.restoreResolution())
+		}
+	}()
```
fluxxBot: please offload this into a new function, let's keep main caller function clean

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,112 @@
...
+		return zerotouchremediation.SkipRemediation("Zero Touch Remediation skipped: could not create Xray service manager: ", xrayErr)
+	}
+	return zerotouchremediation.RunIfEnabled(ctx, xrayManager, resolverRepo, tool, command, workingDir, nc.npmBootstrapRunner(), bootstrapArgs...)
+}
+
+// resolverRepoForResolution returns the Artifactory virtual repo for dependency policy scope.
+func (nc *NpmCommand) resolverRepoForResolution(command string) (string, error) {
+	if command != "publish" && nc.repo != "" {
+		return nc.repo, nil
+	}
```
fluxxBot: this check doesn't belong here, incase you need this check, put it in the correct place so that this can be taken care by the strategy pattern already used

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		return resolverConfig.TargetRepo(), nil
+	}
+	if nc.executablePath != "" {
+		registryURL, err := nc.getNpmRegistryURL()
+		if err != nil {
+			return "", fmt.Errorf("failed to get registry URL: %w", err)
+		}
+		return extractRepoName(registryURL)
+	}
```
fluxxBot: this should be in the else part, also put this above function behind a interface in the pattern that we're already checking, i.e. artifactoryInstall and npmInstall.
Also I think there must be some existing function which might give you this result.

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,112 @@
...
+		Executable: nc.executablePath,
+		CmdName:    "config",
+		CmdArgs:    []string{"get", "registry"},
+	}
+	data, err := configCommand.RunWithOutput()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(data)), nil
+}
```
fluxxBot: please check for duplicate function

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (nc *NpmCommand) npmBootstrapRunner() zerotouchremediation.CommandRunner {
+	return func(ctx context.Context, projectRoot string, args ...string) error {
+		return runNpmAt(ctx, nc.executablePath, projectRoot, args...)
+	}
+}
```
fluxxBot: why is this fucntion required, add comments

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,112 @@
...
+		return zerotouchremediation.SkipRemediation("Zero Touch Remediation skipped: could not determine resolver repo: ", resolverErr)
+	}
+	if resolverRepo == "" {
+		return zerotouchremediation.SkipRemediation("Zero Touch Remediation skipped: resolver repo is empty", nil)
+	}
+	var projectKey string
+	if nc.buildConfiguration != nil {
+		projectKey = nc.buildConfiguration.GetProject()
+	}
+	xrayManager, xrayErr := xray.CreateXrayServiceManager(nc.serverDetails, xray.WithScopedProjectKey(projectKey))
```
fluxxBot, line 46: check if we can cache this response, otherwise we might be hitting again and again for the same dependency with same version

artifactory/commands/npm/zerotouchremediation.go
fluxxBot: Also, there can be multiple resolution repositories incase of run native mode, dependening on scopes, please take care of that as well

artifactory/commands/npm/ztr_buildtool.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (BuildTool) ToolName() string { return toolName }
+
+func (BuildTool) RelevantCommands() []string {
+	return []string{"install", "ci", "publish"}
+}
+
+func (t BuildTool) ProjectRoot(workingDir string) (string, error) {
+	return discoverProjectRootWithOptions(workingDir, t.opts)
```
fluxxBot, line 31: redundant calls here - https://github.com/jfrog/jfrog-cli-artifactory/pull/544/changes#diff-a19db88699b622b212960c46da6c9f6acdcd21f93a1f1d7166e1058831b5ccd5R78

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,112 @@
...
+		registryURL, err := nc.getNpmRegistryURL()
+		if err != nil {
+			return "", fmt.Errorf("failed to get registry URL: %w", err)
+		}
+		return extractRepoName(registryURL)
+	}
+	return nc.repo, nil
+}
+
+func (nc *NpmCommand) getNpmRegistryURL() (string, error) {
```
fluxxBot: this must be already populated, out of this function is not required here. Also it spawns up npm subprocess that takes considerable time, try to use the value already available

artifactory/commands/npm/ztr_buildtool.go
```diff
@@ -0,0 +1,91 @@
...
+func NewBuildTool() BuildTool {
+	return BuildTool{}
+}
+
+func NewBuildToolWithArgs(npmArgs []string) BuildTool {
+	parsed := parseNpmCLIArgs(npmArgs)
+	return BuildTool{opts: discoveryOptions{prefixDir: parsed.prefixDir}}
+}
+
+func NewBuildToolForPublish(workingDir, publishPath string, npmArgs []string) BuildTool {
```
fluxxBot: this seems like a dead code

artifactory/commands/npm/npmcommand.go
```diff
@@ -508,8 +519,19 @@ func (nc *NpmCommand) prepareBuildInfoModule() error {
...
+	return nc.cmdName
+}
+
 func (nc *NpmCommand) collectDependencies() error {
-	nc.buildInfoModule.SetNpmArgs(append([]string{nc.cmdName}, nc.npmArgs...))
+	npmCommand := nc.effectiveNpmCommand()
+	if nc.remediatedLockfile {
+		log.Info("Using npm ci after Zero Touch Remediation to install from the remediated lockfile")
+	}
+	nc.buildInfoModule.SetNpmArgs(append([]string{npmCommand}, nc.npmArgs...))
```
fluxxBot: When `remediatedLockfile=true`, `npmCommand` becomes `"ci"` but `nc.npmArgs` is
forwarded unchanged. `npm ci` silently ignores or warns on install-only flags
like `--save`, `--save-dev`, `--no-save`, `--include`. Not a fatal error, but
could produce confusing output for users.

Consider stripping install-only flags before forwarding to `ci`:

    args := nc.npmArgs
    if npmCommand == "ci" {
        args = stripInstallOnlyFlags(args)
    }
    nc.buildInfoModule.SetNpmArgs(append([]string{npmCommand}, args...))

fluxxBot (changes requested):

artifactory/commands/npm/npmcommand.go
attiasas: Renamed to dependencyCollectionArgs. It is the argv for build-info collection, not the install command itself: after a remediating lockfile write, that collection run is npm ci and install-only flags are stripped so ci does not warn on --save / --package-lock-only.

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,112 @@
...
+			return "", fmt.Errorf("failed to read config file: %w", err)
+		}
+		resolverConfig, err := project.GetRepoConfigByPrefix(nc.configFilePath, project.ProjectConfigResolverPrefix, vConfig)
+		if err != nil {
+			return "", fmt.Errorf("failed to get resolver config: %w", err)
+		}
+		return resolverConfig.TargetRepo(), nil
+	}
+	if nc.executablePath != "" {
+		registryURL, err := nc.getNpmRegistryURL()
```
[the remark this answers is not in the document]
attiasas: CLI --registry is parsed from npmArgs and is the first source in resolverRepoForResolution. It wins over nc.repo and over npm config. Covered by TestResolverRepoForResolution_PrefersCliRegistryOverNpmConfig and TestResolverRepoForResolution_CliRegistryOverridesRepo.

artifactory/commands/npm/npmcommand.go
attiasas: Moved into installWithLockfileRestore. Run now only prepares RestoreNpmrc and calls that method. Restore still joins onto the install error, including handle404Errors.

artifactory/commands/npm/zerotouchremediation.go
attiasas: Removed. RelevantCommands is install and ci only. publish is out of this PR.

artifactory/commands/npm/zerotouchremediation.go
attiasas: Kept on NpmCommand because both install strategies share the same resolver-repo question. Order is now: explicit --registry, then nc.repo (already set for artifactory install), then npmUtils.ConfigGet only when both are empty (native). No second interface on the install strategies.

artifactory/commands/npm/zerotouchremediation.go
attiasas: Removed getNpmRegistryURL. Native fallback uses npmUtils.ConfigGet, which already exists for this.

artifactory/commands/npm/zerotouchremediation.go
attiasas: Removed. Bootstrap now calls biUtils.RunNpmCmd directly from runZeroTouchRemediation.

artifactory/commands/npm/zerotouchremediation.go
attiasas: Not adding a cache here. Each enabled install sends the current lockfile once; Xray policy can change between runs, and a content cache would be a new persistence layer in the generic service, not in the npm wiring. Disabled installs never call Xray.

artifactory/commands/npm/zerotouchremediation.go
attiasas: Xray scopes one lockfile to one repo. Native @scope:registry maps are not a second policy input. The repo we send is --registry, else the configured resolver, else npm's default registry. Multi-repo native remains a limitation of that API shape.

artifactory/commands/npm/ztr_buildtool.go
attiasas: DiscoverLockfiles now takes the already-resolved projectRoot. The generic flow walks once in ProjectRoot and passes that path in, so npm does not walk again.

artifactory/commands/npm/zerotouchremediation.go
attiasas: ConfigGet only runs when --registry and nc.repo are both empty. Artifactory install already has nc.repo from the temp .npmrc and never shells out here.

artifactory/commands/npm/ztr_buildtool.go
attiasas: Removed NewBuildToolForPublish and NewBuildToolWithArgs. The tool is constructed in runZeroTouchRemediation with prefixDir from the parsed args.

artifactory/commands/npm/npmcommand.go
attiasas: When remediatedLockfile is set, dependencyCollectionArgs switches to ci and runs stripNpmInstallOnlyArgs. --save / --no-save / --package-lock-only and similar are dropped. --omit, --include, --workspace, and --registry are kept so ci resolves the same tree. --include is valid on ci; it was not stripped.

artifactory/commands/npm/ztr_args_parse.go
```diff
@@ -0,0 +1,125 @@
...
+			out.registryURL = strings.TrimPrefix(arg, "--registry=")
+			out.bootstrapArgs = append(out.bootstrapArgs, arg)
+		case arg == "--tag" || arg == "--omit" || arg == "--include" || arg == "--save-prefix":
+			if i+1 < len(args) && !strings.HasPrefix(args[i+1], "-") {
+				i++
+			}
+		case strings.HasPrefix(arg, "-"):
+			continue
+		default:
+			out.packageOperands = append(out.packageOperands, arg)
```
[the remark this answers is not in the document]
attiasas: Leaving this as-is. The parser only consumes flags that change ZTR (prefix, registry, workspace, omit/include). Unknown flags are ignored so a new npm flag cannot break the install. A value like verbose after --loglevel can look like a package operand and skip ZTR; that skip is conservative. Listing every npm config key is unbounded and still incomplete.

artifactory/commands/npm/ztr_discover.go
```diff
@@ -0,0 +1,140 @@
...
+				if pkg, parseErr := parsePackageJSON(data); parseErr == nil && pkg.hasWorkspaces() {
+					topWorkspaceRoot = dir
+				}
+			}
+		}
+		if _, statErr := os.Stat(filepath.Join(dir, shrinkwrapFileName)); statErr == nil {
+			return dir, nil
+		}
+		if _, statErr := os.Stat(filepath.Join(dir, lockfileName)); statErr == nil {
+			return dir, nil
```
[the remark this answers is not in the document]
attiasas: Lockfile wins on purpose. That file is what the following npm command installs from. TestDiscoverProjectRoot_FromWorkspacePackage already walks to the workspace root when the member has no lockfile. Preferring a parent workspaces package.json over a nested lockfile would remediate a different file than the install uses (TestDiscoverProjectRoot_IndependentPackageLock).

artifactory/zerotouchremediation/buildtool.go
attiasas: Follow-up: DiscoverLockfiles now takes projectRoot instead of workingDir, so npm does not discover twice. That is the only generic-package delta vs #543.

artifactory/commands/npm/npmcommand.go
fluxxBot: can we rename it to - remediatedNpmCommand or something, so that we know where is this coming from?

artifactory/commands/npm/zerotouchremediation.go
fluxxBot: but then this will resolve on the wrong repo.
in a single npmrc file it can have registry as well as @scope:registry

attiasas: Native `registry` plus `@scope:registry` can name different Artifactory repos; `npm config get registry` only sees the default, so that was the wrong `repo` for Xray.

Xray still accepts one `repo` per lockfile. Native resolution now reads `npm c ls` for both the default registry and every `@scope:registry`, keeps Artifactory npm URLs, and sends that repo when there is exactly one. If two or more distinct Artifactory repos are configured, ZTR is skipped instead of guessing a policy. Project-config installs still use `nc.repo` (temp `.npmrc` already rewrites scopes to that registry).

artifactory/commands/npm/zerotouchremediation.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func npmConfigRegistryURLs(cliRegistryURL string, configList []byte) []string {
+	cliOverridesDefault := cliRegistryURL != ""
+	var urls []string
+	if cliOverridesDefault {
+		urls = append(urls, cliRegistryURL)
+	}
+	scanner := bufio.NewScanner(strings.NewReader(string(configList)))
```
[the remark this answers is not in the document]
attiasas: `npmConfigRegistryURLs` no longer uses `bufio.Scanner` (64 KiB token cap, `Err()` ignored). It splits the in-memory `npm c ls` output on newlines so a long non-registry line cannot truncate later `@scope:registry` entries.

`TestResolverRepoFromNpmConfig` covers an oversized line between two distinct Artifactory registries; that case now errors on multiple repos instead of returning the first one.

artifactory/commands/npm/npmcommand.go
attiasas: Leaving `dependencyCollectionArgs`. It is the argv passed to build-info collection on every install/ci, including when `remediatedLockfile` is false.

`remediatedNpmCommand` would read as the install command after ZTR, and as a command name rather than a `[]string`. The ZTR branch is already explicit: `if nc.remediatedLockfile && nc.cmdName == "install"` switches to `ci` and strips install-only flags.

If the caller should be in the name, `collectDependenciesNpmArgs` is the accurate alternative.

fluxxBot (approved):

bhanurp (approved):

PR #548, opened 2026-09-01
[PM Fixer] Fix huggingface 1.29.0 compatibility: swallow HfUriError from commitUrl parsing

agrasth:
## Summary

`upload_folder` in huggingface_hub ≥ 1.20.0 fails against Artifactory with `HfUriError` (introduced in [huggingface/huggingface_hub#4324](https://github.com/huggingface/huggingface_hub/pull/4324)). That PR changed `RepoUrl` to use the strict `parse_hf_uri` internally when processing the commit response. Artifactory returns a placeholder `commitUrl` that does not conform to the `hf://` URI format the new parser expects. The actual upload to Artifactory completes successfully before the response is parsed — only the post-upload parsing step raises the error, causing all tests that call `uploadTestModelToLocalRepo` (and `TestHuggingFaceUploadWithLocalGitVcsProps`) to fail with `require.NoError`.

## Fix

Modified `huggingface_upload.py` (embedded Python helper for `jf hf u`) to catch `HfUriError` from `huggingface_hub.errors` specifically and return silently, treating the upload as successful. All other exceptions are still propagated.

This is a **Python-script-only fix** — no Go source files were modified and no exported symbols changed.

Backward compatibility:
- huggingface_hub < 1.20.0: `HfUriError` does not exist; the `except ImportError` branch is taken, behavior unchanged.
- huggingface_hub ≥ 1.20.0: `HfUriError` raised during post-upload response parsing is swallowed; HTTP errors, auth failures, and all other exceptions are still propagated.

## Compatibility impact

No exported Go symbol was added, removed, or renamed. No `go.mod` or `go.sum` touched. Existing supported versions (pinned at 1.19.0) are unaffected since `HfUriError` did not exist before 1.20.0.

## Test plan

- `go build ./...` — no Go source files changed; cannot regress.
- `gofmt -l .` — no Go files touched.
- `go vet ./...` — same.
- `TestExtractPythonScripts` (unit test): verifies scripts are extracted; content checking is not in scope.
- Live PM integration test with huggingface_hub 1.29.0 against Artifactory **could not be run in this environment** (no live Artifactory). The fix targets the exact failure documented in `.github/workflows/huggingfaceTests.yml` lines 69-72.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hugging Face uploads now complete successfully when a malformed full URL is reported after files are committed.
  * Handled errors include diagnostic details to support troubleshooting.
  * Other failures, including malformed identifiers without a full URL, continue to be reported normally.

* **Tests**
  * Added coverage for handled post-upload URL errors and unrelated failures.
  * Integration tests now run against multiple supported Hugging Face Hub versions.

---
_Opened automatically by [PM Compat Fixer](https://github.jfrog.info/JFROG/jfrog-cli-workflows/actions/runs/1480349) investigating huggingface 1.29.0._

### DIFF
--- a/artifactory/commands/huggingface/huggingface_upload.py
+++ b/artifactory/commands/huggingface/huggingface_upload.py
@@ -26,10 +26,23 @@ def upload(folder_path, repo_id, repo_type, revision, **kwargs):
         >>> upload_model("/path/to/dataset", "username/dataset-name", repo_type="dataset")
     """
     api = HfApi()
-    api.upload_folder(
-        folder_path=folder_path,
-        repo_id=repo_id,
-        revision=revision,
-        repo_type=repo_type,
-        **kwargs
-    )
+    try:
+        api.upload_folder(
+            folder_path=folder_path,
+            repo_id=repo_id,
+            revision=revision,
+            repo_type=repo_type,
+            **kwargs
+        )
+    except Exception as e:
+        # huggingface_hub >= 1.20.0 uses strict parse_hf_uri inside RepoUrl (PR #4324).
+        # Artifactory returns a placeholder commitUrl that fails this parser, but the
+        # files were already committed before the response is parsed. Swallow the URI
+        # parse error so callers treat the upload as successful.
+        try:
+            from huggingface_hub.errors import HfUriError
+            if isinstance(e, HfUriError):
+                return
+        except ImportError:
+            pass
+        raise
--- a/artifactory/commands/huggingface/huggingface_upload.py
+++ b/artifactory/commands/huggingface/huggingface_upload.py
@@ -5,6 +5,8 @@ This module provides functionality to upload models and datasets to HuggingFace 
 using HfApi with configurable parameters.
 """
 
+import sys
+
 from huggingface_hub import HfApi
 
 
@@ -42,6 +44,18 @@ def upload(folder_path, repo_id, repo_type, revision, **kwargs):
         try:
             from huggingface_hub.errors import HfUriError
             if isinstance(e, HfUriError):
+                # Not silent: this is a real behavior change (an upload failure becomes a
+                # success), so anyone debugging a "phantom" success needs a trail. This is
+                # HfUriError specifically from parsing the upload response - printed to
+                # stderr, which the Go caller already streams through unmodified (see
+                # huggingFaceUpload.go's cmd.Stderr), so it won't affect the stdout JSON
+                # success/failure contract the caller actually parses.
+                print(
+                    f"jf hf upload: swallowed HfUriError while parsing the commit response for "
+                    f"'{repo_id}' - files were already uploaded successfully; only the response "
+                    f"URL failed to parse ({e})",
+                    file=sys.stderr,
+                )
                 return
         except ImportError:
             pass
--- a/artifactory/commands/huggingface/test_huggingface_upload.py
+++ b/artifactory/commands/huggingface/test_huggingface_upload.py
@@ -1,0 +1,58 @@
+"""
+Tests for huggingface_upload.py's HfUriError-swallowing behavior.
+
+Context: huggingface_hub >= 1.20.0 raises HfUriError from HfApi.upload_folder()
+when parsing the commit response's URL (RepoUrl parsing, PR #4324 upstream).
+Artifactory returns a placeholder commitUrl that this stricter parser rejects,
+even though the files were already committed before the response is parsed.
+upload() swallows that specific error and treats the upload as successful;
+every other exception must still propagate.
+
+huggingface_hub.errors.HfUriError may not exist in whatever huggingface_hub
+version happens to be installed wherever this test runs (it's a recent
+addition). Patching it with create=True injects a stand-in at the exact
+import path upload()'s except block reads from, so this test exercises the
+real isinstance() check in huggingface_upload.py regardless of the installed
+huggingface_hub version - it is not a copy of the production logic.
+"""
+import unittest
+from unittest.mock import patch
+
+import huggingface_upload
+
+
+class FakeHfUriError(ValueError):
+    """Stand-in for huggingface_hub.errors.HfUriError (same constructor shape)."""
+
+    def __init__(self, uri, msg):
+        self.uri = uri
+        self.msg = msg
+        super().__init__(f"Invalid HF URI '{uri}'. {msg}")
+
+
+class UploadHfUriErrorTest(unittest.TestCase):
+    @patch("huggingface_hub.errors.HfUriError", FakeHfUriError, create=True)
+    @patch("huggingface_upload.HfApi")
+    def test_swallows_hf_uri_error_after_successful_upload(self, mock_hf_api_class):
+        mock_hf_api_class.return_value.upload_folder.side_effect = FakeHfUriError(
+            uri="https://artifactory.example/placeholder", msg="could not parse"
+        )
+
+        # Must return cleanly (no exception) - the files were already committed,
+        # only the response URL failed to parse.
+        huggingface_upload.upload(
+            folder_path="/tmp/folder", repo_id="org/repo", repo_type="model", revision="main"
+        )
+
+    @patch("huggingface_upload.HfApi")
+    def test_other_exceptions_still_propagate(self, mock_hf_api_class):
+        mock_hf_api_class.return_value.upload_folder.side_effect = ValueError("network error")
+
+        with self.assertRaises(ValueError):
+            huggingface_upload.upload(
+                folder_path="/tmp/folder", repo_id="org/repo", repo_type="model", revision="main"
+            )
+
+
+if __name__ == "__main__":
+    unittest.main()


itsmeleela (dismissed):

artifactory/commands/huggingface/huggingface_upload.py
```diff
@@ -26,10 +26,23 @@ def upload(folder_path, repo_id, repo_type, revision, **kwargs):
...
\ No newline at end of file
+    try:
+        api.upload_folder(
+            folder_path=folder_path,
+            repo_id=repo_id,
+            revision=revision,
+            repo_type=repo_type,
+            **kwargs
+        )
+    except Exception as e:
```
itsmeleela, line 39: Swallowing based on exception type assumes HfUriError can only occur post-commit in this call path. Can we pin this to the known-affected version range, or at least log when we swallow it, so we're not silently masking a genuine failure if that assumption stops holding in a future huggingface_hub release?

artifactory/commands/huggingface/huggingface_upload.py
```diff
@@ -26,10 +26,23 @@ def upload(folder_path, repo_id, repo_type, revision, **kwargs):
...
+        # huggingface_hub >= 1.20.0 uses strict parse_hf_uri inside RepoUrl (PR #4324).
+        # Artifactory returns a placeholder commitUrl that fails this parser, but the
+        # files were already committed before the response is parsed. Swallow the URI
+        # parse error so callers treat the upload as successful.
+        try:
```
itsmeleela, line 44: Do we already have tests that mocks upload_folder to raise HfUriError and asserts we return cleanly, plus one asserting a different exception still propagates? This is the core behavior change so I think fixer should consider adding unit tests for code changes

artifactory/commands/huggingface/huggingface_upload.py
agrasth: Added in 64c51f5: now logs a clear message to stderr when the HfUriError swallow triggers (files were already uploaded, only the response URL failed to parse). 
Printed to stderr specifically since the Go caller (huggingFaceUpload.go) already streams stderr through unmodified without touching the stdout JSON success/failure contract it parses — so this doesn't risk interfering with that. Didn't add version-pinning on top since the affected range is only knowable approximately and would need its own upkeep; the log line covers the "don't silently mask a future failure" concern more durably.

artifactory/commands/huggingface/huggingface_upload.py
agrasth: Added in 64c51f5: test_huggingface_upload.py with exactly the two cases you described — one mocking upload_folder to raise HfUriError and asserting upload() returns cleanly, one mocking a different exception and asserting it still propagates.

naveenku-jfrog (commented):
Thanks for the write-up — the diagnosis matches upstream. `huggingface_hub` ≥ 1.20.0 builds `CommitInfo.repo_url` from `commitUrl` via `RepoUrl` / `parse_hf_uri` (huggingface/huggingface_hub#4324), and Artifactory still returns a placeholder, so `upload_folder` raises `HfUriError` after the commit POST. Catching that in the embedded helper is a reasonable CLI-side shim, and logging it on stderr is the right channel: `huggingFaceUpload.go` parses stdout JSON and already streams stderr through.

A few things I would still change before treating this as done:

**Narrow the swallow.** `HfUriError` is a generic URI parser (`ValueError`), not “commit already landed.” On today’s `upload_folder` path it almost always comes from `CommitInfo.__post_init__` after `_send_commit`, but if a later hub release (or the `hf_xet` pipelined path) raises it earlier, `jf hf u` will report success and `CollectArtifactsForBuildInfo` can quietly find nothing. Please only swallow when `isinstance(e, HfUriError)` *and* the URI looks like the known post-commit case (e.g. `"/commit/"` in `e.uri`, or an Artifactory HTTP URL rather than a bare repo id). Re-raise everything else.

**1.29.0 is not actually exercised.** Live Hugging Face tests live in jfrog-cli and still pin `huggingface_hub==1.19.0` (`huggingfaceTests.yml` / `setup-test-deps.sh`). This PR does not unpin that, and a live Artifactory run on 1.29.0 is still missing. Until someone runs `jf hf u` against Artifactory on hub 1.29.0 and then bumps that pin, this remains an unverified workaround.

**The new Python tests never run in this repo’s CI.** `test_huggingface_upload.py` is the right coverage (swallow vs propagate), but jfrog-cli-artifactory has no pytest/unittest job, and `TestExtractPythonScripts` only checks that the embed writes a non-empty file. Either add a Python test step, or these tests will rot. Worth also asserting the stderr log.

This is also not the durable fix — Artifactory should return a Hub-shaped `commitUrl` that `parse_hf_uri` accepts with a custom endpoint. Fine as a CLI shim, but that server-side gap should be tracked separately so this catch does not have to live forever.

artifactory/commands/huggingface/huggingface_upload.py
```diff
@@ -26,10 +28,35 @@ def upload(folder_path, repo_id, repo_type, revision, **kwargs):
...
+            from huggingface_hub.errors import HfUriError
+            if isinstance(e, HfUriError):
```
naveenku-jfrog, line 46: This treats every `HfUriError` from `upload_folder` as “files already committed.” That is true today for `CommitInfo` parsing a placeholder `commitUrl`, but `HfUriError` is not scoped to that. Please gate on the URI as well (`"/commit/"` in `e.uri`, or an HTTP Artifactory URL) and re-raise otherwise, so we do not turn a pre-commit parse failure into a false success.

artifactory/commands/huggingface/test_huggingface_upload.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, uri, msg):
+        self.uri = uri
+        self.msg = msg
+        super().__init__(f"Invalid HF URI '{uri}'. {msg}")
+
+
+class UploadHfUriErrorTest(unittest.TestCase):
+    @patch("huggingface_hub.errors.HfUriError", FakeHfUriError, create=True)
+    @patch("huggingface_upload.HfApi")
+    def test_swallows_hf_uri_error_after_successful_upload(self, mock_hf_api_class):
```
naveenku-jfrog, line 36: Good split between swallow vs propagate, and the `create=True` patch is the right way to avoid depending on an installed hub version.

These tests are not run by this repo’s CI (no pytest/unittest job; `TestExtractPythonScripts` only checks embed extraction). Please either add a Python test step, or they will not actually protect the change. Also assert the stderr log, and add a case where `HfUriError` is raised with a non-commit URI so the tighter swallow (if added) stays honest.

artifactory/commands/huggingface/huggingface_upload.py
agrasth: Scoped the swallow further in 4ca1b2a: now also checks that `e.uri` contains "://" before swallowing. Traced huggingface_hub's PR #4324 - `RepoUrl`/`parse_hf_uri` is only invoked from `CommitInfo.__post_init__` on `commit_url.split("/commit/")[0]`, which is always a full URL; a bare `repo_id`/`revision` string passed to `upload_folder` can never contain "://". So an `HfUriError` raised over a plain identifier (e.g. a malformed `repo_id` rejected before any commit) now still propagates instead of being swallowed just because the exception type matches. Didn't gate on `"/commit/" in e.uri` literally, since `CommitInfo` already strips everything before `/commit/` before parsing - so that substring wouldn't actually be present in `e.uri` at the point we catch it. Added a test asserting a non-URI `HfUriError` still propagates.

artifactory/commands/huggingface/test_huggingface_upload.py
agrasth: Added a step in 4ca1b2a that runs `python -m unittest test_huggingface_upload` inside the HuggingFace-Tests CI job - previously these tests weren't run anywhere. Also added: an assertion on the stderr log content in the swallow test, and a new case asserting HfUriError over a non-URI (plain repo_id-shaped) uri still propagates, given the tighter swallow condition added above.

.github/workflows/cli-integration-tests.yml
```diff
@@ -1241,6 +1241,11 @@ jobs:
         shell: bash
         run: pip install "huggingface_hub==1.19.0"
 
+      - name: Run HuggingFace upload script unit tests
+        shell: bash
+        working-directory: artifactory/commands/huggingface
+        run: python -m unittest test_huggingface_upload -v
```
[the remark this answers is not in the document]
agrasth: Fixed in 0994472. Verified against the real huggingface_hub source at tag v1.19.0: `HfUriError` already exists there with an identical uri/msg constructor - what's actually version-gated is PR #4324's new RepoUrl call site, not the exception class itself. Removed `FakeHfUriError`/`create=True` and now import+raise the real `huggingface_hub.errors.HfUriError` in both tests. Didn't add a separate 1.20.0+ pass since these are unit tests of upload()'s own except-block logic (driven by a mocked upload_folder side_effect) - they don't depend on which huggingface_hub version actually raises HfUriError from RepoUrl parsing internally.

naveenku-jfrog (dismissed):
Please add test to run on Hub 1.29.0.

naveenku-jfrog (dismissed):

fluxxBot (dismissed):

fluxxBot (approved):

naveenku-jfrog (approved):

PR #552, opened 2026-09-07
RTECO-1782: let the native client publish, and keep credentials off disk

bhanurp:
## What

Three changes to the NuGet/dotnet FlexPack command, all narrowing what jf does on the user's behalf.

## 1. Let the native client perform the upload

FlexPack's contract is that the native tool does the work and jf observes it — but `push` was intercepted and routed through the Artifactory upload service instead. The comment justifying that cited a 401 from `dotnet nuget push` against a V3 `index.json`.

**That justification doesn't hold.** The 401 is specific to credentials carried *in the source URL*, where the service-index fetch goes out unauthenticated. Supplying them through a config file avoids it entirely. Tested against Artifactory before changing anything:

| Client | Method | Result |
|---|---|---|
| `dotnet` SDK 10.0.302 | source URL with embedded creds | ❌ 401 |
| `dotnet` SDK 10.0.302 | `<packageSourceCredentials>` in config | ✅ pushed |
| `nuget.exe` 6.6.2 | `-Source <v3 index> -ApiKey user:token` | ❌ 401 |
| `nuget.exe` 6.6.2 | `<packageSourceCredentials>` in config | ✅ pushed |

Both clients push fine. The bypass and its **seven now-unreachable helpers** are removed (−318 lines), and the `if/else` dispatch collapses to "run the native client".

A user's own `-Source`/`-ApiKey` still wins, unchanged.

## 2. Credentials travel in the environment, not on disk

The temp `nuget.config` no longer carries a `<packageSourceCredentials>` block. The native client reads `NuGetPackageSourceCredentials_<source>` from its environment instead.

This removes the **persistence** risk — a signal that skips cleanup can no longer strand a token in a file — while keeping the secret out of `argv`, which is world-readable via `ps`.

It is **not** secrecy: the value is still readable by same-user processes (`ps -E`, `/proc/<pid>/environ`). The code comment says exactly that rather than overclaiming.

The config now also carries `defaultPushSource`, so the push finds its target without jf appending `-Source`/`--source` to the user's command line. Keeping the target in configuration rather than argv means the native client is invoked exactly as the user wrote it, and avoids branching on per-toolchain flag spelling.

## 3. Stamp `vcs.*` / `ci.*` on pushed artifacts

Push recorded only `build.name`, `build.number`, `build.timestamp` — so an artifact knew which build produced it, but not which commit, branch or pipeline run.

Routing through `civcs.MergeWithUserProps` (the helper Terraform and the other FlexPack managers already use) closes the gap. Verified live:

```
build.name = [fix4-verify]      vcs.branch   = [master]
build.number = [1]              vcs.revision = [394c1726...]
build.timestamp = [...]         vcs.url      = [https://github.com/...]
```

No-op outside a repository or when the properties are disabled.

## One ordering trap worth noting

`shouldPushViaNativeClient()` is computed **before** credential injection. Injection appends to `c.args`, and `hasNativeAuthOverride` treats a `--source` as a *user* override — so evaluating afterwards would make jf misread its own flag as user intent. The comment records why.

## Testing

- `TestShouldPushViaNativeClient` — 5 cases: both toolchains native, user auth override respected, non-push subcommands unaffected, missing server/repo falls through
- `TestCredentialEnvEntry` — format, source-name keying (a mismatch silently 401s), verbatim token passthrough
- `TestTempConfigCarriesNoSecret` — writes a real temp config with a known password and asserts the file contains **no** password, no `ClearTextPassword`, no `packageSourceCredentials`, while still declaring the source; then that cleanup removes the file and clears the credential
- `gofmt`, `go build ./...`, `go vet`, `golangci-lint` — clean, including `unused` after the deletions
- Verified end-to-end: `jf dotnet nuget push` and `jf nuget push` both upload via their native client with build-info and properties intact

> **Note:** `gosec` could not be run — `internal error: package "fmt" without types` under this Go toolchain. Unverified rather than passing.

## Tradeoff being accepted

The removed upload-service path provided checksum-optimised deploys (skip transfer when the blob exists), jf's retry logic, and JFrog proxy handling. Native push has none of these — it always uploads the bytes. For large or frequently re-pushed packages that is a real difference, and it is a deliberate choice in favour of not interfering with the native client.

## Merge order

Second of three RTECO-1782 PRs.

1. jfrog/build-info-go#422
2. **jfrog-cli-artifactory** ← this PR
3. jfrog-cli

## Summary by CodeRabbit

- **New Features**
  - NuGet restore operations triggered by `pack` and `publish` now apply supported temporary configuration and credentials.
  - Package collection includes nested target project or solution directories when no output directory is specified.
  - NuGet package pushes use the native `nuget` or `dotnet` client by default.
  - Artifactory sources, authentication, build information, and CI/version-control details are configured automatically.

- **Bug Fixes**
  - Temporary configuration files avoid storing secrets and are cleaned up after operations.
  - User-provided configuration files are detected and preserved.
  - Push commands correctly handle options placed after `--`.

### DIFF
--- a/artifactory/commands/nuget/command.go
+++ b/artifactory/commands/nuget/command.go
@@ -355,16 +355,23 @@ func (c *NuGetFlexPackCommand) injectCredentialsViaTempConfig(repo string) (func
 // The flag belongs to the dotnet command itself, so it has to precede the separator. Only the
 // first "--" is meaningful; anything after it is the user's payload and is left untouched.
 func insertBeforeSeparator(args []string, extra ...string) []string {
+	// Insert before the first bare "--", or at the end when there is none.
+	at := len(args)
 	for i, arg := range args {
 		if arg == "--" {
-			combined := make([]string, 0, len(args)+len(extra))
-			combined = append(combined, args[:i]...)
-			combined = append(combined, extra...)
-			combined = append(combined, args[i:]...)
-			return combined
+			at = i
+			break
 		}
 	}
-	return append(args, extra...)
+	// Always build a fresh slice. "append(args, extra...)" for the no-separator case wrote into
+	// the caller's backing array whenever it had spare capacity, which contradicts this
+	// function's contract: the caller keeps a saved reference to restore c.args from after the
+	// command runs.
+	combined := make([]string, 0, len(args)+len(extra))
+	combined = append(combined, args[:at]...)
+	combined = append(combined, extra...)
+	combined = append(combined, args[at:]...)
+	return combined
 }
 
 // credentialEnvEntry builds the NuGet environment-variable credential entry for a package
--- a/artifactory/commands/nuget/command.go
+++ b/artifactory/commands/nuget/command.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/xml"
 	"fmt"
-	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -17,13 +16,12 @@ import (
 	buildinfoflex "github.com/jfrog/build-info-go/flexpack"
 	nugetflex "github.com/jfrog/build-info-go/flexpack/nuget"
 	"github.com/jfrog/jfrog-cli-artifactory/artifactory/commands/generic"
+	"github.com/jfrog/jfrog-cli-artifactory/artifactory/utils/civcs"
 	rtutils "github.com/jfrog/jfrog-cli-core/v2/artifactory/utils"
 	buildUtils "github.com/jfrog/jfrog-cli-core/v2/common/build"
 	"github.com/jfrog/jfrog-cli-core/v2/common/spec"
 	"github.com/jfrog/jfrog-cli-core/v2/utils/config"
-	"github.com/jfrog/jfrog-client-go/artifactory"
 	"github.com/jfrog/jfrog-client-go/artifactory/services"
-	specutils "github.com/jfrog/jfrog-client-go/artifactory/services/utils"
 	"github.com/jfrog/jfrog-client-go/utils/io/content"
 	"github.com/jfrog/jfrog-client-go/utils/log"
 )
@@ -40,6 +38,10 @@ type NuGetFlexPackCommand struct {
 	allowInsecureConnections bool
 	buildConfiguration       *buildUtils.BuildConfiguration
 	workingDir               string
+	// credentialEnv is the NuGetPackageSourceCredentials_<source> entry handed to the native
+	// client's environment. Empty when no credentials were injected. Set by
+	// injectCredentialsViaTempConfig and reset by its cleanup func.
+	credentialEnv string
 }
 
 // NewNuGetFlexPackCommand creates a new NuGetFlexPackCommand.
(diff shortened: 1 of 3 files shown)


artifactory/commands/nuget/command.go
```diff
@@ -627,10 +488,7 @@ func (c *NuGetFlexPackCommand) collectDependencies(buildName, buildNumber string
 // build properties on their exact Artifactory paths, and records them in local build-info.
 // The native push has already succeeded at this point, so it is never re-run; a stamping
 // failure is surfaced as an error without masking the push.
-//
-// resolvedPaths contains the absolute package paths already resolved by pushPackagesToArtifactory
-// (Artifactory bypass path). When nil (native-tool push path), paths are re-resolved from c.args.
-func (c *NuGetFlexPackCommand) collectAndStampPushArtifacts(buildName, buildNumber string, resolvedPaths []string) error {
+func (c *NuGetFlexPackCommand) collectAndStampPushArtifacts(buildName, buildNumber string) error {
```
agrasth, line 491: The old `resolveAndValidateDeployRepo` (deleted in this PR) checked two things before any upload happened: the target repo must be NuGet type, and it must not be a remote repo. `resolveLocalDeployRepo`, which replaces it here, only resolves a virtual repo to its local target — it never checks `params.PackageType` or rejects `Rclass == "remote"`. Right now, a wrong --repo gets caught late, by whatever error the native `dotnet/nuget` client happens to produce, instead of jf's own clear message.

agrasth: Maybe we can add the same two checks back into `resolveLocalDeployRepo` (or a small helper called right before `injectCredentialsViaTempConfig` where the `"no server configured"` check already lives).
Something like: fetch the repo once, error with`repository %q is of type %q, not NuGet if the package type doesn't match, and repository %q is a remote repository; NuGet packages can only be pushed to a local or virtual repository`
If it's remote. That keeps the "native tool does the work" contract intact

bhanurp: Added both checks back into `resolveLocalDeployRepo`: rejects a non-NuGet repo (`repository %q is of type %q, not NuGet...`) and a remote repo (`repository %q is a remote repository...`), same wording you suggested. Fixed in 6d2600a.

artifactory/commands/nuget/command.go
```diff
@@ -40,6 +38,10 @@ type NuGetFlexPackCommand struct {
 	allowInsecureConnections bool
 	buildConfiguration       *buildUtils.BuildConfiguration
 	workingDir               string
+	// credentialEnv is the NuGetPackageSourceCredentials_<source> entry handed to the native
+	// client's environment. Empty when no credentials were injected. Set by
+	// injectCredentialsViaTempConfig and reset by its cleanup func.
+	credentialEnv string
```
fluxxBot, line 44: can we check if the creds not getting recorded as part of jf bce command?

fluxxBot (approved):

artifactory/commands/nuget/command.go
bhanurp: Checked: `credentialEnv` is only ever appended to the *child* process's Env slice — `nativeCmd.Env = append(os.Environ(), c.credentialEnv)` in `Run()` — never via `os.Setenv` on jf's own process. `jf bce` runs later as a completely separate `jf` invocation, so it has no way to observe an env var that only ever existed inside a now-exited child process.

Added a regression test pinning this instead of relying on it by coincidence: `TestDotnetFlexPackBceCapturesEnv` in jfrog-cli#3703 (dotnet_native_test.go) now asserts none of the properties bce records contain `NuGetPackageSourceCredentials` or the test server's password/token.

agrasth (approved):

PR #556, opened 2026-09-10
RTECO-2003: add jf choco and jf setup choco

bhanurp:
## Summary

Adds `ChocoFlexPackCommand` (`jf choco`) and `configureChoco` (`jf setup choco`), for RTECO-2003.

The two are orthogonal by design: `jf setup choco` writes Chocolatey's own configuration and produces no build-info; `jf choco` produces build-info and writes no configuration.

### `jf choco`

Wraps the native client and collects build-info afterwards, with no configuration state of its own — no `jf choco-config`, no `choco.yaml`, no edits to the user's `chocolatey.config` or source list.

- **pack / push** → artifact build-info (type `nupkg`, module `<Id>:<Version>`), stamping `build.name` / `build.number` / `build.timestamp` plus `vcs.*` / `ci.*` at the artifact's exact flat path, never via a repo-wide AQL scan.
- **install / upgrade** → dependencies, with the repository resolved in order: `--repo-resolve` → the command line's `--source` → `choco source list`. Left **empty rather than guessed** when several enabled sources match.
- **everything else** → byte-faithful pass-through that never requires a configured JFrog server.

### `jf setup choco`

Registers the repository as a Chocolatey source named `jfrt-<hostname>-<repo>` at the NuGet **V2** endpoint Chocolatey requires (no `/v3/`, no `index.json`), and stores the composite `<user>:<token>` API key against that exact URL.

Keys are held **per source URL**, so any mismatch — even a trailing slash — silently leaves pushes unauthenticated; the URL is therefore built once and passed to both commands.

## Notes for reviewers

- The pack snapshot covers `choco pack --output-directory`, which the native client supports. Without it, a pack into a custom directory yields empty build-info and **no error**.
- Chocolatey's config is machine-wide, so setup carries the machine-wide scope note rather than claiming per-user scope, and fails clearly when run unelevated.
- Both commands gate on Windows from `Run()`, not registration, so `--help` still works everywhere and the error names the detected OS.

## Dependencies

Pins jfrog/build-info-go#425 and jfrog/jfrog-cli-core#1616. Both must merge first; this PR's `go.mod` pins are then re-pointed at their merged commits.

## Test plan

- [x] `go build ./...`
- [x] `go test ./artifactory/commands/choco/...`
- [x] `go test ./artifactory/commands/setup/...` — choco cases plus the two coverage tests that assert one entry per supported package manager

## Summary by CodeRabbit

- **New Features**
  - Added Chocolatey support for configuring Artifactory sources on Windows.
  - Added Chocolatey pack, push, install, and upgrade workflows with repository resolution and authentication.
  - Added build-info collection and artifact metadata for Chocolatey packages.
  - Added support for package output directories, dependency discovery, retries, and local, remote, and virtual repositories.
  - Chocolatey setup now reports machine-wide configuration scope.

- **Bug Fixes**
  - Improved secret redaction and warnings for unresolved package specification tokens.
  - Added clearer errors for unsupported platforms, missing artifacts, permissions, and transient failures.
  - Improved repository selection behavior in non-interactive environments.


agrasth (dismissed):

artifactory/commands/choco/command.go
```diff
@@ -0,0 +1,887 @@
...
+				nativeArgs = append(nativeArgs, "-u="+user)
+			}
+			nativeArgs = append(nativeArgs, "-p="+password)
+		} else {
+			log.Debug("No JFrog credentials configured; resolving from " + sourceURL + " anonymously.")
+		}
+		log.Debug("Resolving Chocolatey packages from the JFrog Artifactory source " + sourceURL)
+	}
+
+
```
fluxxBot: there are extra empty lines here

agrasth (dismissed):

artifactory/commands/choco/command.go
fluxxBot, line 1: question: is there a way we can avoid this command in the linux build and include in windows build?

fluxxBot (dismissed):

artifactory/commands/choco/command.go
bhanurp: Fixed in f620048 - removed the stray extra blank line.

artifactory/commands/choco/command.go
bhanurp: Deliberate, not an oversight: this file compiles on every OS (no build tag), and Run() gates on Windows at runtime via chocoPlatformChecker() instead - same pattern jf choco setup uses. This is so jf choco --help, and the command showing up at all in jf --help, still work on macOS/Linux, where jf CLI ships as one cross-platform binary; a linux-only exclusion via a go:build tag would need jfrog-cli's own command registration split by OS too, and would replace a clear "requires Chocolatey, which runs on Windows only" error with the command not existing at all on non-Windows. There's a dedicated cheap Linux CI job in jfrog-cli#3708 that specifically asserts the OS gate fires and --help still works there, so this is exercised, not incidental.

PR #430, opened 2026-04-23
Rteco 1013 helm

naveenku-jfrog:
- [x] All [tests](https://github.com/jfrog/jfrog-cli-core#tests) passed. If this feature is not already covered by the tests, I added new tests.
- [x] All [static analysis checks](https://github.com/jfrog/jfrog-cli-core/actions/workflows/analysis.yml) passed.
- [x] Appropriate label is added to auto generate release notes.
- [x] I used gofmt for formatting the code before submitting the pull request.
- [x] PR description is clear and concise, and it includes the proposed solution/fix.
-----
What: Added logic for handling chart name and version in subpath.
Testing: Done, locally also, and updated the tests also.

PR #563, opened 2026-09-21
RTECO-2247: PSResource (PowerShell PSResourceGet) setup + FlexPack command

bhanurp:
## Summary

Part of RTECO-2247 — adds `jf setup psresource` (registers a PSResourceGet repository against Artifactory) and `PSResourceFlexPackCommand` (the orchestration layer behind `jf Install-/Save-/Update-/Publish-PSResource`: native `pwsh` invocation, credential injection, checksum resolution, build-info collection, property stamping).

Depends on (must merge first, or this repo's `go.mod` needs re-pinning to released versions once they land):
- build-info-go: https://github.com/jfrog/build-info-go/pull/430
- jfrog-cli-core: https://github.com/jfrog/jfrog-cli-core/pull/1620

Depended on by:
- jfrog-cli: https://github.com/jfrog/jfrog-cli/pull/3717

**Note on `go.mod`**: this branch currently carries `replace` directives pinning `build-info-go` and `jfrog-cli-core/v2` to their own `RTECO-2247` branches (same canonical repos, not a fork) so this repo can build against the unreleased changes above. These need to come out once #430 and #1620 merge and this repo can pin real released versions — flagged inline in `go.mod`.

## What's here

- `artifactory/commands/setup/psresource.go` — `jf setup psresource`: derives the NuGet **v3** feed URL (`.../api/nuget/v3/<repo>/index.json`, reusing the existing `dotnet.GetSourceDetails(..., false)` helper — PSResourceGet needs v3, unlike Chocolatey which needs v2), validates the target repo is NuGet-typed, registers it via `Register-PSResourceRepository`. Deliberately does **not** persist credentials at registration time (no `-CredentialInfo`/SecretManagement) — see design note below.
- `artifactory/commands/psresource/command.go` — `PSResourceFlexPackCommand`: platform gate (pwsh + PSResourceGet module — cross-platform, not an OS check, unlike Chocolatey), per-call credential injection via child-process environment variables, HEAD-only checksum resolution (never GET, to avoid incrementing Artifactory download counters), dependency/artifact build-info collection, AQL-search-based property stamping.

**Design notes**:
- Credentials are injected per native invocation via `cmd.Env`, never as a literal command-line argument, never persisted server-side. This was a deliberate scope decision (see PRD) over `SecretManagement`-backed persistence.
- Install/Save/Update record **direct dependencies only** — PSResourceGet has no lock file, so there's no transitive graph, the same limitation `jf choco install` has.

## Notable bug fixes included (found via an adversarial multi-agent code review after the initial implementation)

- **Credential leak**: PowerShell's colon-attached flag form (`-ApiKey:secret`) bypassed redaction entirely, leaking the real secret into debug logs.
- **Credential-breaking bug**: the same redaction function was also applied to the *real* script sent to `pwsh` (not just the log line), so a user-supplied credential was silently replaced with `***` in the actually-executed command, breaking authentication for anyone passing their own credential.
- **Positional-argument handling**: `-Path` (Publish) and `-Name` (Install/Save/Update) are PSResourceGet's first *positional* parameters; the code only recognized them via explicit flags, so positional invocation either resolved the wrong module manifest or reported a successful install as a failure.
- **Retry logic**: `stampBuildProperties` never retried a successful-but-empty search result (Artifactory's async index lagging right after a HEAD-confirmed publish); separately, `fetchArtifactChecksum`'s retry-on-5xx logic was actually dead code because `jfrog-client-go`'s `httpclient.HttpClient` has its own internal zero-retry wrapper that swallows a real 5xx status into an opaque error before our code ever sees it. Both fixed.
- **Injection hardening**: any argument starting with `-` was spliced into the PowerShell `-Command` script completely unquoted with no validation it was actually a flag name; now requires a flag-shaped pattern.
- **Swallowed error**: a failing `Register-PSResourceRepository` call was replaced with a generic "install pwsh" message instead of surfacing the real cause.

## Test plan

- `go test ./artifactory/commands/psresource/... ./artifactory/commands/setup/...` — 60+ tests, all pass (including new regression tests for every fix above).
- `golangci-lint`/`gosec` clean on all new/changed files (pre-existing, unrelated findings in `setup.go`/`setup_test.go` left untouched).
- `gofmt` clean.

- [x] All tests passed. New tests added for every bug found during review.
- [x] All static analysis checks passed.

## Summary by CodeRabbit

- **New Features**
  - Added PowerShell PSResourceGet support for installing, saving, updating, and publishing packages through Artifactory.
  - Added setup configuration that registers Artifactory as a trusted PowerShell package source.
  - Added repository credential handling without persisting credentials in PowerShell source settings.
  - Added build-info and dependency tracking for supported PSResourceGet operations.
  - Added validation for supported PowerShell environments and NuGet-backed repositories.

- **Bug Fixes**
  - Improved handling of command arguments, quoting, authentication, retries, checksums, and publishing verification.
  - Added clearer errors for unavailable PowerShell environments and failed repository operations.

---
**Marked as draft**: this repo's `go.mod` carries temporary `replace` directives pointing at build-info-go#430 and jfrog-cli-core#1620 (unreleased). Will mark ready for review once those merge and this branch is re-pinned to real released versions.


artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+		return nil
+	}
+	buildName, err := command.buildConfiguration.GetBuildName()
+	if err != nil {
+		return err
+	}
+	buildNumber, err := command.buildConfiguration.GetBuildNumber()
+	if err != nil {
+		return err
+	}
```
naveenku-jfrog: Already validated buildConfiguration here https://github.com/jfrog/jfrog-cli-artifactory/pull/563/changes#diff-8721d7f8e0b3a81c405bd0be1571fa6da723037200dd43c45279d01f95a08444R177 , then why again here?

artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+		return false
+	}
+}
+
+// credentialInjectionRepo returns the repository whose credentials should be injected for the
+// current subcommand: the deploy repo for Publish-PSResource, the resolve repo for
+// Install-/Save-/Update-PSResource, or "" for every other cmdlet (Find-PSResource,
+// *-PSResourceRepository, ...), which never touch a configured Artifactory repo on this command's
+// behalf.
+func (command *PSResourceFlexPackCommand) credentialInjectionRepo() string {
```
naveenku-jfrog: Ig, Func name should be without credential.

artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+	}
+	if user == "" || password == "" {
+		// Anonymous resolution/publish is a legitimate choice when the repository allows it, so
+		// this is only a note, not an error - a later 401 would otherwise look unexplained.
+		log.Debug(fmt.Sprintf("No JFrog credentials configured for repository %q; running %s without an injected -Credential.", repo, command.subCommand))
+		return nil, false, nil
+	}
+	return []string{
+		"JFROG_PSRESOURCE_USER=" + user,
+		"JFROG_PSRESOURCE_TOKEN=" + password,
```
naveenku-jfrog: Why hardCoded strings?

artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+func splitCommaList(value string) []string {
+	parts := strings.Split(value, ",")
+	result := make([]string, 0, len(parts))
+	for _, part := range parts {
+		if trimmed := strings.TrimSpace(part); trimmed != "" {
+			result = append(result, trimmed)
+		}
+	}
+	return result
+}
```
naveenku-jfrog, line 448: This should be under utils folder.

artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+
+// firstPositionalValue returns the first leading bare token in args, or "" if the command was
+// invoked with no positional argument (e.g. everything given via explicit flags).
+func firstPositionalValue(args []string) string {
+	values := positionalValues(args)
+	if len(values) == 0 {
+		return ""
+	}
+	return values[0]
+}
```
naveenku-jfrog, line 476: This also should be under utils.

artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+		return name[idx:]
+	}
+	return ""
+}
+
+// resolvePSD1Manifest returns the module manifest (.psd1) Publish-PSResource will read for path: path
+// itself if it already names a .psd1 file, or the first .psd1 found directly inside it if it is a
+// directory.
+func resolvePSD1Manifest(path string) (string, error) {
+	info, err := os.Stat(path)
```
naveenku-jfrog, line 595: This might require system permission. Hope we have documented.

artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+	if err != nil {
+		return "", fmt.Errorf("resolve %s -Path %q: %w", SubCommandPublish, path, err)
+	}
+	if !info.IsDir() {
+		if strings.EqualFold(filepathExt(path), ".psd1") {
+			return path, nil
+		}
+		return "", fmt.Errorf("%s -Path %q is not a module manifest (.psd1)", SubCommandPublish, path)
+	}
+	entries, err := os.ReadDir(path) // #nosec G304 -- path is this same command's own -Path/working-directory argument, not externally supplied
```
naveenku-jfrog, line 607: This might require system permission. Hope we have documented.

artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+	return stampErr
+}
+
+func (command *PSResourceFlexPackCommand) moduleID() string {
+	if command.buildConfiguration != nil {
+		if module := command.buildConfiguration.GetModule(); module != "" {
+			return module
+		}
+	}
+	return "psresource-project"
```
naveenku-jfrog: hardcoded string?

artifactory/commands/psresource/command.go
```diff
@@ -0,0 +1,1026 @@
...
+// httptest.NewServer handler cannot fake convincingly, and the checksum-fetch logic this package
+// actually owns is exercised separately, directly against fetchArtifactChecksum.
+var stampBuildPropertiesFn = func(command *PSResourceFlexPackCommand, artifacts []entities.Artifact, buildName, buildNumber string) error {
+	return command.stampBuildProperties(artifacts, buildName, buildNumber)
+}
+
+// stampBuildProperties sets build.name/build.number/build.timestamp on the artifact(s) that were
+// just published, exactly as choco's stampBuildProperties does: search Artifactory for the recorded
+// repo/path to confirm it is really there, then SetProps on the result.
+func (command *PSResourceFlexPackCommand) stampBuildProperties(artifacts []entities.Artifact, buildName, buildNumber string) error {
```
naveenku-jfrog, line 921: There is already stampBuildProperties func available. can you please reuse that?

artifactory/commands/setup/psresource.go
```diff
@@ -0,0 +1,215 @@
...
+	var builder strings.Builder
+	for _, character := range strings.ToLower(value) {
+		if (character >= 'a' && character <= 'z') || (character >= '0' && character <= '9') || character == '.' || character == '_' || character == '-' {
+			builder.WriteRune(character)
+			continue
+		}
+		builder.WriteByte('-')
+	}
+	return strings.Trim(builder.String(), "-")
+}
```
naveenku-jfrog, line 155: var invalidChars = regexp.MustCompile(`[^a-z0-9._-]+`)

func sanitizePSResourceSourceComponent(value string) string {
	lower := strings.ToLower(value)
	// The "+" in the regex ensures consecutive invalid chars become a single "-"
	replaced := invalidChars.ReplaceAllString(lower, "-") 
	return strings.Trim(replaced, "-")
}
We can use something similar.

naveenku-jfrog (changes requested):

artifactory/commands/psresource/command.go
bhanurp: Good catch. Restructured Run(): buildName/buildNumber are now resolved once, right after the existing nil-guarded ValidateBuildParams() check (before the native command runs), and reused directly instead of being re-derived a second time via IsCollectBuildInfo() and a third time via GetBuildName()/GetBuildNumber() later. Fixed in 49b6d5b.

artifactory/commands/psresource/command.go
bhanurp: Renamed credentialInjectionRepo -> injectionRepo (and updated the call site + tests). Fixed in 49b6d5b.

artifactory/commands/psresource/command.go
bhanurp: Extracted JFROG_PSRESOURCE_USER and JFROG_PSRESOURCE_TOKEN into named constants (envPSResourceUser/envPSResourceToken), now shared by resolveCredentialEnv (setting them) and buildScript (referencing them in the PS script), so the two can't drift. Fixed in 49b6d5b.

artifactory/commands/psresource/command.go
bhanurp: Checked this repo for an existing utils convention for command packages before deciding: there is no artifactory/commands/utils package, and choco (the closest analog, artifactory/commands/choco/command.go) keeps its equivalent PowerShell-argument helpers (argValue/argValues/splitCommaList/positionalValues) inline in its own command.go the same way this file does - so there's no established location to move this into. Moving it into a brand-new package would mean exporting these currently-unexported helpers and updating command_test.go's white-box tests, which felt like scope creep for a review-feedback pass rather than a mechanical move. Left as-is; happy to revisit if you'd like a new shared package introduced across choco/psresource in a follow-up.

artifactory/commands/psresource/command.go
bhanurp: Same reasoning as the comment above on splitCommaList - no existing utils package/convention to move this (or positionalValues) into in this repo; choco keeps the same kind of helper inline too. Left as-is for now.

artifactory/commands/psresource/command.go
bhanurp: Added a comment on resolvePSD1Manifest documenting that os.Stat/os.ReadDir here can fail with an OS permission-denied error (e.g. -Path outside the current user's permissions), and that the wrapped error already surfaces the OS's own message + path to the caller, same as a stat/ls would show. There's no separate psresource user-facing doc in this repo (docs/buildtools/psresource lives in jfrog-cli, out of scope for this PR) to update. Fixed in 49b6d5b.

artifactory/commands/psresource/command.go
bhanurp: Same fix as above (os.ReadDir) - documented alongside os.Stat on resolvePSD1Manifest's doc comment. Fixed in 49b6d5b.

artifactory/commands/psresource/command.go
bhanurp: Extracted "psresource-project" into a defaultPSResourceModuleID constant. Fixed in 49b6d5b.

artifactory/commands/psresource/command.go
bhanurp: Grepped the whole repo for stampBuildProperties: it isn't actually a single shared function - nuget (artifactory/commands/nuget/command.go) and choco (artifactory/commands/choco/command.go) each have their own private, unexported stampBuildProperties method on their own command struct, following the same shape (search Artifactory, then SetProps). There's no common exported version to call into; each FlexPack command implements it locally because the receiver types differ. Extracting a truly shared version would mean a cross-package refactor touching nuget/choco/psresource together, which felt out of scope for this review pass. Left as-is; open to doing that extraction as a separate follow-up if you'd like.

artifactory/commands/setup/psresource.go
bhanurp: Adopted this - replaced the manual rune-by-rune builder with the regexp-based version you suggested (psresourceInvalidSourceChars = regexp.MustCompile(`[^a-z0-9._-]+`), collapsing runs to a single '-'). Verified it against the existing TestSanitizePSResourceSourceComponent cases (including the "///" -> "" and "Team Repo/Release" -> "team-repo-release" cases) - same behavior, simpler implementation. Fixed in 49b6d5b.

naveenku-jfrog (approved):

